Issue #25, opened 2019-12-09
Module selection & page offset on mobile devices


christian-heusel:
# current state :exclamation: 
![Screenshot_20191209-023021_Samsung Internet](https://user-images.githubusercontent.com/26827864/70400725-a063ce00-1a2c-11ea-8110-869ba72db187.jpg)

# expected behavior :100: 
* the selection buttons do not differ in size
* the selection buttons are aligned on mid level to the search bar
* there is equal spacing between the search bar and the 2 buttons

:snake: :man_firefighter: :1st_place_medal:

fosterfarrell9:
I will take care of this after the holidays.

TheEbolaDoc:
Whilst playing around with boostrap for another project I noticed that the strange page offset is caused by the `row` class here:
https://github.com/fosterfarrell9/mampf/blob/28c054d9375450e6b47bcd46e4b857284088d297/app/views/shared/_footer.html.erb#L2
https://github.com/fosterfarrell9/mampf/blob/28c054d9375450e6b47bcd46e4b857284088d297/app/views/shared/_footer.html.erb#L35
https://github.com/fosterfarrell9/mampf/blob/28c054d9375450e6b47bcd46e4b857284088d297/app/views/shared/_footer.html.erb#L54

Whilst removing it resolves the issue ofc the content formatting is lost, but knowing what causes the bug is atleast something.

Merry Christmas :christmas_tree: :gift:

christian-heusel:
Why did this issue get closed? As far as I know one of the two misbehaviours remains unsolved @henrixapp  ...?

henrixapp:
Sorry. We just saw your one commit and thought this addresses the whole issue.

christian-heusel:
Its better now but some of this stuff should collapse into a menu and the search field and Course name should be on different lines
![image](https://user-images.githubusercontent.com/26827864/95321577-86d52200-089b-11eb-9601-b026482dc449.png)

PR #27, opened 2019-12-27
Fixed permanent page offset

TheEbolaDoc:
This fixes one of the issues described in #25.
I reworked the code for the footer, but the final result should look similar to the one that is used at the moment.

# New appearance :fire:

## :computer: Desktop :computer: 
image

## :iphone: Mobile devices :iphone: 
image


fosterfarrell9:
Thanks a lot! I will merge it into production when I am back from vacation.

PR #59, opened 2020-07-27
Mail design

christian-heusel:
This Pull Request tries to fix the Issue #41 and make email great again :100:


fosterfarrell9:
The html head should not be inside the `confirmation_instructions.html.erb` partial. Instead,  create a file `devise_mailer.html.erb` inside `app/views/layouts` that contains the head as in your partial and has
```
<body>
  <%= yield %>
</body>
```
for a body. `devise_mailer.html.erb` only needs the code inside the body. Then, add the line `layout 'devise_mailer'` to `mailers/my_mailer.rb`. This will make sure that all devise mailer partials will use this layout. It might be worth thinking about using one unified layout for notification mails for media etc. and mails coming from devise (registration, password change etc.). For that I think one would have to apply the changes done in this pull request to `my_mailer.rb` to `application_mailer.rb` and set `config.parent_mailer = 'ApplicationMailer'` in `config/initializers/devise.rb`. The `MyMailer` class could then be deleted and the unified layout should be put in `app/views/layouts/mailer.html.erb`.

christian-heusel:
I will take care of both mail systems separately, mabe I will need some help merging both together as suggested ...
So far I only changed the devise part.

app/views/devise/mailer/reset_password_instructions.html.erb
```diff
@@ -1,19 +1,21 @@
-<p>
-  <%= t('.greetings', name: @resource.name) %>
+<p class="h1">
+  <%= t('.greetings',
+        name: @resource.name.present? ? name : t('.no_name_fallback')) %>
```
fosterfarrell9, line 3: This is not going to work, as `name` will be an unknown variable. It should be
```
<%= t('.greetings', 
      name: @resource.name.present? ? @resource.name : t('.no_name_fallback')) %>
```

app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -1,8 +1,9 @@
-<p>
+<p class="h1">
   <%= t('.greeting',
         recipient: @resource.email,
         locale: @resource.locale,
-        default: "Welcome #{@resource.email}!") %>
+        name: name.present ? name : t('no_name_fallback'),
```
fosterfarrell9, line 5: This is not going to work, as `name` will bean undefined local variable. It should be
```
name: @resource.name.present? ? @resource.name : t('no_name_fallback'),
```
But I do not understand this anyway. In `devise.de.yml`, `confirmation_instructions.greeting` does not have a `name` variable (it is just "Willkommen bei MaMpf"), the same goes for `devise.en.yml`. So, that is what will be shown. Furthermore, at this point in the registration process, the `name` attribute of the user who registers is still set to `nil`. So, the above ternary operator will always give `t('no_name_fallback')`.

app/views/devise/mailer/reset_password_instructions.html.erb
fosterfarrell9: I just realized you can make it even shorter by using the `#presence` method of Rails:
```
<%= t('.greetings', 
      name: @resource.name.presence || t('.no_name_fallback')) %>
```

app/views/devise/mailer/confirmation_instructions.html.erb
christian-heusel: It was too late when I worked on this, thanks for spotting my mistake! :hankey:

app/views/devise/mailer/reset_password_instructions.html.erb
christian-heusel: Thanks, imo the syntax of this is a bit weird, but that seems like the way to go.
I verified that the name is displayed when existing (after adding the `@resource.` prefix ofc)

app/views/notification_mailer/announcement_email.html.erb
```diff
@@ -1,7 +1,13 @@
-<%= "#{t('mailer.new_announcement')} #{@announcement_details}:" %>
-<p>
-  <%= "#{t('mailer.author')}: #{@announcement.announcer.name}" %>
-  <br>
-  <%= "#{t('mailer.text')}: " %>
-  <%= @announcement.details.html_safe %>
-</p>
\ No newline at end of file
+<p class="h1">
+  Mitteilungstitel
```
christian-heusel, line 2: @fosterfarrell9 Imo this would be a nice change, but its too much for me at the moment since the model / db / form etc. has to be changed

fosterfarrell9 (changes requested):

app/views/notification_mailer/new_course_email.html.erb
```diff
@@ -1,5 +1,10 @@
-<%= t('mailer.new_course', title: @course.title) %>
-<br>
-<%= t('mailer.subscribe_course_html',
-      profile: link_to(t('notifications.profile'),
-                       edit_profile_url)) %>
\ No newline at end of file
+<%= t('mailer.new_course', title: @lecture.title_with_teacher) %>
+<br>
+<%= t('mailer.subscribe_course_html',
+      profile: t('notifications.profile')) %>
```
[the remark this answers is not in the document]
fosterfarrell9: Hm, now it won't work because `mailer.subscribe_course` does not exist in the locale files. Maybe my explanation was not very good. (See [here](https://guides.rubyonrails.org/i18n.html) in 4.4 for an explanation what the `_html` means.)
I would suggest
```
<%= t('mailer.subscribe_course') %>
```
and setting `mailer.subscribe_course` appropriately in the .yml files (without profile paramter), deleting `mailer.subscribe_course_html`

app/views/notification_mailer/new_lecture_email.html.erb
```diff
@@ -1,5 +1,10 @@
-<%= t('mailer.new_lecture', title: @lecture.title_with_teacher) %>
-<br>
-<%= t('mailer.subscribe_lecture_html',
-      profile: link_to(t('notifications.profile'),
-                       edit_profile_url)) %>
\ No newline at end of file
+<%= t('mailer.new_lecture', title: @lecture.title_with_teacher) %>
+<br>
+<%= t('mailer.subscribe_lecture_html',
+      profile: t('notifications.profile')) %>
```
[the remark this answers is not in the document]
fosterfarrell9: This will work but as above I would suggest using
```
<%= t('mailer.subscribe_lecture') %>
```
and setting `mailer.subscribe_lecture` appropriately in the locale files (in particular, without profile parameter).

PR #65, opened 2020-08-24
Correct the problems with Travis CI & the dockersetup

christian-heusel:


docker/run_tests/run_tests.sh
```diff
@@ -5,9 +5,9 @@ if ! [ -f completed_initial_run ]
 then
   echo 'Initialising mampf' &> >(tee -a /usr/src/app/log/initialisation.log)
   echo running: bundle exec rails db:migrate &> >(tee -a /usr/src/app/log/initialisation.log)
-  bundle exec rails db:migrate &> >(tee -a /usr/src/app/log/initialisation.log)
+  bundle exec rails db:migrate > /usr/src/app/log/initialisation.log
```
fosterfarrell9, line 8: There should be a `rails db:create` before, otherwise the migrations will fail (and they do fail, at least in my attempt to build the container).

fosterfarrell9 (changes requested):

fosterfarrell9 (approved):
Looks good now. By the way, if you want the tests to run through (and well, have half of them fail), the following files should be deleted because they "test" stuff that does no longer exist and therefore block the whole process:

- spec/controllers/api/v1/media_controller_spec.rb
- spec/controllers/api/v1/tags_controller_spec.rb
- /spec/models/vote_spec.rb

(That maybe more related to #66 and would be the first step there, but it also makes the tests look more realistic :-) )

christian-heusel:
Feel free to merge the PR then, I wont touch the tests :blush:

PR #72, opened 2020-09-07
Autosetup

christian-heusel:
This pullrequest automates the setup for the db & uploads significantly, but whether that has any bad side effects is yet to be tested :P
It also fixes a bug which made the initialization not persist.

### DIFF
--- a/docker/development/Dockerfile
+++ b/docker/development/Dockerfile
@@ -11,7 +11,8 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
     curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
     curl -sS https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && echo "deb http://apt.postgresql.org/pub/repos/apt/ `lsb_release -cs`-pgdg main" | tee  /etc/apt/sources.list.d/pgdg.list && \
     sed -i "s/deb\.debian\.org/ftp.halifax\.rwth-aachen.de/g" /etc/apt/sources.list && \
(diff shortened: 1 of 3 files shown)


christian-heusel:
Since heiBOX does not set the Upload date correctly it would be good to put the files somewhere else to make use of the `--timestamping`option for `wget` (only redownload if there is a newer version of the file on the server).

tynsh:
Maybe move entrypoint.sh to docker/development as well?

christian-heusel:
Sure, why not :)

docker/development/Dockerfile
```diff
@@ -10,7 +10,9 @@ ENTRYPOINT ["/usr/src/app/entrypoint.sh"]
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
     curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
     curl -sS https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && echo "deb http://apt.postgresql.org/pub/repos/apt/ `lsb_release -cs`-pgdg main" | tee  /etc/apt/sources.list.d/pgdg.list && \
-    apt-get update && apt-get install -y nodejs yarn postgresql-client-12 ffmpeg imagemagick pdftk ghostscript sqlite3 --no-install-recommends && rm -rf /var/lib/apt/lists/* && apt-get clean
+    sed -i "s/deb\.debian\.org/ftp.halifax\.rwth-aachen.de/g" /etc/apt/sources.list && \
```
tynsh, line 13: Why this?

christian-heusel: I noticed downloadspeeds are kinda suboptimal and added a german mirror ...
I hope there is a better way to do this but I dont know of any, maybe debian has a progam like arch's [`reflector`](https://wiki.archlinux.org/index.php/reflector)?

tynsh:
ftp.de.debian.org should be the meta mirror for germany

tynsh:
@fosterfarrell9 You wanted my opinion on this: You can merge this it its current state before merging mampf-next into main. Ideally I'd like to play around with this to ensure I have no further ideas for improvements :)

PR #75, opened 2020-09-16
fixes #74

### LINKED ISSUE #74: Empty locale cookie crashes rendering
If you supply an empty locale cookie 🥠 (done by accident  while testing), the application renders an error and is unusable.

The culprit is here https://github.com/fosterfarrell9/mampf/blob/f117f860a277e6219183861605c0aadc17f4fb58/app/controllers/application_controller.rb#L80

This causes the `enforce_available_locales!` to crash, because the locale "" is unknown.

A possible way to mitigate this, would be:
```ruby
if cookies[:locale].in?(I18n.available_locales.map(&:to_s))
  cookie_param = cookies[:locale]
end
I18n.locale = current_user.try(:locale) || locale_param ||
                    cookie_param || I18n.default_locale
```

fosterfarrell9:
This should fix #74 (and make the code a little cleaner along the way).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -70,7 +70,7 @@ class ApplicationController < ActionController::Base
 
   def set_locale
     I18n.locale = current_user.try(:locale) || locale_param ||
-                    cookie_param || I18n.default_locale
+                    cookie_locale_param || I18n.default_locale
     unless user_signed_in?
       cookies[:locale] = I18n.locale
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -73,11 +73,8 @@ class ApplicationController < ActionController::Base
   end
 
   def set_locale
-    if params[:locale].in?(I18n.available_locales.map(&:to_s))
-      locale_param = params[:locale]
-    end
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -98,4 +95,18 @@ def store_interaction
                                    request.referrer,
                                    study_participant)
   end
+
+  def locale_param
+    return unless params[:locale].in?(available_locales)
+    params[:locale]
+  end
+
+  def cookie_param
```
henrixapp, line 104: I think this name `cookie_param` is a not straight forward and confusing, because the method does return the locale and not any cookie param (as I would expect). I would suggest  renaming it to `cookie_locale_param`or `locale_cookie_param`.

henrixapp (commented):
Would rename one function.

app/controllers/application_controller.rb
fosterfarrell9: Good point.

henrixapp (changes requested):
The `set_locale`method still called the old `cookie_param`, so the cookie will be never considered.

fosterfarrell9:
Oops.

henrixapp (approved):

PR #92, opened 2020-10-05
use strict_cookie mthod for cookie storing, remove jquery filedownloa…

### LINKED ISSUE #77: Firefox complains about cookie same site attribute
Apparently, there is a problem with our cookies and their same site attribute. In the future, these cookies will be declined.

fosterfarrell9:
This should fix #77. In the process, the jquery_filedownload-rails gem was removed (it triggered a cookie warning and is not really necessary).


app/controllers/courses_controller.rb
```diff
@@ -9,7 +9,6 @@ class CoursesController < ApplicationController
 
   def edit
     I18n.locale = @course.locale || I18n.default_locale
-    cookies[:edited_course] = params[:id]
```
henrixapp, line 12: Why is this line removed?

Gemfile.lock
```diff
@@ -246,15 +246,12 @@ GEM
       nokogiri (>= 1.6)
     jbuilder (2.10.0)
       activesupport (>= 5.0.0)
-    jquery-datetimepicker-rails (2.4.1.0)
```
henrixapp, line 249: Was this removed earlier?

henrixapp (commented):
We have to discuss the use of `x_sendfile`.

app/controllers/courses_controller.rb
fosterfarrell9: I did not see any place in the codebase where this cookie is actually used (anymore).

Gemfile.lock
fosterfarrell9: That is just an artefact of me playing around with that gem.

PR #106, opened 2020-10-16
Proper user deletion

fosterfarrell9:
This pull request implements the feature that lectures/media can remain in the database when their editor/teacher deletes her/his MaMpf account. The account gets destroyed, but these materials (and only these) will be transfered to a new user who gets an `archived` flag.


henrixapp (changes requested):
As admin I can delete myself without confirming my password :see_no_evil: . Maybe this sanity check should be done for every deletion.

fosterfarrell9:
Good point, I added the sanity check (users have to enter their password).

PR #107, opened 2020-10-18
Add exception handler gem

fosterfarrell9:
This adds  the `exception_handler` gem in order that we have more professional looking exception messages than Rails' default message, and to get notification mails when exceptions happen.
I opted for a rather minimalistic email layout and used `application_no_sidebar` for the view. Feel free to improve that. 
In order to get it running in development, set `config.consider_all_requests_local` in `config/environments/docker_development.rb` to `false`. (The suggestion of the gem to  set `dev` to true in the `config.exception_handler` hash in `config/application.rb` does not work as that only covers development, but not `docker_development`.)


app/views/exception_handler/exceptions/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="jumbotron border border-danger exceptionBox">
+	<div>
+	 	<%= t('exception.user_text') %>
+		<i class="fas fa-2x fa-bug"></i>
```
henrixapp, line 4: That looks quite ugly: Maybe the icon should be moved upwards or use a red version of the mampf logo with a broken M. Maybe we need a little animal mascot (the mampfling? a squirrel?), that is crying.

config/locales/de.yml
```diff
@@ -3052,6 +3052,14 @@ de:
         Die Hausaufgabe konnte nicht gelöscht werden. Das kann z.B. daran
         liegen, dass es dafür bereits Abgaben gibt, für die Dateien hochgeladen
         wurden.
+  exception:
+    exception: Exception auf %{host}
+    exception_report: 'Es gab eine Exception auf %{host}:'
+    user_text: >
+      Ups, das hätte nicht passieren sollen. Das ist ein Bug in MaMpf.
+    infos_wanted_html: >
```
henrixapp, line 3060: We already get notified, so maybe tell this the user. and ask for more information or use it as advertisment for programmers.

henrixapp (changes requested):
Although users should not see the page that often, the general design is not satisfying.

app/views/exception_handler/exceptions/show.html.erb
fosterfarrell9: That sounds good. But I guess these graphical things are outside of my expertise.

config/locales/de.yml
fosterfarrell9: I reformulated it. The contact infos are in the footer anyway.

PR #131, opened 2020-11-04
Date picker

henrixapp:


app/views/assignments/_form.html.erb
```diff
@@ -11,9 +11,14 @@
...
       <div class="form-group col-3">
-        <%= f.datetime_field :deadline,
+        <%= f.text_field :deadline,
                              class: 'form-control',
                              id: "assignment_deadline_#{assignment.id}" %>
+        <script type="text/javascript">
+         $("#assignment_deadline_<%= @assignment.id %>").datetimepicker({ format:'d.m.Y H:i',
+  inline:false,
+});
+        </script>
```
fosterfarrell9, line 21: Is it possible to do this unobtrusive like in some extra pack via webpacker  (or in app/asstes/javascript via sprockets)?

fosterfarrell9 (changes requested):
Works like a charm.

app/views/assignments/_form.html.erb
henrixapp: Yes, it must be wrapped in a document ready function. I placed it there, that is obvious what is happening.

fosterfarrell9: I would rather like it in a separate datetimepicker.js or. coffee, and then wrapped inside a turbolinks:load funktion.

henrixapp: The problem is, that the datepickers are not available in the dom in the suggested event. I have moved it to a seperate file, but this makes it not obvious, when it is needed.

henrixapp:
Closed in favour of #190

Issue #135, opened 2020-11-05
Lecture locale ignored in submission views


fosterfarrell9:
The views for submissions are always displayed in the default locale because the corresponding controllers do not set the locale based on the lecture.

PR #140, opened 2020-11-08
Locales for submissions

fosterfarrell9:

Bug fix (#135)

- Linter
    - [ ] `rubocop` reports equal or less errors and warnings **in total**
    - [ ] `yarn lint` reports equal or less errors and warnings **in total**
    - [ ] `coffeelint .` reports equal or less errors and warnings **in total**
    - [ ] `erblint .` reports equal or less errors and warnings **in total**

The controllers associated to the following models do not set specific locales: `assignment, submission, tutorial, quiz_certificate`
So, although the localizations are present in `.yml` locale files, they are not displayed.

The correct locales are set in the controllers, so they are displayed in the views.

* **Other information**:

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -7,14 +7,17 @@ class AssignmentsController < ApplicationController
 
   def new
     @assignment = Assignment.new
-    lecture = Lecture.find_by_id(params[:lecture_id])
-    @assignment.lecture = lecture
+    @lecture = Lecture.find_by_id(params[:lecture_id])
+    @assignment.lecture = @lecture
+    set_assignment_locale
   end
 
   def create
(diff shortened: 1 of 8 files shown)


henrixapp (commented):
I think that sometimes the user's locale should be preferred, esp. for tutors.

fosterfarrell9:
We once made the policy that the teacher of a lecture determines the language, and that applies to anything related to the lecture. I still think that is the right policy, as this makes things very uniform. Otherwise you might have things like a sidebar in english, and different cards in the body in different languages. But we could discuss this over on basecamp.

henrixapp:
I just noticed the fact, that for now the user setting has virtually no impact, we should probably postpone this discussion.

PR #153, opened 2020-11-18
Add fallback links

henrixapp:
Feature

No copy links displayed

In some mails there are now copyable links

no

* **Other information**:

### DIFF
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -18,3 +18,6 @@
                 confirmation_url(@resource, confirmation_token: @token, locale: I18n.locale),
                 :class => "btn btn-primary" %>
 </p>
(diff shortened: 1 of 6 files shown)


fosterfarrell9 (changes requested):
For the confirmation mail, I think it is fine. For the invitation mail, I think the styling makes the fallback link way too dominant - when I open that mail, my looks are immediately centered at the fallback link (together with "possible problem" text around it). I think the look should be way more defensive, in particular given the fact that we have not received one single complaint about non-working links in real production.

fosterfarrell9:
To resolve this, maybe we could just add a sentence in the footnote in the generic mail layout, like "If you have trouble with the links in the above mail, try to copy and paste them directly to your browser.".

fosterfarrell9:
We decided to pursue a more defensive strategy as in #183

Issue #143, opened 2020-11-11
Timeout on upload


christian-heusel:
Steps to reproduce:

1. Grab a large Video
2. Upload it as media
3. See shrine getting stuck on "Please Wait/Bitte warten"

This is probably due to a NginX configuration error.
https://stackoverflow.com/questions/24453388/nginx-reverse-proxy-causing-504-gateway-timeout/36589120#36589120

```
today at 09:37 nginx.1 | 2020/11/10 08:37:52 [warn] 99#99: *9761 upstream server temporarily disabled while reading response header from upstream, client: 192.168.144.1, server: localhost, request: "POST /videos/upload HTTP/1.1", upstream: "http://192.168.144.6:3000/videos/upload", host: "localhost:3009", referrer: "https://mampf.mathi.uni-heidelberg.de/media/16161/edit"
today at 09:37 nginx.1 | 2020/11/10 08:37:52 [error] 99#99: *9761 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 192.168.144.1, server: localhost, request: "POST /videos/upload HTTP/1.1", upstream: "http://192.168.144.6:3000/videos/upload", host: "localhost:3009", referrer: "https://mampf.mathi.uni-heidelberg.de/media/16161/edit"
```

fosterfarrell9:
Enlarging the RAM of the server has solved this problem (together with the optimizations of #155 and #166). Files of several GB can be uploaded without problems, so this issue can be considered closed for now.

PR #155, opened 2020-11-18
Extract metadata later

fosterfarrell9:

Bug fix attempt.

Upload of large files videos sometimes (#143). One reason maybe that metadata extractiontakes too long (especially when RAM is scarce) and causes a timeout.

This PR moves the extraction of video metadata to a sidekiq worker that is triggered only when the user saves the video.

* **Other information**:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -170,7 +170,7 @@ Layout/SpaceInsideParens:
 # Check quotes usage according to lint rule below.
 Style/StringLiterals:
   Enabled: true
-  EnforcedStyle: double_quotes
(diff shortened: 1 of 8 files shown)


henrixapp:
There is a warning, but I think this omit-able:

```
SHRINE WARNING: The "mime_type" Shrine metadata field will be set from the "Content-Type" request header, which might not hold the actual MIME type of the file. It is recommended to load the determine_mime_type plugin which determines MIME type from file content.
```
Furthermore, you might want reduce the number of ruby lint errors.

henrixapp (approved):

PR #158, opened 2020-11-21
limit filename length in bulk download zip

fosterfarrell9:

Bug fix

File names in bulk download zip can become too large for Windows.

File names in bulk download zip cannot become too large for Windows.

* **Other information**:

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -94,7 +94,7 @@ class Submission < ApplicationRecord
   end
 
   def filename_for_bulk_download
-		(users.map(&:tutorial_name).join('-') + '-' +
+		(team.first(180) + '-' +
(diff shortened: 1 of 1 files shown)


app/models/submission.rb
```diff
@@ -94,7 +94,7 @@ def file_path
   end
 
   def filename_for_bulk_download
-		(users.map(&:tutorial_name).join('-') + '-' +
+		(team.first(180) + '-' +
```
henrixapp, line 97: I'm not sure how you came up with 180.

fosterfarrell9: I roughly estimated that the ID and the Date and possibly 'LATE' should be around 60 characters. Maximum filename length in Windows seems to be 256. So I took 180 in case my estimate was rather bad :-) But feel free to make an exact count.

PR #159, opened 2020-11-21
allow putting announcements on MaMpf home page

fosterfarrell9:

Feature

As admin, you can create general announcements, bu there is no way to put them on the MaMpf homepage (which would be useful e.g. to announce server downtimes)

As admin you can set announcements on the homepage (this is done under /announcements) and remove them from the homepage.

* **Other information**:
Run db:migrate (the announcement table has a new boolean column :on_main_page).

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -2,6 +2,7 @@
 class AnnouncementsController < ApplicationController
   authorize_resource
   layout 'administration'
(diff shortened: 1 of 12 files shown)


henrixapp (commented):
Works as intended, but the notice is not shown to already logged in users on their overview page. So we might want to add their announcements too.

fosterfarrell9:
I added the announcement to the start page as well.

PR #203, opened 2020-12-23
Add substitute method for users in course

fosterfarrell9:

Bug fix

In #189 we removed the course/user join table. Unfortunately we overlooked to references to to course#users in the views.

Write a course#users method that takes into account all users who have subscribed to lectures in that course.

* **Other information**:

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -207,6 +207,11 @@ class Course < ApplicationRecord
     false
   end
 
+  def users
+    User.where(id: LectureUserJoin.where(lecture: lectures)
+                                  .pluck(:user_id).uniq)
+  end
+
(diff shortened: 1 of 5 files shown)


henrixapp (commented):
In effect this removes some information. Do we need this subscribers count?

fosterfarrell9:
Since the point when we introduced the new start page it is no longer possible to subscribe to courses (only to lectures), so there is no generic subscribers count for a course any longer. Of course, we could add a substitute method like this in `course.rb`:

```
def users
  User.where(id: LectureUserJoin.where(lecture: lectures).pluck(user_id).uniq)
end
```
which returns all users that have subscribed a lecture that belongs to this course.

henrixapp (approved):

Issue #66, opened 2020-08-24
fix the test or fix the bugs


christian-heusel:
Since I neither really know how the tests work nor how MaMpf itself really functions someone else should take care of this.
We are striving to get a working CI configuration at the moment, but will not be done by getting this done since the tests currently either fail for no reason (which means the tests should be fixed) or there are some bugs which should be addressed in their issues.

christian-heusel:
Is there any progress on this?

Splines:
@christian-heusel #441 and #453 should close this, right?

PR #204, opened 2020-12-23
Establish factories

fosterfarrell9:

This PR establishes working factories for all models. This represents step zero towards the resolution of #66. 
Furthermore, tests of model methods, controllers, views and helpers have been temporarily commented out - in the next step they all need to be refactored completely. The factories all have working tests of their own. So at the moment, all tests are working, but overall coverage is rather low.
Also, rubocop rules have been enforced on the `spec` folder.

If anyone of the referees has an idea on how to prevent shrine from cluttering the output of rspec when testing attaching of files please let me know or change it yourself.

* **Other information**:


config/locales/de.yml
```diff
@@ -3221,6 +3221,12 @@ de:
             title:
               blank: 'Es muss ein Titel angegeben werden.'
               taken: 'Dieser Titel ist bereits vergeben.'
+        clicker_vote:
```
henrixapp, line 3224: Why are these translations introduced now? just curious :grinning:

henrixapp (approved):
I have skimmed over it, looks good! Definitely needed for the cypress e2e tests.

config/locales/de.yml
fosterfarrell9: IIRC I built some invalid stuff in the creation of some factory. But the error message was not that it is invalid but that the translation of this error for ActiveRecord was missing. So I thought, better correct it right now...

PR #206, opened 2020-12-26
Unit tests for courses

fosterfarrell9:

Unit tests for the Course model 

For `rspec`, `config.use_transactional`_fixtures has been set to `false` to allow database_cleaner to do its'  job. `database_cleaner` is configured in `spec/support/database_cleaner.rb`. This has been done as `rspec` did not roll back inside `before :all` blocks, resulting in problems with validations with uniqueness checks.
Also, some factories have been slightly modified to better accomodate the needs of the u nit tests.

* **Other information**:


henrixapp (approved):
The failing linter counts can be ignored, they are caused by intermediate removes/wrong configuration.

PR #209, opened 2020-12-29
Refactor course

fosterfarrell9:

The `Course` model has been refactored. In the process several methods have been extracted to new service PORO models `QuestionSampler` and `TeachableParser`.

* **Other information**:

### DIFF
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -241,7 +241,8 @@ class Medium < ApplicationRecord
   def self.search_by(search_params, page)
     search_params[:types] = [] if search_params[:all_types] == '1'
     if search_params[:teachable_inheritance] == '1'
-      search_params[:teachable_ids] = Course.search_inherited_teachables(search_params)
+      search_params[:teachable_ids] =
+        TeachableParser.new(search_params).inherited_teachables_as_strings
     end
     search_params[:teachable_ids] = [] if search_params[:all_teachables] == '1'
     search_params[:editor_ids] = [] if search_params[:all_editors] == '1'
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -240,11 +240,8 @@ class Medium < ApplicationRecord
   # by the controller
   def self.search_by(search_params, page)
     search_params[:types] = [] if search_params[:all_types] == '1'
-    if search_params[:teachable_inheritance] == '1'
-      search_params[:teachable_ids] =
-        TeachableParser.new(search_params).inherited_teachables_as_strings
-    end
-    search_params[:teachable_ids] = [] if search_params[:all_teachables] == '1'
+    search_params[:teachable_ids] = TeachableParser.new(search_params)
+                                                   .teachables_as_strings
     search_params[:editor_ids] = [] if search_params[:all_editors] == '1'
     if search_params[:all_tags] == '1' && search_params[:tag_operator] == 'and'
       search_params[:tag_ids] = Tag.pluck(:id)
--- a/app/models/teachable_parser.rb
+++ b/app/models/teachable_parser.rb
@@ -1,60 +1,65 @@
-class TeachableParser
+# frozen_string_literal: true
 
-  # params is a hash with keys :all_teachables, :teachable_ids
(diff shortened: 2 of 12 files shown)


app/models/medium.rb
```diff
@@ -240,10 +240,8 @@ def self.search_sorts(search_params)
   # by the controller
   def self.search_by(search_params, page)
     search_params[:types] = [] if search_params[:all_types] == '1'
-    if search_params[:teachable_inheritance] == '1'
-      search_params[:teachable_ids] = Course.search_inherited_teachables(search_params)
-    end
-    search_params[:teachable_ids] = [] if search_params[:all_teachables] == '1'
+    search_params[:teachable_ids] = TeachableParser.new(search_params)
+                                                   .teachables_as_strings
```
henrixapp, line 244: This line is not covered by tests explicitly.

app/models/teachable_parser.rb
```diff
@@ -0,0 +1,65 @@
...
+                    .map { |t| t.remove('Lecture-') }
+    end
+
+    def lectures
+      Lecture.where(id: lecture_ids)
+    end
+
+    def course_ids
+      @teachable_ids.select { |t| t.start_with?('Course') }
+                    .map { |t| t.remove('Course-') }
```
henrixapp, line 43: I don't know how type-safe this operation is, I would expect it to return an array of int and not string.

app/models/teachable_parser.rb
```diff
@@ -0,0 +1,65 @@
...
+  def teachables_as_strings
+    return [] if @all_teachables
+    return @teachable_ids unless @inheritance
+
+    teachables_with_inheritance.map { |t| "#{t.class}-#{t.id}" }
+  end
+
+  private
+
+    def lecture_ids
```
henrixapp, line 32: See below.

spec/models/teachable_parser_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(teachable_parser.teachables_as_strings).to eq([])
+    end
+
+    it 'returns the given teachable strings if teachable_inheritance flag'\
+       'is not set' do
+      teachable_parser = FactoryBot.build(:teachable_parser,
+                                          teachable_ids: [@course1_str],
+                                          teachable_inheritance: '0')
+      expect(teachable_parser.teachables_as_strings)
+        .to eq([@course1_str])
```
henrixapp, line 39: likely this should be an int.

henrixapp (commented):
One line is actually not covered.

The types of the ids in teachableparser are somehow not those of ActiveRecord I think.

app/models/medium.rb
fosterfarrell9: I would like to postpone that until I deal with tests for `medium.rb`. I will need to think about how to test this Solr stuff in `medium#search_by` best, and there will be probably a bigger setup needed for testing this method, so I would prefer not to write anything ad hoc right now as it would most likely have to be changed a lot later anyway.

app/models/teachable_parser.rb
fosterfarrell9: Yeah, conceptionally it should definitely be an array of integers (even though ActiveRecord does not care and accepts strings for ids as well). I will add a `.map(&:to_i)`.

spec/models/teachable_parser_spec.rb
fosterfarrell9: I do not quite understand what you mean. `teachable_ids` and `teachables_as_strings` both have to encode the type of teachable (Course, Leecture, Lesson) as well, so these are contain strings made up of `Course-`, `Lecture-`, `Lesson-` and an integer id.

app/models/teachable_parser.rb
fosterfarrell9: I will add a .map(&:to_i).

spec/models/teachable_parser_spec.rb
henrixapp: oh sorry. I miss read that line.

henrixapp (approved):

PR #215, opened 2021-01-16
Subscribe lecture button fixes #197

### LINKED ISSUE #197: Subscribe button on no access page

henrixapp:

Feature

- [x] E2E Tests for the changes have been added  via Cypress

#197 

You can try to subscribe on to a lecture on its page.

No.

* **Other information**:

### DIFF
--- a/app/controllers/profile_controller.rb
+++ b/app/controllers/profile_controller.rb
@@ -79,6 +79,8 @@ class ProfileController < ApplicationController
                 !@lecture.in?(current_user.lectures) &&
                 @lecture.passphrase != @passphrase
     @success = current_user.subscribe_lecture!(@lecture)
+
+    redirect_to lecture_path(@lecture) if @parent == "redirect"
   end
 
   def unsubscribe_lecture
@@ -148,6 +150,7 @@ class ProfileController < ApplicationController
 
   def set_lecture
     @lecture = Lecture.find_by_id(lecture_params[:id])
+    p(@lecture)
     @passphrase = lecture_params[:passphrase]
     @parent = lecture_params[:parent]
     @current = !@parent.in?(['lectureSearch', 'inactive'])
--- a/app/controllers/profile_controller.rb
+++ b/app/controllers/profile_controller.rb
@@ -150,7 +150,6 @@ class ProfileController < ApplicationController
 
   def set_lecture
     @lecture = Lecture.find_by_id(lecture_params[:id])
-    p(@lecture)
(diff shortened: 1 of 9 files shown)


henrixapp:
http_caching issue must be resolved.

app/models/ability.rb
```diff
@@ -213,7 +213,7 @@ def initialize(user)
 
       can [:show_announcements, :organizational,
            :show_structures, :search_examples, :search, :show_random_quizzes,
-           :display_course], Lecture
+           :display_course, :subscribe_page], Lecture
```
fosterfarrell9, line 216: This does the trick only for generic users. The corresponding rules should also be added to admins/editors/teachers (their rulesets are some lines above in the other parts of the `if` block).

app/controllers/profile_controller.rb
```diff
@@ -148,6 +150,7 @@ def email_params
 
   def set_lecture
     @lecture = Lecture.find_by_id(lecture_params[:id])
+    p(@lecture)
```
fosterfarrell9, line 153: This can be removed :-)

config/locales/en.yml
```diff
@@ -2163,6 +2163,7 @@ en:
     click_and_paste: >
       If you click this term, it will be copied to the "fulltext" field.
   errors:
+    not_subscribed: You are currently not subscribed to this course.
```
fosterfarrell9, line 2166: It should be "You are currently not subscribed to this lecture".

config/locales/de.yml
```diff
@@ -2296,6 +2296,7 @@ de:
     click_and_paste: >
       Wenn Du den Begriff anklickst, wird er in das Suchfeld "Volltext" kopiert.
   errors:
+    not_subscribed: Du bist zur Zeit nicht in dieser Vorlesung AbonnentIn.
```
fosterfarrell9, line 2299: I would suggest some rewording like "Aktuell hast Du diese Veranstaltung nicht abonniert."

fosterfarrell9 (changes requested):

fosterfarrell9 (changes requested):
The availability of this feature for editors and teachers has yet to be added, is does not work for them as of now: see the comment in the last feature. The reason ist that a `can :subscribe_page, Lecture` is missing for them in `ability.rb` (without that, they are redirected to main with the usual error message).

fosterfarrell9 (approved):

PR #236, opened 2021-02-21
Improve THymE volume level updating

alyaeanyx:
This pull request changes the behaviour of the volume bar of the THymE media player.  Previously, the volume level was only updated as soon as the volume bar was released (which was a bit confusing at least for me). This change sets the event listener for the "volume-bar" input element from "change" to "input", causing the update function to be triggered instantly as soon as the volume bar is moved.


henrixapp (approved):
Works fine. Actually this is only a change by one line. The other changes are caused by an inconsistent merge by @fosterfarrell9.

For future contributions please start working from mampf-next @alyaeanyx 

Thank you for your first contribution!

PR #244, opened 2021-03-04
Release date for media

fosterfarrell9:

Feature

The release of a medium happens just after clicking the "Publish" button.

You can optionally set a date when the medium is scheduled to be released. Additionally, if the Medium is an exercise sheet, you can set a flag such that an assignment will be auto-created upon publication. 

* **Other information**:
All business logic concerning media publication has been extracted to a service model `MediumPublisher`.


henrixapp:
Reminder to @henrixapp: Write cypress tests for this.

app/models/medium.rb
```diff
@@ -49,12 +49,14 @@ class Medium < ApplicationRecord
 
   has_many :quiz_certificates, foreign_key: 'quiz_id', dependent: :destroy
 
-  has_one :assignment
+  has_many :assignments
```
henrixapp, line 52: Why is this changed and not documented?

fosterfarrell9: I guess I forgot - the `has_one` version is actually a bug. Currently you can create multiple assignments for a single medium without being stopped - it is simply not checked. And I think there is no reason why one medium should not be able to have more than one assignment.

henrixapp (approved):

PR #264, opened 2021-04-01
send submission deletion info mails in batches

fosterfarrell9:

bug fix

* **Other information**:

### DIFF
--- a/app/models/submission_cleaner.rb
+++ b/app/models/submission_cleaner.rb
@@ -79,8 +79,8 @@ class SubmissionCleaner
       local_submitter_ids = @submitters.where(locale: l).pluck(:id)
       next if local_submitter_ids.empty?
 
-      local_submitter_ids.in_groups_of(200) do |group|
-        NotificationMailer.with(recipients: group.compact,
+      local_submitter_ids.in_groups_of(200, false) do |group|
+        NotificationMailer.with(recipients: group,
                                 term: @previous_term,
                                 locale: l)
                           .submission_destruction_email.deliver_now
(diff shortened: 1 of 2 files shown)


app/models/submission_cleaner.rb
```diff
@@ -78,10 +78,13 @@ def send_destruction_mail_to_submitters
     I18n.available_locales.each do |l|
       local_submitter_ids = @submitters.where(locale: l).pluck(:id)
       next if local_submitter_ids.empty?
-      NotificationMailer.with(recipients: local_submitter_ids,
-                              term: @previous_term,
-                              locale: l)
-                        .submission_destruction_email.deliver_now
+
+      local_submitter_ids.in_groups_of(200) do |group|
```
henrixapp, line 82: The last group will contain nil values: 
```ruby
[1,2,3].in_groups_of(20)
 => [[1, 2, 3, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil]]
```
Changing it to `in_groups_of(200,false)` should fix it:
```ruby
[1,2,3].in_groups_of(20,false)
 => [[1, 2, 3]]
```

app/models/submission_cleaner.rb
```diff
@@ -102,13 +105,15 @@ def send_info_mail_to_submitters
...
       local_submitter_ids = @submitters.where(locale: l).pluck(:id)
       next if local_submitter_ids.empty?
-        NotificationMailer.with(recipients: local_submitter_ids,
-                                term: @previous_term,
-                                deletion_date:
-                                  @previous_term.submission_deletion_date,
-                                reminder: @reminder,
-                                locale: l)
-                          .submission_deletion_email.deliver_now
+        local_submitter_ids.in_groups_of(200) do |group|
```
henrixapp, line 108: Same issue here

henrixapp (changes requested):
Possible causing a problem for `nil` ids

henrixapp (approved):

PR #271, opened 2021-04-08
re-structure and clean up dockerfile

christian-heusel:
This PR increases the robustness of the production dockerfile as requested by @fosterfarrell9 😄 

The sidekiq assets are copied from a fixed path determined by the version

The sidekiq assets are copied from the currently installed version

hopefully not!

### DIFF
--- a/docker/production/Dockerfile
+++ b/docker/production/Dockerfile
@@ -16,19 +16,16 @@ EXPOSE 3000
 ENTRYPOINT ["/usr/src/app/docker/entrypoint-worker.sh"]
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-        ffmpeg imagemagick pdftk ghostscript sqlite3 rsync shared-mime-info && \
-    rm -rf /var/lib/apt/lists/* && \
-    curl -sL https://deb.nodesource.com/setup_14.x | bash && \
-    apt-get update && \
-    apt-get install -y nodejs && \
-    rm -rf /var/lib/apt/lists/* && \
+    # setup all the repos
     curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
     echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    curl -sL https://deb.nodesource.com/setup_14.x | bash && \
+    # update the sources with the repos set up
     apt-get update && \
-    apt-get install -y yarn && \
-    rm -rf /var/lib/apt/lists/*
-
+    # install all the dependencies
+    apt-get install -y --no-install-recommends \
+        ffmpeg imagemagick pdftk ghostscript rsync shared-mime-info && \
+    apt-get install -y nodejs yarn
 
 RUN sed -i '/disable ghostscript format types/,+6d' /etc/ImageMagick-6/policy.xml
 
--- a/docker/production/Dockerfile
+++ b/docker/production/Dockerfile
@@ -15,33 +15,34 @@ EXPOSE 3000
 
 ENTRYPOINT ["/usr/src/app/docker/entrypoint-worker.sh"]
 
-# https://github.com/nodesource/distributions#installation-instructions
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-
-RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
-
-RUN apt-get update && apt-get install -y nodejs yarn ffmpeg imagemagick pdftk ghostscript sqlite3 rsync shared-mime-info --no-install-recommends && rm -rf /var/lib/apt/lists/*
-
-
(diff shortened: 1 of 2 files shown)


docker/production/Dockerfile
```diff
@@ -15,33 +15,34 @@ EXPOSE 3000
...
-
-
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash \
- && apt-get update && apt-get install -y nodejs && rm -rf /var/lib/apt/lists/* \
- && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
- && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
- && apt-get update && apt-get install -y yarn && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        ffmpeg imagemagick pdftk ghostscript sqlite3 rsync shared-mime-info && \
```
tynsh, line 20: Why did we install sqlite? Can it be removed?

docker/production/Dockerfile
```diff
@@ -15,33 +15,34 @@ EXPOSE 3000
...
+    rm -rf /var/lib/apt/lists/* && \
```
tynsh, line 21: This can be removed

docker/production/Dockerfile
```diff
@@ -15,33 +15,34 @@ EXPOSE 3000
...
+    curl -sL https://deb.nodesource.com/setup_14.x | bash && \
+    apt-get update && \
+    apt-get install -y nodejs && \
+    rm -rf /var/lib/apt/lists/* && \
```
tynsh, line 25: This is also redundant

docker/production/Dockerfile
```diff
@@ -15,33 +15,34 @@ EXPOSE 3000
...
 WORKDIR /usr/src/app
 USER app
 COPY --from=build-pdfcomprezzor /go/src/pdfcomprezzor.wasm  /usr/src/app/public/pdfcomprezzor/pdfcomprezzor.wasm
 
 COPY --chown=app:app ./Gemfile ./Gemfile.lock ./yarn.lock ./package.json /usr/src/app/
 
-RUN bundle install
-RUN yarn  install --production=false
+RUN bundle install && \
+    yarn  install --production=false
```
tynsh, line 45: Why did we disable production again?

docker/production/Dockerfile
```diff
@@ -15,33 +15,34 @@ EXPOSE 3000
...
 COPY --chown=app:app . /usr/src/app
-RUN cp -r /usr/local/bundle/gems/sidekiq-6.2.0/web/assets /usr/src/app/public/sidekiq
-RUN SECRET_KEY_BASE="$(bundle exec rails secret)" DB_ADAPTER=nulldb   bundle exec rails assets:precompile
+RUN cp -r $(bundle info --path sidekiq)/web/assets /usr/src/app/public/sidekiq && \
```
tynsh, line 47: Looks sane

tynsh (approved):
So my opinion is, this is safe to merge and nothing changes. I just don't remember why we disabled yarn production mode and installed sqlite

docker/production/Dockerfile
fosterfarrell9: It is probably a remnant from the first dockerfiles where we used sqlite in development.

fosterfarrell9: So it can be removed.

docker/production/Dockerfile
fosterfarrell9: That is probably a question for @henrixapp .

henrixapp: I think webpacker and other dev-Deps would not not be installed, if this is `production=true`. We could move the devDeps in the package.json to dependencies and everything should work as expected.

docker/production/Dockerfile
christian-heusel: Why can everything here be removed? 🤔

docker/production/Dockerfile
christian-heusel: How is it redundant? 
According to [the script](https://deb.nodesource.com/setup_14.x) itself it just installs the repo and not the program:
```
#!/bin/bash

# Discussion, issues and change requests at:
#   https://github.com/nodesource/distributions
#
# Script to install the NodeSource Node.js 14.x repo onto a
# Debian or Ubuntu system.
#
# Run as root or insert `sudo -E` before `bash`:
#
# curl -sL https://deb.nodesource.com/setup_14.x | bash -
#   or
# wget -qO- https://deb.nodesource.com/setup_14.x | bash -
#
# CONTRIBUTIONS TO THIS SCRIPT
#
# This script is built from a template in
# https://github.com/nodesource/distributions/tree/master/deb/src
# please don't submit pull requests against the built scripts.
```

henrixapp (approved):

henrixapp:
The one failing test is not caused by your code segment. Merge it at any time

docker/production/Dockerfile
christian-heusel: Resolving because of missing feedback!

docker/production/Dockerfile
christian-heusel: Resolving because of missing feedback!

PR #283, opened 2021-05-04
Add playback speed fixes #113

### LINKED ISSUE #113: Playback speed in editor
It should be possible to set the video speed while taking.

henrixapp:

* **Other information**:


fosterfarrell9 (changes requested):
Now I remember why I did not implement this feature in the thyme editor: Depending on the window size (and this happens even on my 4k monitor with rather large windows), the additional button overlaps with the three editing buttons that are not in the thyme player. Therefore I am not really sure we gain more than we lose by having this additional button.

PR #299, opened 2021-05-19
Synced translation files, Fixes #297

### LINKED ISSUE #297: English Submission Invitation Email
Submission invitation emails do not contain any content or links when sent in English.

It looks like this is (at least in part) due to incorrect naming in the `en.yml` locale file (currently under `mampf/config/locales/`). The item `submission_invitation` in `en.yml` is called `submission_invitation_html` in the German locale `de.yml`. More items differ in name, but I will not list them here.

My proposed (but untested) fix would be to rename the items in the English locale file to have the same names as in the German locale file.

Edit:

An affected email carries the tag `translation missing: en.mailer.submission_invitation_html [...]` instead of the link.

henrixapp:

Fixes translation files.

* **Other information**:

### DIFF
--- a/app/views/administration/index/_my_lectures.html.erb
+++ b/app/views/administration/index/_my_lectures.html.erb
@@ -42,7 +42,7 @@
     <% if current_user.lectures_as_module_editor.present? %>
       <%= t('admin.main.as_course_editor') %>
       (<%= current_user.lectures_as_module_editor.count %>)
-      <%= helpdesk(t('admin.main.info.lectures_as_course_editor'), false) %>
+      <%= helpdesk(t('admin.main.info.lectures_as_module_editor'), false) %>
       <%= render partial: 'administration/index/lectures_card',
                  locals: { lectures:
                              Lecture.sort_by_date(current_user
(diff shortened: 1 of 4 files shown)


app/views/administration/index/_my_lectures.html.erb
```diff
@@ -42,7 +42,7 @@
     <% if current_user.lectures_as_module_editor.present? %>
       <%= t('admin.main.as_course_editor') %>
       (<%= current_user.lectures_as_module_editor.count %>)
-      <%= helpdesk(t('admin.main.info.lectures_as_course_editor'), false) %>
+      <%= helpdesk(t('admin.main.info.lectures_as_module_editor'), false) %>
```
fosterfarrell9, line 45: I would prefer to leave the naming as it is, since it refers to the `course` model. As a consequence there is probably some line in the `.yml` file that needs changing.

fosterfarrell9 (changes requested):

app/views/administration/index/_my_lectures.html.erb
henrixapp: There is also `lectures_as_module_editor` in `user.rb`. should this be also renamed?

fosterfarrell9: Yes, that makes sense.

fosterfarrell9 (approved):

PR #303, opened 2021-05-31
outdated link removed

hungrywords:
The mampf-package is now available elsewhere. It might be necessary to delete "resources: Weg" (l. 2400) too.


henrixapp:
I don't think that we should remove the link, I think I updated it.

PR #304, opened 2021-05-31
outdated link removed

hungrywords:
The MaMpf-package can be found elsewhere. It might be necessary to delete "resources: here" (l. 2257) too.


henrixapp:
We should provide a link. The pull request #278  updated the link.

PR #314, opened 2021-06-07
Return more fileformats

henrixapp:

Adds the ability to upload different fileformats then the submission as correction.


henrixapp:
Let lecturers set fileformat for corrections.

PR #316, opened 2021-06-11
Updated sheets

zebleck:

Indicator for changed manuscripts after a mediums release.

-added the migration for the file_last_edited attribute (default value: nil)
-is set to the update_date of the medium when the medium is released and the manuscript is changed.

Note that file_last_edited is nil as long as the manuscript is not updated. 
Note that file_last_edited is also set when the medium is already released and a manuscript is uploaded for the first time for that medium.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -63,6 +63,9 @@ class MediaController < ApplicationController
     # create screenshot for manuscript if necessary
     changed_manuscript = @medium.manuscript_data != old_manuscript_data
     if @medium.manuscript.present? && changed_manuscript
+      if @medium.released
+        @medium.file_last_edited = @medium.updated_at
+      end
       @medium.manuscript_derivatives!
       @medium.save
     end
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -63,9 +63,7 @@ class MediaController < ApplicationController
     # create screenshot for manuscript if necessary
     changed_manuscript = @medium.manuscript_data != old_manuscript_data
     if @medium.manuscript.present? && changed_manuscript
-      if @medium.released
-        @medium.file_last_edited = @medium.updated_at
-      end
+      @medium.file_last_edited = @medium.updated_at if @medium.released
       @medium.manuscript_derivatives!
       @medium.save
     end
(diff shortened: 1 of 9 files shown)


app/controllers/media_controller.rb
```diff
@@ -63,6 +63,9 @@ def update
     # create screenshot for manuscript if necessary
     changed_manuscript = @medium.manuscript_data != old_manuscript_data
     if @medium.manuscript.present? && changed_manuscript
+      if @medium.released
+        @medium.file_last_edited = @medium.updated_at
```
fosterfarrell9, line 67: To make it more ruby-esque one would probably write this as
`@medium.file_last_edited = @medium.updated_at if @medium.released`

app/views/media/_manuscript_card.html.erb
```diff
@@ -62,6 +62,13 @@
             </span>
           </div>
         </div>
+        <% if medium.sort == 'Nuesse' && medium.file_last_edited != nil %>
```
fosterfarrell9, line 65: I would suggest putting this in a card-footer, see [here](https://getbootstrap.com/docs/4.6/components/card/#header-and-footer).

app/models/medium.rb
```diff
@@ -580,6 +580,14 @@ def sort_localized
     Medium.sort_localized[sort]
   end
 
+  def subheader_style
```
fosterfarrell9, line 583: Similar as above. It would be more ruby-esque to use a guard clause. The `self` is redundant.
```
def subheader_style
  return "badge badge-secondary" unless sort == "nuesse" && file_last_edited
  "badge badge-danger"
end
```

app/views/media/_medium.html.erb
```diff
@@ -90,7 +90,7 @@
               </h5>
             </div>
             <div>
-              <span class="badge badge-secondary"
+              <span class="<%= medium.subheader_style %>"
```
fosterfarrell9, line 93: I would suggest adding  a tooltip that explains the situation (file update etc.) in case that there is a badge-danger. Otherwise users will not know what this means.

fosterfarrell9 (commented):
First of all, thanks for the PR! I just got some Ruby-Style and layout comments. For the Ruby stuff you can also run rubocop (its suggestions will probably be more or less the same).

fosterfarrell9 (approved):

henrixapp:
Closed in favor of #313

PR #328, opened 2021-07-09
Improved teacher tutorial view

zebleck:

Improved the view of editors and teacher when you click on "Tutorials" in a lecture. The view is similar to that of tutors, but instead of bulk uploading they can bulk download corrections.

Screenshots:

image

image


app/controllers/tutorials_controller.rb
```diff
@@ -84,6 +95,23 @@ def bulk_download
...
+    @zipped_corrections = Submission.zip_corrections!(@tutorial, @assignment)
+    if @zipped_corrections.is_a?(StringIO)
+    send_data @zipped_corrections.read,
+              filename: @assignment.title + '@' + @tutorial.title + '-Corrections.zip',
+              type: 'application/zip',
+              disposition: 'attachment'
+    else
+      flash[:alert] = I18n.t('controllers.tutorials.bulk_download_failed',
+                             message: @zipped_corrections)
+      redirect_to lecture_tutorials_path(@tutorial.lecture,
```
fosterfarrell9, line 108: Is it intended that the redirect is to `lecture_tutorials_path`? `lecture_tutorials_path` goes to 'tutorials#index' and there access is only granted to an actual tutor due to `before_action :check_tutot_status, only: :index`. So, this would lead to a no access warning for editors.

app/models/submission.rb
```diff
@@ -105,13 +110,25 @@ def filename_for_bulk_download
...
+			last_modification_by_users_at.strftime("%F-%H%M") +
+			(too_late? ? '-LATE' : '') +
+			+ '-ID-' + id + '-CORRECTION' +
+			assignment.accepted_file_type)
+			.gsub(/[\x00\/\\:\*\?\"<>\|]/, '_')
+	   	.gsub(/^.*(\\|\/)/, '')
+   		# Strip out the non-ascii characters
+   		.gsub(/[^0-9A-Za-z.\-]/, '_')
+  end
+
```
fosterfarrell9, line 124: It seems to me that `filename_for_corrections_bulk_download` and `filename_for_submissions_bulk_download` are quite similar, so it should be possible to refactor this (e.g. using some common private method that they both use) in order to avoid code duplication.

app/models/submission.rb
```diff
@@ -122,6 +139,25 @@ def self.zip_submissions!(tutorial, assignment)
...
+          end
+        end
+      end
+      archived_filestream.rewind
+    rescue => e
+      archived_filestream = e.message
+    end
+    archived_filestream
+  end
+
```
fosterfarrell9, line 160: Same as above: It should be possible to refactor `zip_submissions!` and `zip_corrections!` in order to avoid code duplication.

app/views/submissions/_correction.html.erb
```diff
@@ -1,23 +1,39 @@
...
-                    data: { turbolinks: false } %>
-        <%= link_to t('buttons.delete'),
-                    delete_correction_path(submission),
-                    class: 'btn btn-sm btn-outline-secondary mb-2',
-                    data: { confirm: t('confirmation.generic') },
-                    method: :delete,
-                    remote: true %>
-      <% end %>
-    </span>
+    <% if !current_user.editor_or_teacher_in?(@lecture) %>
```
fosterfarrell9, line 3: I think that this is a bug: It is possible that someone is an editor of the lecture and a tutor at the same time (e.g. if the "Obertutor" also has a tutorial which happens sometimes). In this logic this person would not have access to the "tutor's buttons".

app/views/tutorials/_submission_row.html.erb
```diff
@@ -39,9 +39,11 @@
     <%= render partial: 'submissions/correction_column',
                locals: { submission: submission } %>
   </div>
-  <div class="col-3 submission-actions"
-       data-id="<%= submission.id %>">
-    <%= render partial: 'submissions/other_actions',
-               locals: { submission: submission } %>
-  </div>
+  <% if !current_user.editor_or_teacher_in?(@lecture) %>
```
fosterfarrell9, line 42: Probably the same as above.

app/views/tutorials/index_teacher.html.erb
```diff
@@ -0,0 +1,109 @@
+<% if @assignments.present? %>
```
fosterfarrell9, line 1: `tutorials/index_teacher.html.erb` and `tutorials/index.html.erb` share a lot of code. I think it is advisable to refactor, e.g. by putting stuff that they both use into partials.

app/views/tutorials/_table.html.erb
```diff
@@ -13,9 +13,11 @@
           <%= t('basics.correction') %>
           (<%= assignment.accepted_file_type %>)
         </b>
-        <b class="col-3">
-          <%= t('basics.actions') %>
-        </b>
+        <% if !current_user.editor_or_teacher_in?(@lecture) %>
```
fosterfarrell9, line 16: Probably same as above.

app/views/tutorials/overview.html.erb
```diff
@@ -1,113 +1,137 @@
...
+          <button class="btn btn-secondary dropdown-toggle"
+                  data-toggle="dropdown"
+                  type="button">
+            <%= @assignment.title %>
+          </button>
+          <div class="dropdown-menu">
+            <% (@assignments - [@assignment]).each do |a| %>
+              <%= link_to a.title,
+                          lecture_tutorial_overview_path(@lecture,
+                                                        params: { assignment: a.id }),
```
fosterfarrell9, line 14: `params` should be aligned to `@lecture`.

app/views/tutorials/overview.html.erb
```diff
@@ -1,113 +1,137 @@
...
+          <button class="btn btn-secondary dropdown-toggle"
+                  data-toggle="dropdown"
+                  type="button">
+            <%= t('basics.overview') %>
+          </button>
+          <div class="dropdown-menu">
+            <% (@tutorials - [@tutorial]).each do |t| %>
+              <%= link_to t.title,
+                          lecture_tutorials_teacher_path(@lecture,
+                                                        params: { assignment: @assignment.id,
```
fosterfarrell9, line 36: `params` should be aligned to `@lecture`.

app/views/tutorials/overview.html.erb
```diff
@@ -1,113 +1,137 @@
...
+          <% @lecture.tutorials.natural_sort_by(&:title).each do |t| %>
+            <tr>
+              <td>
+                <%= t.title %>
+              </td>
+              <td>
+                <%= t.tutor_names || t('basics.tba') %>
+              </td>
+              <td>
+                <%= Submission.where(tutorial: t, assignment: @assignment)
```
fosterfarrell9, line 83: This was not introduced by you, so I am sorry to bother you with it, but I think we should use the chance to change it. The code for this and the following queries should not be within the view. There should be methods in some model that do these queries. (Additionally, as these queries happen to occur in a loop, I guess we have some n+1 action going on here (which should not really be problematic as the number of tutorials i alecture should be rather small, but still)).

fosterfarrell9 (changes requested):
Thank you for the PR. I think there is one bug that is related to users that are editors and tutors at the same time (see comments). I also recommend some refactorisations. 
Also, I just remembered that there where some teachers who complained that as of now, one can only see how many submissions are already in for a certain assignment after the deadline is over (they just want to see the number of (not necessarily final) submissions to be sure that everything is working alright). Maybe this PR is a chance to incorporate this as well.

app/models/submission.rb
henrixapp: done. I also moved the `filename_for_bulk_download` to `private`.

app/views/submissions/_correction.html.erb
henrixapp: should we grant the lecturer/"Obertutor" all access to the submission/correction process?

app/models/tutorial.rb
```diff
@@ -79,7 +79,19 @@ def teams_to_csv(assignment)
...
+  end
+  def too_late_count_for(assignment)
+    Submission.where(tutorial: self, assignment: assignment)
+              .where.not(manuscript_data: nil)
+              .select { |s| s.too_late? }.size
+  end
+  def corrected_count_for(assignment)
+    Submission.where(tutorial: self, assignment: assignment)
+              .where.not(correction_data: nil).size 
+  end
```
henrixapp, line 94: This code seems to not in use and be replaced by a portion in `models/submissions.rb`. You can remove it.

app/models/user.rb
```diff
@@ -520,6 +520,14 @@ def editor_or_teacher_in?(lecture)
     in?(lecture.editors) || self == lecture.teacher
   end
 
+  def tutorials(lecture)
+    given_tutorials.where(lecture: lecture)
+  end
+
+  def has_tutorials?(lecture)
+    given_tutorials.where(lecture: lecture).size.positive?
```
henrixapp, line 528: I think this could be expressed in a simpler way: `!given_tutorials.where(lecture:lecture).empty?`

henrixapp (changes requested):
Thank you for this pull request. There are two places that can be fixed in a minute and then we can merge.

henrixapp (approved):

fosterfarrell9 (approved):

PR #331, opened 2021-08-15
Directify

henrixapp:

This PR removes uppy and replaces it by an own XHR upload.

- [x] E2E Tests for the changes have been added  via Cypress

Uppy is used for controlling uploads

directUpload manages now this upload process.

No

* **Other information**:

Removes Uppy dependencies (about half mb)!


app/models/submission.rb
```diff
@@ -190,6 +190,51 @@ def check_file_properties(metadata, sort)
...
+    submissions = Submission.where(tutorial: tutorial,
+                                   assignment: assignment).proper
+    report = { successful_saves: [], submissions: submissions.size,
+    					 invalid_filenames: [], invalid_id: [], in_subfolder: [],
+               no_decision: [], rejected: [], invalid_file: [] }
+    tmp_folder = Dir.mktmpdir
+    begin
+      files.each do |file_shrine|
+      #Zip::File.open(zipfile) do |zip_file|
+      #  zip_file.each do |entry|
```
fosterfarrell9, line 203: The comments can probably be removed as they represent outdated code (?). They seem rather confusing to me.

app/models/submission.rb
```diff
@@ -190,6 +190,51 @@ def check_file_properties(metadata, sort)
...
+        p "EEE2",file_shrine
+          filename = file_shrine["metadata"]["filename"]
```
fosterfarrell9, line 205: It seems that l205ff need no intendation.

app/models/submission.rb
```diff
@@ -190,6 +190,51 @@ def check_file_properties(metadata, sort)
     { sort => errors }
   end
 
+  def self.bulk_corrections!(tutorial, assignment, files)
```
fosterfarrell9, line 193: This method replaces Tutorial#add_bulk_corrections! which hence becomes redundant, but it is still contained in tutorial.rb.

app/views/tutorials/bulk_upload.coffee
```diff
@@ -1,4 +1,5 @@
 $('#upload-bulk-errors').empty().hide()
+#directUpload('upload-bulk-correction',)
```
fosterfarrell9, line 2: I do not understand the comment.

app/assets/javascripts/upload.coffee
```diff
@@ -373,60 +271,162 @@ bulkCorrectionUpload = (fileInput) ->
...
+        f = fileInput.files[0]
+        formData = new FormData()
+        formData.append("file", f, f.name)
+        xhr.open('POST', endpoint, true)
+        xhr.send formData
+      else
+        alert(
+          $(progressBarElement).data 'tr-missing-consent'
+        )
+
```
fosterfarrell9, line 416: Is there a way to add an additional info just after 100% is reached (like "upload is being processed")? With large files, there is this delay between the point when 100% is reached and the file info is being displayed. This might confuse some users.

fosterfarrell9 (changes requested):
Addendum: It would probably be good if this could be tested in production on an iPad.

app/assets/javascripts/upload.coffee
henrixapp: Done!

app/assets/javascripts/upload.coffee
```diff
@@ -365,6 +365,7 @@ directUpload provides an interface to upload (multiple) files to an endpoint
         onprogress = (e) ->
             percentUpload = Math.floor(100/fileInput.files.length * e.loaded / e.total+100*(i)/fileInput.files.length)
             $(progressBarElement).text(percentUpload+" %")
+            $(progressBarElement).text("Processing...") if percentUpload == 100
             return
```
fosterfarrell9, line 369: I would suggest that "Processing..." is localized.

fosterfarrell9 (changes requested):

fosterfarrell9 (approved):

PR #333, opened 2021-08-25
Playback statistics

henrixapp:

Updates the chart.js dependency and creates plots for video access and downloads.

Bildschirmfoto 2021-08-25 um 10 59 09

The statistics only show the total number of downloads/thyme play backs.

There is a nice plot.

no

* **Other information**:

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -375,12 +375,12 @@ class MediaController < ApplicationController
     medium_consumption = Consumption.where(medium_id: @medium.id)
     if @medium.video.present?
       @video_downloads = medium_consumption.where(sort: 'video',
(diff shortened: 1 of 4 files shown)


fosterfarrell9 (changes requested):
I would be nice if the manuscript downloads also had a graph.
Also, I would suggest a little padding for the access graph.
If the numbers are low (like 2 accesses), it seems that nonnatural numbers are displayed by chart.js along the vertical axis (counts: 1 - 1,5 - 2).

fosterfarrell9 (changes requested):
Now that I have played around with it it again, I think that the total number of downloads/player accesses should not be disregarded and still be listed (otherwise people will have to manually integrate the graph ;-) ).

PR #345, opened 2021-09-29
Protected homework

henrixapp:

Allows the protection of homework. This homework will be stored indefinitely.

All Homework is deleted on a fixed date after the semester.

Some Assignments can be protected.

* **Other information**:

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -65,7 +65,7 @@ class AssignmentsController < ApplicationController
 
   def assignment_params
     params.require(:assignment).permit(:title, :medium_id, :lecture_id,
-                                       :deadline, :accepted_file_type)
+                                       :deadline, :accepted_file_type,:protected)
   end
 
   def check_editor_status
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -65,7 +65,7 @@ class AssignmentsController < ApplicationController
 
   def assignment_params
     params.require(:assignment).permit(:title, :medium_id, :lecture_id,
-                                       :deadline, :accepted_file_type,:protected)
+                                       :deadline, :accepted_file_type, :protected)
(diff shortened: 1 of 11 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -65,7 +65,7 @@ def set_assignment_locale
 
   def assignment_params
     params.require(:assignment).permit(:title, :medium_id, :lecture_id,
-                                       :deadline, :accepted_file_type)
+                                       :deadline, :accepted_file_type,:protected)
```
fosterfarrell9, line 68: Space is missing before `:protected`.

app/models/term.rb
```diff
@@ -62,11 +62,18 @@ def previous
   def assignments
     Assignment.where(lecture: lectures)
   end
+  def unprotected_assignments
+    Assignment.where(lecture: lectures,protected: false)
```
fosterfarrell9, line 66: Space is missing before `protected`.

config/locales/de.yml
```diff
@@ -2685,6 +2685,9 @@ de:
       Es wurde noch nichts abgegeben.
     nothing_yet_in_lecture: >
       In dieser Vorlesung war bisher noch nichts abzugeben.
+    protected: Geschützte Abgabe
+    protected_info: >
+      "Geschützte" Abgaben werden bis zur manuellen Löschung aufbewahrt und werden nicht automatisch gelöscht.
```
fosterfarrell9, line 2690: I suggest a reformulation:
... und werden nicht automatisiert nach Ablauf der Aufbewahrungsfrist (Semesterende + Karenzzeit von 14 Tagen) gelöscht.

fosterfarrell9 (approved):
In the text, it is written that protected assignments are stored until "manual deletion". But there is no such thing (unless an admin does it on the console). I guess this should be the topic of a future PR.

PR #364, opened 2022-03-03
Simplify docker-gen and nginx setup

tynsh:
This fixes the broken proxy image. Currently a hack to generate the default.conf for nginx no longer works due to an image update. This should provide a long term solution which is also easier to work with.

To simplify the setup we separate docker-gen from the actual nginx instance. From the original nginx.tmpl we only take the necessary parts.

I've added nginx to the development setup and would like some input on that. My reasoning was, that it is closer to production and allows testing new nginx configs. But during operation requests are only handled through. An other thought was, that I'd like to replace nginx with the openresty distribution in order to use the nginx lua module. We could fix the current problems with the prometheus exporter with that by concatinating responses from all workers.


tynsh:
Henrik and I had a talk about this. There are some issues left, that prevent the exporter from working.

tynsh:
This is nearly done. So if you want to have a working setup, merge this now.

There is a minor problem left: /metrics shows garbage text in browsers but works with curl. Chrome lets you download a gzip somehow.

fosterfarrell9:
In my development setup, localhost refuses to connect when I try to log in:

`HTTP Origin header (http://localhost:3000) didn't match request.base_url (http://workers)`
`Redirected to http://workers/`

tynsh:
> In my development setup, localhost refuses to connect when I try to log in:
> 
> `HTTP Origin header (http://localhost:3000) didn't match request.base_url (http://workers)` `Redirected to http://workers/`

Thanks for pointing this out. It is fixed now. I've tested this in experimental, so from a functionality point of view I consider this to be ready for review and merging.

I'm only unsure if it is easier to work with the codebase without nginx in development. So if you have any issues with code auto reloading or anything else, I can also remove it from the development docker-compose file.

**Production deployment needs to be done manually once!** This is because new containers are spawned and the config needs to be ready for this. So when this PR goes into production, let me know. mampf-next/mampf-dev is already configured.

tynsh:
If new containers are added more often, we could think about making the docker-compose files part of this repository. But then github workflows need to be setup to decline changes to them. If someone can make changes to docker-compose files in production, they are effectively root there.

fosterfarrell9:
Thanks, it is working nicely in development now as well. I think the addition/removal of containers will not happen so often, so I think the setup given in this PR should be fine.

docker/development/docker-compose.yml
```diff
@@ -134,7 +134,32 @@ services:
...
 
+  nginx:
+    image: openresty/openresty:alpine
+    # Until https://github.com/nginx-proxy/docker-gen/pull/311 is merged, use hardcoded name
+    container_name: mampf-docker-nginx
+    ports:
+      - "127.0.0.1:3000:80"
+    volumes:
+      - "nginx-conf-d:/etc/nginx/conf.d"
+      - "submissions:/private/submissions:ro,nocopy"
```
henrixapp, line 145: What is the `submission` volume?

docker/development/docker-compose.yml
```diff
@@ -134,7 +134,32 @@ services:
...
+
+  dockergen:
+    build:
+      context: ./../..
+      dockerfile: docker/Dockerfile.dockergen
+    command: -notify-sighup mampf-docker-nginx -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    restart: always
+    volumes:
+      - "nginx-conf-d:/etc/nginx/conf.d"
+      - "../nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl"
```
henrixapp, line 156: why `../nginx.tmpl` and not `docker/nginx.tmpl`?

docker/production/docker-compose.production.yml
```diff
@@ -85,16 +99,17 @@ volumes:
...
+      o: "addr=yourfileserverdomain,noatime,nolock,soft,vers=4.2,rw"
       device: ":/mampf/submissions/mampf"
   public:
+  nginx-conf-d:
   caches:
     driver_opts:
       type: "nfs"
-      o: "addr=krikkit,noatime,nolock,soft,vers=4.2,rw"
+      o: "addr=yourfileserverdomain,noatime,nolock,soft,vers=4.2,rw"
       device: ":/mampf/caches/mampf"
```
henrixapp, line 115: maybe add some infos about this into `INSTALL.md`

henrixapp (commented):
Haven't tried it out on a local machine, but looks good. Probably some infos about these changes should be reflected in `INSTALL.md`

docker/development/docker-compose.yml
tynsh: Submissions are exercise sheets submitted by students. The need to be handled by nginx, because they are served there. Nginx still asks rails for permissions on those files and rails answers this without sending the actual file contents. This is called accelerated redirect and is done on the server anyway.

docker/development/docker-compose.yml
tynsh: The working directory is such that the intuitive path docker/nginx.tmpl becomes ../nginx.tmpl internally.

docker/production/docker-compose.production.yml
tynsh: Done.

tynsh:
If there are no more comments, I think this could be merged.

PR #376, opened 2022-03-17
added xkcd back without gem dependency

zebleck:

* **Other information**:

### DIFF
--- a/app/models/xkcd.rb
+++ b/app/models/xkcd.rb
@@ -1,9 +1,11 @@
 # Xkcd model for getting Xkcd images
-class Xkcd < ApplicationRecord
+class Xkcd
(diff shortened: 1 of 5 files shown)


app/models/xkcd.rb
```diff
@@ -0,0 +1,9 @@
+# Xkcd model for getting Xkcd images
+class Xkcd < ApplicationRecord
```
fosterfarrell9, line 2: This class does not need to inherit from `ApplicationRecord`, it should be a plain Ruby class:

`class Xkcd`

fosterfarrell9 (changes requested):

zebleck:
requests fixed

PR #420, opened 2022-12-01
Media search

zebleck:

Generic user search for media (on start page)
Allows for different filtering criteria to be selected (tags, description, teacher, lecture, etc.)
Consistent with admin search if same criteria chosen 
Also setup for request tests and implementation for media search
Fixes of some unrelated cypress tests

- [-] E2E Tests for the changes have been added  via Cypress
- [X] Meaningful rspec tests have been added
- [-] Docs have been added / updated 

* **Other information**:


app/controllers/media_controller.rb
```diff
@@ -207,17 +211,49 @@ def inspect
   # return all media that match the search parameters
   def search
     authorize! :search, Medium.new
+
+    # get all media, then set them to only those that are visible to the current user
+    if search_params[:access].blank?
```
fosterfarrell9, line 216: If generic users manipulate their search form and add some non-blank entry for the access parameter, they will see unfiltered results, since the filtering is triggered by `filter_media`, see line 240. This should be prevented. I think they should always be filtered for non-gneric users.

fosterfarrell9 (changes requested):
One other thing I noticed is that when you hit the back button of the browser after having performed a media search, you are shown the json data of the last search. This does not happen in the media search in the admin area.

PR #422, opened 2022-12-15
Updates ws202223

zebleck:

* **Other information**:


fosterfarrell9 (changes requested):
I noticed two things aside from the changes:

- We should synchronize the postgres versions: postgresql-client-13 in the Dockerfile and postgres:13 in docker-compose
- Sidekiq requires a minimal version of  redis (6.2.6). In my development environment this resulted in an error because my local redis image was older:  `You are connecting to Redis 6.0.5, Sidekiq requires Redis 6.2.0 or greater`. I suggest we add  a version for redis in the docker-compose file for development , e.g. `redis:6.2.6-alpine`.

fosterfarrell9:
Closed in favor of #443.

PR #429, opened 2023-02-28
fix the autosetup

christian-heusel:
**To be discussed:** Shoud the env vars `DB_SQL_PRESEED_URL` and `UPLOADS_PRESEED_URL` be present by default so that mampf automatically populates its database?

---

Bug fix

- [ ] ~~E2E Tests for the changes have been added  via Cypress~~
- [ ] ~~Meaningful rspec tests have been added~~
- [x] Docs have been added / updated

Currently the autosetup was broken due to out of date links.

The autosetup works again and the docs read a little nicer

Nope

* **Other information**:

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,17 +25,21 @@ You now have the following things ready:
 
 ### Database setup
 
-#### Use an empty database
+#### Using a prefilled database
 
-1. register via <a href="http://localhost:3000/users/sign_up?" target="_blank">MaMpf-GUI</a>
-2. visit the <a href="http://localhost:1080/" target="_blank">mailcatcher webinterface</a> and confirm your account
-3. make your account admin
-   ```
-   # docker-compose exec mampf rails r "User.first.update(admin: true)"
-   ```
-4. There you go :tada:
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -25,17 +25,21 @@ You now have the following things ready:
...
+- `admin@mampf.edu`
+- `teacher@mampf.edu`
+- `tutor@mampf.edu`
+- `student1@mampf.edu`,..., `student5@mampf.edu`
+
+##### Prefilled database: automatic setup
 
-#### Use a prefilled database
+Just uncomment the lines containing `DB_SQL_PRESEED_URL` and `UPLOADS_PRESEED_URL` in the development docker-compose.yml, start Mampf and you should be good to go after it finished initialising.
+If that for some reason does not work, open a new issue about it (https://github.com/MaMpf-HD/mampf/issues/new) and follow the manual setup instead.
```
fosterfarrell9, line 40: I do not know if it is clear to all users that if would be best to restore the comment after the first initialisation has been done. Otherwise it will autopopulated each time the container is built, which might not be what is wanted.
Maybe add some short info on this?

fosterfarrell9 (changes requested):
I think it makes sense to switch to autopopulation of the database.

Issue #435, opened 2023-03-13
Linter run fails because of old ruby version


christian-heusel:
See for example [🛠 Job #7651761131](https://github.com/MaMpf-HD/mampf/actions/runs/4373504057/jobs/7651761131)

> Run bin/bundle config path vendor/bundle
> Your Ruby version is 3.0.3, but your Gemfile specified 3.1.2
> Error: Process completed with exit code 18.

Referencing the following code snippets:

https://github.com/MaMpf-HD/mampf/blob/8b5b7baadd6083503ffec1a8713f0f99f4f8d541/Gemfile#L4

https://github.com/MaMpf-HD/mampf/blob/8b5b7baadd6083503ffec1a8713f0f99f4f8d541/.github/workflows/linter.yml#L8-L17

It should be an easy fix, but I dont know for sure😄 🤷🏻‍♂️

PR #439, opened 2023-03-21
Linter CI fix

christian-heusel:
* **What kind of change does this PR introduce?**
Bug fix for Issue #435

Linter run fails

Linter run works

No

* **Other information**:
🚫

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -5,62 +5,62 @@ jobs:
     runs-on: ubuntu-latest
     name: A job to check rubocop linter errors
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Ruby 3
-      uses: ruby/setup-ruby@v1
-      with:
-        ruby-version: 3.1.2
-    - name: Install gems # usual step to install the gems.
-      run: |
-        bin/bundle config path vendor/bundle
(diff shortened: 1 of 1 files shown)


Splines:
@christian-heusel Looks good to me, although I'm a bit worried about the result:

> 418 files inspected, 3739 offenses detected, 2974 offenses auto-correctable

But this is since we only check if there are *more* linter errors compared to `mampf-next`, isn't it?
Why don't we compare the number in the branch we are merging into (which does not have to be `mampf-next` necessarily, right?)

christian-heusel:
The change reduces to the following if you exclude whitespace, which makes https://github.com/MaMpf-HD/mampf/pull/439/commits/7b27ceab9c1bd480add5c11b17a17e8fe8bf074d the only commit that actually changes something 😊 
```diff
-      ruby-version: 3.0.3
+          ruby-version: 3.1.2
```

> I'm just a bit worried about the output, e.g. in your run:
> [...]
> But this is since we only check if there are _more_ linter errors compared to `mampf-next`, isn't it? Why don't we compare the number in the branch we are merging into (which does not have to be `mampf-next` necessarily, right?)

I have no clue about how all of this works 🤷🏻  While the change you are suggesting sounds reasonable it needs to be done in a separate PR (and not by me) so feel free to open a issue about this 👍🏻

PR #311, opened 2021-06-06
Docs

henrixapp:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

* **Please check if the PR fulfills these requirements**

- [ ] E2E Tests for the changes have been added  via Cypress
- [ ] Meaningful rspec tests have been added
- [ ] Docs have been added / updated 

* **What is the current behavior?** (You can also link to an open issue here)

* **What is the new behavior (if this is a feature change)?**

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

* **Other information**:

PR #441, opened 2023-03-31
CI/CD Fix unit test workflow

Splines:
**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Old:
Improve the unit test GitHub Actions workflow and get it to run without failures. To do so, I temporarily disabled failing unit tests. These should be fixed according to issue #442 

Edit:
This PR will fix the unit test workflow. Other than originally stated, it will *not* disable any unit tests.


Splines:
Hi @henrixapp, what is the "reindex sunspot" task used for, which you've introduced [here](https://github.com/MaMpf-HD/mampf/commit/4edef8fb6db0b78874cb87759cbca863daffc346)? Why do we need it?

```
docker-compose exec mampf  sh -c "RAILS_ENV=test  rake sunspot:reindex" 
```

Furthermore, I've just noticed that I replaced `docker compose run` with `docker compose exec`.

> docker-compose exec will run inside your existing, running service container while docker-compose run will start a new, independent container.
[from here](https://stackoverflow.com/a/65110945/)

Based on this, `exec` sounded better to me and it seems to work. However, you explicitly used `run` instead of `exec` starting from [here](https://github.com/MaMpf-HD/mampf/commit/c2690bb0eb45e07c7e39b7e7da320e0cc09f71f5). Is there a reason for that?

Splines:
I think it should be ok to merge this directly to `main` instead of `mampf-next`, so that other branches can directly profit from this. This does not touch on any code and instead just affects the CI/CD pipeline (which is currently not working correctly anyways ;))

henrixapp:
Hi @Splines,

> what is the "reindex sunspot" task used for, which you've introduced [here](https://github.com/MaMpf-HD/mampf/commit/4edef8fb6db0b78874cb87759cbca863daffc346)? Why do we need it?

Sunspot is used for indexing the data and providing search results. The reindexing has to be triggered to index the test data, ingested in the step before.

> Based on this, exec sounded better to me and it seems to work (I use docker compose up --build --detach now in a previous step and then use docker compose exec to execute the database related commands in the new mampf container). However, you explicitly used run instead of exec starting from [here](https://github.com/MaMpf-HD/mampf/commit/c2690bb0eb45e07c7e39b7e7da320e0cc09f71f5). Is there a reason for that? (maybe even indicating I should refrain from exec in this case?)

I can not remember this directly. As the operations are altering the volumes/data of the database, I found it more convenient
 to spin them up via `run`, but your way of doing it probably works fine.

> I think it should be ok to merge this directly to main instead of mampf-next, so that other branches can directly profit from this. This does not touch on any code and instead just affects the CI/CD pipeline (which is currently not working correctly anyways ;))

I would suggest keeping the direction of merging clean, because otherwise you will have to sync between `main` and `mampf-next` at a later point. Furthermore, as you can see by the pull request #205, my  changes between run and exec were in one pull request. It might be worth considering squashing commits, keeping the history cleaner and more understandable. Especially, #311 is a candidate for squashing.

Finally, let's take a look on what you try to achieve: You want the tests to be passing, to get that ✅ .

You commented out the tests that are failing, achieving your goal, but I would suggest going a different route, fixing the setup/environment error.

 If we take a look at the [logs](https://github.com/MaMpf-HD/mampf/actions/runs/4565030758/jobs/8055572883) of a failing job, we see that the unit test fails due to a `webpacker` error:

```
     # ------------------
1428
     # --- Caused by: ---
1429
     # Webpacker::Manifest::MissingEntryError:
1430
     #   Webpacker can't find application.js in /usr/src/app/public/packs-test/manifest.json. Possible causes:
1431
     #   1. You want to set webpacker.yml value of compile to true for your environment
1432
     #      unless you are using the `webpack -w` or the webpack-dev-server.
1433
     #   2. webpack has not yet re-run to reflect updates.
1434
     #   3. You have misconfigured Webpacker's config/webpacker.yml file.
1435
     #   4. Your webpack configuration is not creating a manifest.
1436
     #   Your manifest contains:
1437
     #   {
1438
     #   }
1439
     #   /usr/local/bundle/gems/webpacker-5.4.3/lib/webpacker/manifest.rb:79:in `handle_missing_entry'
```

So apparently we do not build the manifest file.
 This can be fixed by adding a precompile directive to the steps while building.

```
docker compose run --entrypoint="" mampf  sh -c "RAILS_ENV=test  rake assets:precompile"
```

I use https://github.com/nektos/act to run actions locally and the tests passed on `main` with your updated file.

Splines:
@henrixapp 
> I would suggest keeping the direction of merging clean, because otherwise you will have to sync between main and mampf-next at a later point.

True, fair enough ;)

> It might be worth considering squashing commits, keeping the history cleaner and more understandable.

Yes, especially for the "release" branch `mampf-next`, we could squash commits from the individual feature branches. Then, merge `mampf-next` to the `main` branch with a merge commit.

> I would suggest going a different route, fixing the setup/environment error.

Thanks for also giving me the fix, will try this out and commit. I overlooked this and thought that some tests were just not correct.

> I use https://github.com/nektos/act to run actions locally

Wow, didn't know about that project, thank you, will try it out. This should render ugly force pushes (just to trigger the pipeline) unnecessary.

Splines:
@henrixapp Ok, so it works, thanks for your fix. Ready for review now ;)

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Build and start docker containers
+      working-directory: docker/development
+      run: |
+        docker compose up --build --detach
+
+    - name: Create and prepare test database
+      working-directory: docker/development
+      run: |
+          docker compose exec mampf sh -c "RAILS_ENV=test rake assets:precompile"
```
henrixapp, line 44: This  does not match the description "Create and prepare test database". Maybe merge this step and the reindex step and rename it. you can squash multiple steps like `db:create db:migrate` into one line.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,63 @@
...
+    # It also restores the cache if it exists.
+    # https://github.com/jpribyl/action-docker-layer-caching
+    - name: Use Docker layer caching
+      uses: jpribyl/action-docker-layer-caching@v0.1.1
+      continue-on-error: true
+
+    - name: Build and start docker containers
+      working-directory: docker/development
+      run: |
+        docker compose up --build --detach
```
henrixapp, line 39: I would not use `up --detach`, rather just build it and run with an empty entrypoint.

This has the advantage, that you don't have to expect GitHub Actions to be stateful (in keeping the process running) and each step is clearly terminating.

Splines:
@henrixapp Apparently, removing webpacker from the dev dependencies of the `package.json` does not work here but in your pull request #453. Do you know why?

henrixapp:
> Apparently, removing webpacker from the dev dependencies of the package.json does not work here but in your pull request https://github.com/MaMpf-HD/mampf/pull/453. Do you know why? Maybe GitHub has cached Webpacker for you?

You are using the `docker-compose.yml` file/setup in docker/development and try to compile with the running container. The running container is mounting the whole `mampf`  root folder [here](https://github.com/MaMpf-HD/mampf/blob/03493db086b2166464a5a4a035d5da2a460de1d5/docker/development/docker-compose.yml#L100), overriding any `node_modules` folder with it. Maybe the docker-compose setup in run_tests should be used.

I would suggest

Splines:
Also see https://github.com/rails/webpacker/issues/522

henrixapp:
@Splines the problem is that the Dockerfile in run_tests does not install the js with `yarn` dependencies. Forgot to mention that you need to update the Dockerfile properly.

Splines:
[This SO question](https://stackoverflow.com/questions/59161189/rails-webpacker-4-2-cant-find-application-in-app-public-packs-manifest-json-h) might be relevant as it's discussing the same error we get currently: `can't find application.js in /usr/src/app/public/packs-test/manifest.json`

```
# --- Caused by: ---
# Webpacker::Manifest::MissingEntryError:
#   Webpacker can't find application.js in /usr/src/app/public/packs-test/manifest.json. Possible causes:
#   1. You want to set webpacker.yml value of compile to true for your environment
#      unless you are using the `webpack -w` or the webpack-dev-server.
#   2. webpack has not yet re-run to reflect updates.
#   3. You have misconfigured Webpacker's config/webpacker.yml file.
#   4. Your webpack configuration is not creating a manifest.
#   Your manifest contains:
#   {
#   }
#   /usr/local/bundle/gems/webpacker-5.4.4/lib/webpacker/manifest.rb:79:in `handle_missing_entry'
```

Splines:
[This](https://github.com/rails/webpacker/issues/2825#issuecomment-764888322) could also be related. Will try this in the next commit.

Splines:
@henrixapp @christian-heusel Ok, I think I'm a bit stumped on this one (@christian-heusel I'm mentioning you as git blame tells me you worked on the Dockerfiles I'm currently working (despairing) on. Somehow, `webpacker` is not properly working in the test environment and I can't figure out why, probably because I seem to miss the bigger picture of how exactly the whole pipeline works. [This](https://github.com/MaMpf-HD/mampf/pull/441#issuecomment-1517787355) is still the error we get in the CI GitHub environment.

- In the development docker compose file, we have a separate `webpacker` service section which is probably starting webpacker and let it compile all frontend files. Why don't we have this section in the docker compose file for `run_tests` as well? Where is webpacker supposed to be started there?

- @henrixapp [Here](https://github.com/MaMpf-HD/mampf/blob/mampf-next/docker/run_cypress_tests/Dockerfile#L27-L35), you use `COPY ...` and `RUN bundle install` two times consecutively (lines 27/28 and 33-35), is this relevant to the current issue?

- @henrixapp I cannot find any changes you made in  #216 specifically related to Webpacker, except for maybe an additional `yarn install` in the Dockerfile, but that didn't work for me. Where is webpacker installed and where do you run the command to actually go off compiling all the frontend .js and .scss files? What is the purpose of `yarn install --production=false` if we don't have any dev dependencies declared in the `package.json`?

I think a quick and dirty solution for this would be fine, as we will eventually replace webpacker anyways, see #454.

henrixapp:
Closed in favor of #472

PR #444, opened 2023-04-01
Redesign landing page

### LINKED ISSUE #231: Usability
Rethinking the UI and taking the perspective of an outsider might be helpful. Esp. the starting page could be revamped a little bit.

Splines:
Create a landing page that includes login options. Closes #231 

image

<details>
<summary>With announcements</summary>

image
</details>


config/routes.rb
```diff
@@ -850,7 +850,7 @@
 
   # main routes
 
-  root 'main#home'
+  root to: redirect("/users/sign_in")
```
fosterfarrell9, line 853: Currently, we have a lot of `redirect_to :root` statements in our codebase when errors are occuring (see e.g. in the lectures controller). In case :root is changed to the sign in page, :root in these statements would have to be replaced by something else, because you would not want to redirect a user that is already logged in to the login page in that case (rather to main/start; while users that are not logged in can be redirected to the new :root). Also, the current implementation seems to have an infinite loop problem: Try a path like lectures/4978593 and you get a redirection cascade.

fosterfarrell9 (changes requested):
I like the new design. It looks fresher. Only the placement of the link for the locale still seems a little off to me.

config/routes.rb
Splines: Fixed this now. I direct the user to `main#start`, so that `main#home` is never reached anymore and could be deleted in the future since it does not give any added value (because the welcome and about text was moved to the landing page which now encompasses the greeting as well as the login/signup).

Splines: Also restyled the "language switcher".

app/assets/images/landing-background.jpg
fosterfarrell9, line 1: There is this thin rectangle in the lower right quadrant of the image which kind of disturbs the pattern.

config/routes.rb
```diff
@@ -848,9 +848,27 @@
...
+	
+		unauthenticated do
+			root to: 'devise/sessions#new', as: :unauthenticated_root
+		end
+	end
+
+	# Allow /login besides /users/sign_in
+	devise_scope :user do
+    get "/login" => "devise/sessions#new"
+  end
```
fosterfarrell9, line 872: I like that solution.

app/views/layouts/devise.html.erb
fosterfarrell9, line 1: This layout does not have a place for announcements like the old landing page had (you can create them as admin under `/announcements` and put them on the main page). It would be nice if the new landing page had a place for that, otherwise unauthenticated users will not see the announcements.

fosterfarrell9 (changes requested):

app/assets/images/landing-background.jpg
Splines: Oh yeah, that was probably due to Photoshop having too little RAM available when I was exporting the image 😅 Will fix this.

app/controllers/sessions_controller.rb
```diff
@@ -1,5 +1,13 @@
 class SessionsController < Devise::SessionsController
 
+  def initialize
+    @announcements = Announcement
+      .where(on_main_page: true, lecture: nil)
+      .pluck(:details)
+      .join
+    super
+  end
+
```
Splines, line 10: I added the announcements to the landing page view (right now called `devise.html.erb`). The problem is that when I set the `@announcements` variable here in the `SessinosController`, announcements will only show up for the `/users/sign_in` page, not even the landing page. So, I could just go ahead and add this to the devise registration controller and confirmation controller, but this seems repetitive (not DRY), and this won't solve the problem that it doesn't work for the landing page that gets mapped to the `devise/sessions#new` controller in the `unauthenticated` block in `routes.rb`.

What we would need is a controller that handles the landing page on its own. The corresponding view should have the devise layout as nested layout. However, I don't know how to do that, especially with the controller, and I can't figure this out after endless searching. These might be some links to consider:

- [devise `authenticate` method](https://github.com/heartcombo/devise/blob/3926e6d9eb139cc839faec8ea6c8f8cefa2d95f6/lib/devise/rails/routes.rb#L296-L334)
- [devise configuring controllers](https://github.com/heartcombo/devise#configuring-controllers)
- [Devise: Sign Up Page as Welcome/Landing Page then to User Profile](https://stackoverflow.com/questions/5825135/devise-sign-up-page-as-welcome-landing-page-then-to-user-profile)]
- [How to render a view inside of another view (rails 4)](https://stackoverflow.com/questions/19844652/how-to-render-a-view-inside-of-another-view-rails-4)

Splines: Thanks, that works ;)

Splines:
@fosterfarrell9 Do we have a `.env` file for testing purposes? For example, the link to the blog is supposed to be located in such a file (see `default_setting.rb`). That's why on localhost, I get a link to `localhost:3000` instead of `https://mampf.blog`.

fosterfarrell9:
> @fosterfarrell9 Do we have a `.env` file for testing purposes? For example, the link to the blog is supposed to be located in such a file (see `default_setting.rb`). That's why on localhost, I get a link to `localhost:3000` instead of `https://mampf.blog`.

This is a bug. The environment variables are set in `docker/development/docker-compose.yml`, but `BLOG_LINK` is missing for some reason.

app/views/layouts/devise.html.erb
```diff
@@ -21,12 +67,31 @@
...
+          <%# Actual login form %>
+          <div class="pt-4">
+            <%= yield %>
+          </div>
+        </div>
       </div>
+
+      <%# Announcements %>
+      <% announcements = get_announcements() %>
+      <% if announcements.present? %>
```
fosterfarrell9, line 80: In case you implement the scope mentioned above, you would not need the announcements variable and could do
`if Announcement.active_on_main.any?`

app/assets/images/landing-background.jpg
fosterfarrell9, line 1: It seems that the corrupt rectangle has moved upwards.

app/views/layouts/devise.html.erb
```diff
@@ -21,12 +67,31 @@
...
+        <div id="announcement-box" class="landing-box">
+          <i class="bi bi-megaphone p-1"></i>
+          <%= t('announcements')%>
+          <hr>
+          <%= get_announcements().html_safe %>
+        </div>
+      <% end %>
```
fosterfarrell9, line 87: I would personally prefer the announcement box above the login box, but maybe we should leave this discusion to basecamp.

fosterfarrell9 (changes requested):

app/helpers/application_helper.rb
```diff
@@ -283,4 +283,12 @@ def first_course_independent?
...
                &.first&.term_independent
   end
+
+  def get_announcements
+    return Announcement
+      .where(on_main_page: true, lecture: nil)
+      .pluck(:details)
+      .join
+  end
+
```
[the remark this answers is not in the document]
Splines: Thanks, pretty nice concept, these scopes. I've just implemented your suggestion. However, I left the `return` statement in there as it does not do any harm, right? So, it's just more explicit, adds semantic meaning to the code, which is therefore more readable. See also [this discussion](https://stackoverflow.com/questions/1023146/is-it-good-style-to-explicitly-return-in-ruby).

app/views/layouts/devise.html.erb
Splines: Just found [this](https://www.ombulabs.com/blog/benchmark/performance/rails/present-vs-any-vs-exists.html) concerning `.present?` vs. `.any?` vs. `.exists?` and therefore used `.exists?`.

app/assets/images/landing-background.jpg
Splines: Ok, this is really strange, those bars were definitely not there when I put the image there. Will investigate this problem, maybe has sth to do with Git or the upload to GitHub...

app/views/layouts/devise.html.erb
Splines: Ok, still moved it up for now as these announcements might be easily overlooked when they are at the bottom of the page. But of course, we can further discuss that on Basecamp ;)

app/helpers/application_helper.rb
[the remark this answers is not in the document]
fosterfarrell9: No, it does not do any harm. Personally, I do not care whether it is used or not. Only my old buddy, the rubocop, won't like it :-) In the codebase, you will not see many `return`s when they can be omitted.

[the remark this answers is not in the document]
Splines: Ok, yeah I still have to install rubocop to format the code ;) Will do this in a next step alongside adjusting the pipeline, so you won't be able to push any code that was incorrectly formatted anymore (otherwise it's really tedious to have different codestyles and also different indentations, e.g. spaces vs. tabs). But still then, rubocop may like a `return` statement if we tell him to do so in the config 😅

fosterfarrell9 (approved):

PR #446, opened 2023-04-03
Redesign left navbar

Splines:
**Redesign of the left navbar.**
⚠ Merge this PR into `mampf-next` before merging #451 into `mampf-next`.

## Features

- Stick footer to the bottom
- Use hover effects for menu items
- Use bread-crumb like site navigation visual indicator (current site you are viewing is now colored in the sidebar)
- Add icons to the buttons
- Reorganize/Regroup the items that belong together semantically
- Navbar is responsive (future PRs could make the navbar collapsible in the mobile view)


christian-heusel:
Looks pretty nice! 😊 

I am just wondering why you used normal css...? Is this something that cannot be achieved with the predefined bootstrap classes? 🤔

Splines:
> Looks pretty nice! 😊
> 
> I am just wondering why you used normal css...? Is this something that cannot be achieved with the predefined bootstrap classes? 🤔

Thanks ;)

I do use the usual bootstrap classes in most cases. However, some elements require more subtleties than bootstrap can offer or just require lots of styling, so that it would be cluttered to write all the classes next to each other in a big line in the html file. Also, as far as I know, with bootstrap classes we cannot handle styling of multiple elements at the same time, e.g. style all anchor tags or style all list items that are located in an unordered list with class name ... etc. With css selectors, we have more flexibility here.

TL;DR: I use Bootstrap classes for simple tasks and oneliners (e.g. `<div class="col-lg-7"></div>`), but Sass CSS styling for more complex tasks.

app/views/shared/_sidebar.html.erb
```diff
@@ -1,195 +1,208 @@
...
-      <% else %>
-        <span class="nav-link disabled">
-          <%= t('categories.organizational_html') %>
-        </span>
-      <% end %>
+    <%# Divider %>
+    <span class="border-top my-3"></span>
+
+    <%# Forum %>
+    <% path = "/forum/" + lecture.forum.slug %>
```
Splines, line 132: One error we get right now:

Log in as teacher and click on helpdesk. This will show up:



Apparently, not every lecture forum model has a slug. What would be the best option here to avoid this error? Test if slug is present and if not disable the forum button?

fosterfarrell9: In production, there are a lot of lectures that do not have forum (some teachers just do not want them). So, it should be tested whether there is a forum using `lecture.forum?`.

Splines:
Current look

Splines:
New item organization:

app/helpers/application_helper.rb
```diff
@@ -288,4 +288,25 @@ def get_announcements
...
+
+  def get_class_for_any_path(paths)
+    for path in paths do
+      if request.path == path
+        return $active_css_class
+      end
+    end
+    return ''
+  end
+
```
fosterfarrell9, line 311: I would suggest refactoring this method as

```
def get_class_for_any_path(paths)
  return '' unless paths.include?(request.path)
  return $active_css_class
end
```
or
```
def get_class_for_any_path(paths)
  return paths.include?(request.path) ? $active_css_class : ''
end
```

fosterfarrell9 (approved):

app/helpers/application_helper.rb
Splines: Thanks, `.include?` is indeed also more readable ;) I went for your second option.

Issue #448, opened 2023-04-07
User overview misses edited media in some cases


fosterfarrell9:
In the administration mode of mampf under `users/:id/edit` you can see given the user's lectures, edited lectures and courses and edited media. However, if the user has none of the first three, the edited media will not be displayed. The reason is a missing case in an `if` statement in the corresponding view.

fosterfarrell9:
Closed via #449

PR #449, opened 2023-04-07
fix missing media in user edit view

fosterfarrell9:
This PR adds a missing case in an `if` statement in the user edit view and closes issue #448.

### DIFF
--- a/app/views/users/_data.html.erb
+++ b/app/views/users/_data.html.erb
@@ -44,7 +44,7 @@
       </div>
       <div class="card-body">
         <% if user.edited_courses.any? || user.edited_lectures.any? ||
-              user.given_lectures.any? %>
+              user.given_lectures.any? || user.edited_media.any? %>
           <%= render partial: 'users/content',
                      locals: { user: user,
                                f: f } %>


app/views/users/_data.html.erb
```diff
@@ -44,7 +44,7 @@
       </div>
       <div class="card-body">
         <% if user.edited_courses.any? || user.edited_lectures.any? ||
-              user.given_lectures.any? %>
+              user.given_lectures.any? || user.edited_media.any? %>
```
Splines, line 47: `.exists?` should be a bit faster, see [here](https://stackoverflow.com/a/43774445/9655481)

Splines (changes requested):

app/views/users/_data.html.erb
Splines: > In Rails 5.1 and up, the performance of any? and exists? has been fixed, so they are now equally performant.
~ from [here](https://www.ombulabs.com/blog/benchmark/performance/rails/present-vs-any-vs-exists.html)

Ah no, I missed that, so `.any?` should also be fine

Splines (approved):

PR #451, opened 2023-04-07
Redesign top navbar

Splines:
**Redesign of the top navbar**.
⚠ This PR should be merged into `mampf-next` (not `feature/left-navbar-design`) _after_ having merged #446 into `mampf-next`.

## Features

- Add background image to navbar (responsive)
- Add shadow to navbar
- Use SVG version of MaMpf logo

## Preview
image


christian-heusel (changes requested):
In my opinion the Image background for the top navbar is not actually an improvement over the solid one. 
Sorry, but currently thats a NACK from me.

Splines:
@christian-heusel True, it might even distract with the image. So, what do you think of this version:
image

christian-heusel (approved):
I think that looks better, so I approve the changes!
I dont think I can judge this PR from a technical standpoint 🤷🏻 

Edit: but since @fosterfarrell9 is also a reviewer I guess he will have some more input wrt implementation ;)

Splines:
Current look

image

app/helpers/application_helper.rb
```diff
@@ -309,4 +313,13 @@ def get_class_for_any_path(paths)
...
 
+  def get_class_for_any_path_startswith(paths)
+    for path in paths do
+      if request.path.starts_with?(path)
+        return $active_css_class
+      end
+    end
+    return ''
+  end
+
```
fosterfarrell9, line 324: I would suggest refactoring this into something more rubyesque like
```
def get_class_for_any_path_startswith(paths)
  if paths.any? { |path| request.path.starts_with?(path) }
    return $active_css_class
  end
  return ''
end
```

app/helpers/application_helper.rb
```diff
@@ -309,4 +313,13 @@ def get_class_for_any_path(paths)
     return ''
   end
```
fosterfarrell9, line 315: For a possible refactorisation of this method see PR #446.

fosterfarrell9 (approved):

app/helpers/application_helper.rb
Splines: Yeah, this is similar to Python list comprehension, didn't know Ruby had that ;)

app/assets/stylesheets/navbar.scss
```diff
@@ -0,0 +1,43 @@
...
+            position: absolute;
+            top: 21px;
+            left: 16px;
+        }
+    }
+}
+
+.new-comment {
+    color: #ffc3d0 !important;
+}
```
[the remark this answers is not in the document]
Splines: looks good to me on the blue background 👍

PR #452, opened 2023-04-10
Restrict editor access

fosterfarrell9:
This PR makes some changes to access rights for users who have edited some media, but are not teachers or editors of lectures or courses.


app/abilities/lecture_ability.rb
```diff
@@ -33,7 +33,7 @@ def initialize(user)
     end
 
     can :subscribe_page, Lecture do |lecture|
-      lecture.published? || !user.generic?
+      lecture.published? || user.active_teachable_editor?
```
Splines, line 36: Why can only active teachable editors subscribe to a lecture?

Splines: ah I see now, you're shortcircuiting here ;) If a lecture is unpublished, then only users with the respective rights can subscribe to it. Ok, so this is fine.

app/models/user.rb
```diff
@@ -293,10 +293,34 @@ def teacher?
...
+  # but just an additional editor of some lecture, you
+  # will not be considered active if all of your edited lectures
+  # are older than one year
+  def active_teachable_editor?
+    return false unless can_edit_teachables?
+    return true if admin || course_editor? || teacher?
+
+    edited_lectures.select { |l| l.term.nil? || !l.older_than?(1.year)}
+                   .any?
+  end
```
Splines, line 318: I don't get the wording `active_teachable_editor` here. "Teachable" means "able to learn" (lernfähig) and I don't understand what it means in this context. It rather sounds like it tries to describe the students themselves; them being hopefully "teachable".

You define a teachable editor to be a user that has edited any course _or_ has edited any lecture (see line 297). For me, that does not sound related to a teacher, but rather just like an "editor" of a lecture or a course (Veranstaltungseditor:in).

According to line 324, the only difference between a "teachable_editor" and a normal "editor" is that an editor might be someone that has edited any media other than "Talks". So what is the difference then really? How can someone edit media without having editor rights (= without being a "Zusätzlicher Veranstaltungseditor:in")?

app/models/user.rb
Splines, line 1:
```
  # a user has many lectures as an editor
  has_many :edited_lectures, through: :editable_user_joins,
                             source: :editable, source_type: 'Lecture'
```
Does this only refer to "edited" lectures in the past or also to lectures that a user has edit rights for, but hasn't edited anything yet?

app/models/user.rb
```diff
@@ -691,6 +715,17 @@ def generic?
...
 
+  # for lectures that are older than one year, only the teacher or an editor
+  # of the course it belongs to can view users in forms related to the lecture
+  def can_view_users?(lecture)
+    return false unless can_edit?(lecture)
+    return true if can_edit?(lecture.course) || lecture.teacher == self
+    return true if lecture.course.term_independent
+    return true if !lecture.older_than?(1.year)
+
+    false
```
Splines, line 726: might wanna use `return` here as well as it is used in the previous lines beforehand

app/models/user.rb
```diff
@@ -691,6 +715,17 @@ def generic?
     !(admin? || teacher? || editor?)
   end
 
+  # for lectures that are older than one year, only the teacher or an editor
+  # of the course it belongs to can view users in forms related to the lecture
+  def can_view_users?(lecture)
+    return false unless can_edit?(lecture)
+    return true if can_edit?(lecture.course) || lecture.teacher == self
+    return true if lecture.course.term_independent
+    return true if !lecture.older_than?(1.year)
```
Splines, line 724: This condition `lecture.older_than?(1.year)` is used in multiple places, so it's probably a good idea to outsource it, e.g. write a short method `is_lecture_stale()`.

config/locales/de.yml
```diff
@@ -2053,7 +2058,7 @@ de:
     data_request: >
       Daten anfordern
     data_request_sent: >
-      Wir haben deine Anfrage erhalten und werden dir deine Daten zeitnah schicken. 
+      Wir haben deine Anfrage erhalten und werden dir deine Daten zeitnah schicken.
```
Splines, line 2061: "Du" instead of "du" to stay consistent.

config/locales/en.yml
```diff
@@ -618,6 +618,11 @@ en:
       insert_after_chapter: 'Insert after %{chapter_name}'
       insert_after_section: 'Insert after section'
       subscribers: 'Subscribers'
+      no_access_to_users_html: >
+        This lecture is older than 1 year. For reasons of data privacy,
+        only the teahcer and course editors can make changes here.
```
Splines, line 623: "teacher"

config/locales/en.yml
```diff
@@ -1080,7 +1085,8 @@ en:
         editors: >
           This is the list of all persons that are responsible for this
           medium as editor. At least one person has to be given here, and each
-          of these persons gets editing access to this medium
+          of these persons gets editing access to this medium. Only eligible
```
Splines, line 1088: "Persons" sounds archaic. Use "people" instead. E.g. refer to Cambridge dictionary:

> Persons (plural) is a very formal word. We only use it in rather legalistic contexts:
> Any person or persons found in possession of illegal substances will be prosecuted.

config/routes.rb
```diff
@@ -775,10 +775,6 @@
       to: 'users#fill_user_select',
       as: 'fill_user_select'
 
-  get 'users/list',
-      to: 'users#list',
-      as: 'list_users'
-
```
Splines, line 781: What does this have to do with this PR?

Splines (changes requested):
Just went through the code, but will have to do another more in-depth review later on. Here some first remarks.

app/models/user.rb
fosterfarrell9: "Teachable" is an artificial word that we created to describe things that can be taught. I will use it in the following explanation. Usually, polymorphic relations in rails are ending with -able see e.g. [here](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations). A teachable in general can be a course/lecture/lesson/talk. Each medium belongs to a teachable via a polymorphic relation. Among these teachables, only courses/lectures/talks have associated editors (in a relation sense). Lessons can be edited by all editors of the corresponing lecture (and users with higher rights, like the course editor of the course the lecture is in). Also, each medium has an editor. A `teachable_editor` in the sense of line 297 can be a course editor (=Modulverantwortlicher) or a lecture editor. An `editor` in the sense of line 322 is a `teachable_editor` or an editor of a medium that is not a talk. We exclude talk editors (Vortragseditoren) for seminars talks here because these are typically students in seminars that are not supposed to have elevated rights: they are supposed to only edit their talk and create media for it. It is possible that someone can edit media without being a course or lecture editor. E.g. when you were such a course/lecture editor at some point in time and created your medium then: you wil keep the editing rights for this medium (but not the lecture) even if you lose the editing rights for the course/lecture. Also, you could be a student who created a medium for a talk in a seminar. This is indeed all a bit messy.

app/models/user.rb
fosterfarrell9: It refers to all edited lectures, past or present..

app/models/user.rb
fosterfarrell9: The ommision of return is  correct according to the [ruby style guide](https://rubystyle.guide/#no-explicit-return) und the standard rubocop settings. But I can change it, sure.

config/routes.rb
fosterfarrell9: This route is out of use so I deleted it. We can put that in another pR if you want.

app/models/user.rb
Splines: Ah ok, I mean, if the ruby style guide says this is best practice, go for it ;) It's probably just me coming from other programming languages that I'm unused to this concept, but it should be fine, especially since this only affects the last line of a method.

config/routes.rb
Splines: Yes, I've seen you already did this [here](https://github.com/MaMpf-HD/mampf/commit/de9e1b102d7c4a53b92644ae778cbbb301b13cbd#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5L778-L781), so no need to have it here.

app/models/user.rb
```diff
@@ -293,10 +293,34 @@ def teacher?
...
+
+  # if you are not a teacher of lecture or a module editor,
+  # but just an additional editor of some lecture, you
+  # will not be considered active if all of your edited lectures
+  # are older than one year
+  def active_teachable_editor?
+    return false unless can_edit_teachables?
+    return true if admin || course_editor? || teacher?
+
+    edited_lectures.select { |l| l.term.nil? || !stale? }
```
Splines, line 316: Why do we select lectures that don't have a `nil` term here as well? What does that mean for a lecture, why is term optional? Is it omitted for lectures that are term independent (`lecture.course.term_independent`)? What does that mean semantically?

app/models/user.rb
```diff
@@ -293,10 +293,34 @@ def teacher?
...
+                   .any?
+  end
+
   # a user is an editor iff he/she is a course editor or lecture editor or
   # editor of media that are not associated to talks
```
Splines, line 321: Comment should not specify what a teachable_editor is. Instead specify it as comment for the `teachable_editor` method.

Splines (changes requested):

app/views/lectures/edit/_people.html.erb
```diff
@@ -19,96 +19,104 @@
...
+                <%= helpdesk(t('admin.lecture.info.teacher'), false) %>
+                <div id="lecture_teacher_select" %>
+                  <%= f.select :teacher_id,
+                               options_for_select([[lecture.teacher.info,
+                                                    lecture.teacher.id]],
+                                                    lecture.teacher.id),
+                               {},
+                               { class: 'selectize',
+                                 data: { ajax: true,
+                                         model: 'user',
```
Splines, line 37: I've seen this line in:
```
fill_path = Routes.fill_user_select_path();
```

in `app/assets/javascripts/_selectize_turbolink_fix.js`. But I can't figure out, how this `_people.html.erb` knows which method on the model `user` to call by just specifying `model: 'user'` here.

Splines (changes requested):
I went through every line of code now. But still have to experiment with the changes on localhost a bit more.

app/models/lecture.rb
```diff
@@ -771,6 +771,20 @@ def import_toc!(imported_lecture, import_sections, import_tags)
...
   end
 
+  def speakers
+    return User.none unless seminar?
+    User.where(id: SpeakerTalkJoin.where(talk: talks).pluck(:speaker_id))
+  end
+
+  def older_than?(timespan)
+    return true unless term
+    term.begin_date < Term.active.begin_date - timespan
```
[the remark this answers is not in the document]
fosterfarrell9: You're right about the `<=` Do you have a suggestion for a different name? I mean, `older_than?` makes sense if one accepts that the age of a lecture is not defined by the usual metric.

app/models/user.rb
fosterfarrell9: I do not quite understand the comment (but I admit that sometimes even I get confused about the different sorts of editors, hence the comment). The comment describes what an editor is, not what a teachable_editor is: You can be an editor (e.g. if you have editing rights to some medium that is not associated to a talk in a seminar) without being a teachable_editor.

app/models/user.rb
fosterfarrell9: There are lectures that do not belong to a term, like the "Helpdesk" (this has several advantages: you do not need to resubscribe as a student in a new term, materials for these should be independent of the term anyway and some others). 
By the curent code, they are considered stale. This is wrong, I think. I think the code in `older_than?` should return false then; we could then remove the `l.term.nil?` condition here (and I would have to look if the logic still fits everywhere).

app/models/user.rb
```diff
@@ -691,6 +715,17 @@ def generic?
     !(admin? || teacher? || editor?)
   end
 
+  # for lectures that are older than one year, only the teacher or an editor
+  # of the course it belongs to can view users in forms related to the lecture
+  def can_view_users?(lecture)
+    return false unless can_edit?(lecture)
+    return true if can_edit?(lecture.course) || lecture.teacher == self
```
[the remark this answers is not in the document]
fosterfarrell9: I am no expert on these finer points, but I have seen `===` rather seldom in Rails applications; in our codebase we do not have a single occurence of it and always use `==.` without problems that are known to me. Of course that is no real argument.

app/views/lectures/edit/_people.html.erb
fosterfarrell9: This is all done in `app/assets/javascripts/_selectize_turbolink_fix.js`. In

`fill_path = Routes.fill_user_select_path();`

the route to the `fill_user_select` action in the users controller is provided in JS by the [js-routes](https://github.com/railsware/js-routes) gem. Then this route is given to TomSelect in 

```
        load: function(query, callback) {
          var url;
          if (send_data || !loaded) {
            url = fill_path + "?course_id=" + courseId + "&q=" + encodeURIComponent(query);
            fetch(url).then(function(response) {
              return response.json();
```
(see e.g. [here](https://tom-select.js.org/examples/remote/)).

app/views/users/list.json.jbuilder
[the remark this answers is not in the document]
fosterfarrell9: It is a view that is no longer in use. It should be removed in different PR.

app/views/media/_basics.html.erb
```diff
@@ -54,30 +54,15 @@
...
-      <% end %>
+      <%= f.select :editor_ids,
+                   options_for_select(medium.eligible_editors(current_user)
+                                            .map { |e| [e.info, e.id] },
+                                      medium.editors.map(&:id)),
+                   {},
+                   { multiple: true,
+                     class: 'selectize',
+                     style: 'display: none;',
+                     data: { ajax: false } } %>
```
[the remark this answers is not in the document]
fosterfarrell9: Generally we have the philosophy (at least until now) to use ajax only when we are dealing with expected larger datasets and retrieve the data directly in Rails otherwise. In this case, the number of eligible editors for a single medium is expected to be rather small (in our production setting, I guess the maximum number will be in the low two digits). That's why the form is already completely populated in the Rails view. If we wanted to use ajax, we would have to write a method similar to fill_user_select but which returns only eligible editors instead of all users.

app/models/user.rb
Splines: True, but right now you also include the definition of a `teachable_editor` in the comment for `editor?` by saying an user is an editor "iff he/she is a course editor or lecture editor or ...". If we changed the definition of `teachable_editor` in the future, we would have to remember: "Oh, we use `teachable_editor` in the `editor?` method as well and must therefore also change the comment over there. I know this is nitpicking and maybe it's just fine to leave it here, but then comments might get stale themselves more quickly.

app/views/media/_basics.html.erb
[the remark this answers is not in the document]
Splines: ah ok, thanks for clarifying

app/models/user.rb
Splines: Ok, but then I think the way it is right now is fine. Otherwise, you include logic in `older_than?` that is not even concerned with specifying is something is older than something else violating the single-responsibility principle.

app/models/lecture.rb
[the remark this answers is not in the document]
Splines: Ok true, we just use a different metric. Maybe just add a comment to `older_than?` making clear what we really compare.

app/models/user.rb
fosterfarrell9: Ah, thank you for the explanation, now I understand. Will change it.

app/models/user.rb
```diff
@@ -293,10 +293,36 @@ def teacher?
...
+
+  # if you are not a teacher of lecture or a module editor,
+  # but just an additional editor of some lecture, you
+  # will not be considered active if all of your edited lectures
+  # are too old
+  def active_teachable_editor?
+    return false unless can_edit_teachables?
+    return true if admin || course_editor? || teacher?
+
+    edited_lectures.select { |l| l.term.nil? || !stale? }
```
Splines, line 318: This fails as there is no `stale?` method on the user. Use `!l.stale?` instead.

Splines (approved):

app/views/lectures/edit/_tutorials.html.erb
```diff
@@ -19,43 +19,48 @@
...
        data-parent="#lectureAccordion">
     <div class="card-body">
-      <div class="row mb-3">
-        <div class="col-12 text-center">
-          <%= link_to t('admin.tutorial.new'),
-                      new_tutorial_path(params: { lecture_id: lecture.id }),
-                      class: 'btn btn-sm btn-primary',
-                      id: 'newTutorialButton',
-                      remote: true %>
+      <% if current_user.can_view_users?(lecture) %>
```
Splines, line 22: You disabled the Tutorials section for non-active teachable editors. However, the corresponding POST route `http://localhost:3000/tutorials` is still left open, so these users can continue to create a new tutorial although they cannot see the section in the UI. This might be the case for other tutorial routes as well. Non-active teachable editors should not be able to access those at all.

app/models/user.rb
Splines: Fixed in commit https://github.com/MaMpf-HD/mampf/pull/452/commits/6e28d5df8cb6e288d4964a7f0625a0fb34cc113d

app/controllers/lectures_controller.rb
```diff
@@ -300,17 +300,19 @@ def check_for_subscribe
...
-                                    :submission_grace_period,
-                                    editor_ids: [])
+    allowed_params = [:term_id, :start_chapter, :absolute_numbering,
+                      :start_section, :organizational, :locale,
+                      :organizational_concept, :muesli, :organizational_on_top,
+                      :disable_teacher_display, :content_mode, :passphrase,
+                      :sort, :comments_disabled, :submission_max_team_size,
+                      :submission_grace_period]
+    if action_name == 'update' && current_user.can_update_personell?(@lecture)
+      allowed_params.concat([:teacher_id, editor_ids: []])
```
Splines, line 310: I don't understand the syntax here. Why is `editor_ids` not preprended by `:` in this case? Is `[]` the default argument for the parameter?

fosterfarrell9: This is because `editor_ids` is supposed to be an Array. Therefore the `editors_ids` key is mapped to `[]`, see [here](https://edgeguides.rubyonrails.org/action_controller_overview.html#permitted-scalar-values).

Splines (dismissed):
Approving despite failing GitHub actions, since this only concerns the post-caching of some docker images while all tests pass successfully.

Confirmed that the tutorium routes are now also inaccessible for "non-active teachable editors".

app/models/medium.rb
```diff
@@ -1263,9 +1262,31 @@ def delete_answers
       becomes(Question).answers.delete_all
     end
 
-  def answers_count
-    return -1 unless type == 'Question'
-    becomes(Question).answers.count
-  end
```
Splines: I'm wondering why this has been a merge conflict. Have you touched this file in this PR? I haven't seen it so far.

fosterfarrell9: I was also very confused why there was a merge conflict here, as I did not touch this in the PR. How do we proceeed?

Splines: Ok, I (probably?) figured it out: run a `git show 202635fa57` on your command line and you will see that most changes came from #495. Git could almost seamlessly merge this. However, maybe your editor indented these specific lines that are changed, so you touched them and my commit on mampf-next touched them, hence, a merge conflict arose. Maybe you could run the layout part of rubocop locally as described in #495 and see if it changes `app/models/medium.rb` again?

fosterfarrell9: Running `rubocop` autocorrection in the `app/` folder only changes one file, namely `lecture.rb`:

```
# git diff
diff --git a/app/models/lecture.rb b/app/models/lecture.rb
index a81bda89..bb7da4ee 100644
--- a/app/models/lecture.rb
+++ b/app/models/lecture.rb
@@ -803,11 +803,13 @@ class Lecture < ApplicationRecord

   def speakers
     return User.none unless seminar?
+
     User.where(id: SpeakerTalkJoin.where(talk: talks).pluck(:speaker_id))
   end

   def older_than?(timespan)
     return true unless term
+
     term.begin_date <= Term.active.begin_date - timespan
   end
```

Maybe I should revert the merge using `git revert HEAD -m 1` (?) so we can see again where the conflict lies?

Splines: Ok, I finally pinpoined this. I just did the following locally:
```
git reset --hard origin/restrict_editor_access`
git reset --merge HEAD~1
```

so that I could replay your merge situation.

```
git checkout mampf-enxt
git pull
git checkout restrict_editor_access
git merge mampf-next
git blame -L 1279,+1 ./models/medium.rb
```

Then I saw the merge conflict in `medium.rb` and saw that you added a space in the very first commit of this PR right before the last `end` of the `medium.rb` file:
```
git blame restrict_editor_access -L 1217,+1 ./app/models/medium.rb
>> fcdcd45da (fosterfarrell9 2023-04-10 14:14:18 +0200 1217)
```

Alternatively, you directly see this in the VS Code merge editor (including `git blame` output right in the file if you have the GitLens extension enabled):
image

So that's where the merge conflict here came from. And the way you merged it is fine, so I approve this PR.


---

Bare in mind, that `git revert HEAD -m 1` might not do what is intended, see the [docs](https://git-scm.com/docs/git-revert#Documentation/git-revert.txt--mparent-number):

> Reverting a merge commit declares that you will never want the tree changes brought in by the merge. As a result, later merges will only bring in tree changes introduced by commits that are not ancestors of the previously reverted merge. This may or may not be what you want.

Also see Linus Torvald's answer [here](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt), although this is not really our situation, so ``git revert HEAD -m 1` might have worked as intended.

Splines (approved):

PR #453, opened 2023-04-10
Pipeline/e2e

henrixapp:
This pull request is to be merged after #441 .

Upgrade cypress to 12 and fix the setup.

CI for e2e tests does not work.

no

no

* **Other information**:


spec/cypress/e2e/courses_spec.cy.js
```diff
@@ -74,8 +74,8 @@ describe("Courses", function () {
                     'user_id':1
                 }]
             ]).then((records) => {
-                cy.server();
-                cy.route('**/new').as('new');
+                // cy.server();
+                // cy.route('**/new').as('new');
```
Splines, line 78: Do we still need this code? If not, why not remove it entirely? We should also use `cy.intercept()` instead of `cy.server()` and `cy.route()`, see [the docs](https://docs.cypress.io/guides/references/migration-guide#cyserver-cyroute-and-CypressServerdefaults).

docker/run_cypress_tests/docker-compose.local.yml
Splines, line 1: What's the purpose of this `docker-compose.local.yml` file? From what I see, the _local_ file just uncomments the cypress video and screenshot volumes, adds two SOLR env variables and defines a different entrypoint.

spec/cypress/support/e2e.js
```diff
@@ -0,0 +1,2 @@
+import './commands'
+import './on-rails'
```
Splines, line 2: Why didn't we need these packages beforehand? What's their use? Where is this new file `e2e.js` imported/used?

Splines:
Could you also refine the name of the workflow/jobs, so that it's better readable in the overview. E.g. for the unit tests, we now get:

image

You can do this by putting a `name: ...` up to the top of the file (here: "Unit tests") and add a `name` to the job (here: "Execute unit tests & upload to Codecov".

---

We should also think of when we need to run this action, e.g. put something along these lines in the beginning of the compose file:

```
name: Unit tests

on:
  push:
    branches:
      - main
      - mampf-next
      - production
      - experimental
  pull_request:
```

docker/run_cypress_tests/docker-compose.yml
```diff
@@ -24,9 +24,6 @@ services:
   mailcatcher:
     restart: on-failure:10
     image: dockage/mailcatcher:latest
-    ports:
-      - "127.0.0.1:1080:1080"
```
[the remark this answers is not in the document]
henrixapp: Removed also all the other unused port forwarding rules. It does not make sense to forward them to the host in CI.

docker/run_cypress_tests/docker-compose.local.yml
henrixapp: To run the tests locally and see the screenshots and videos it produces.

docker/run_cypress_tests/Dockerfile_cypress
```diff
@@ -1,5 +1,5 @@
-FROM cypress/included:7.6.0
+FROM cypress/included:12.9.0
 
 COPY ./spec/cypress/support /cypress/support
 # install the needed support stuff
-RUN cd /cypress/support && yarn install
+RUN apt-get update && apt-get install curl -y
```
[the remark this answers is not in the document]
henrixapp: Yes the file upload helper is now part of cypress. Unfortunately they removed curl, so re added it.

spec/cypress/support/e2e.js
henrixapp: It's automatically used by cypress per their new docs. They were imported differently before.

Splines:
Ah nice, that's even better with no individual job name and only a global name in the docker compose file. Will do this for unit tests as well.

.github/workflows/unit-tests.yml
Splines, line 1: If we merge this PR after #441, you will overwrite the unit test workflow file. So please remove the changes here.

Splines (changes requested):

.github/workflows/unit-tests.yml
henrixapp: These changes, are here because, I currently based them on your branch (and the diff is against mampf-next). Will rebase once #441 is merged.

Splines:
Integration tests are currently timing out with message:

```
waiting for MaMpf to come online at http://mampf:3000
```

henrixapp:
@Splines found the bug, the test adapter env was set to `postgresql:13`, which is not a valid name in `database.yml`, in a commit in mampf-next. I rebased to `mampf-next`. The checkmark for the `e2e` test is now ✅.

I would fix the linter pipeline in a separate pull request.

Splines:
> @Splines found the bug, the test adapter env was set to postgresql:13, which is not a valid name in database.yml, in a commit in mampf-next. I rebased to mampf-next. The checkmark for the e2e test is now ✅.

Ah wow, these are bugs we love, aren't they? 🙈😅 Thanks for fixing this.

> I rebased to mampf-next. 

What do you mean by that exactly? Is that the one force-push you made?

> I would fix the linter pipeline in a separate pull request.

Awesome, had that in mind too, but if you tackle this, even better. I'm not that much of an CI/CD expert.
Ideally, the linter should reject every code that is not 100% to our standards we defined. It's then the task of every user to continuously lint locally.

Splines:
image
Maybe put curl into silent mode again? Or change the sleep to like 2 or even 3 seconds?

henrixapp:
> What do you mean by that exactly? Is that the one force-push you made?

I based the changes on `mampf-next`, not your branch as before, so yes.

>  Or change the sleep to like 2 or even 3 seconds?

Increased.

Splines (approved):

PR #464, opened 2023-04-17
Sync production with mampf-next

Splines:


Splines:
Why is the production branch 41 commits ahead of main? Just wanted to bring back the hotfix to main and then mampf-next, but apparently there's a lot more to bring back 😅

fosterfarrell9:
I was wondering the same thing yesterday😅 . I don't get it.

henrixapp:
Maybe you should first merge with mampf-next and afterwards into main. This way you don't have to care for syncing it separately.

Splines:
@henrixapp Ok, will sync with `mampf-next` first, then with `main` later on.

app/views/shared/_sidebar.html.erb
```diff
@@ -4,15 +4,17 @@
...
-  <% fullpath = lecture_path(lecture)%>
-  <% active_class = get_class_for_path(lecture_path()) %>
-  <li class="sidebar-item nav-link <%= active_class %>
-          <%= !lecture ? "disabled" : ""%>">
-    <a href="<%= url_for(fullpath)%>">
-      <i class="bi bi-house<%= active_class.present? ? "-fill" : "" %>"></i>
-      <%= t('basics.overview') %>
-    </a>
-  </li>
+  <% if lecture %>
```
fosterfarrell9, line 7: Technically, this check for `lecture` (and the other ones in this file) are no longer necessary because the hotfix prepended a check for `lecture` in `app/views/layouts/application.html.erb`. But maybe we want to keep them in for now until we think of a different solution.

fosterfarrell9 (approved):

app/views/shared/_sidebar.html.erb
Splines: Ah ok, yeah, I think we can keep them for now and then replace them once we've come up with a better solution.

PR #471, opened 2023-04-21
Always show module sidebar link

Splines:
Fixes a bug

"Module" sidebar link is disabled when there are no announcements.

Module sidebar link is never disabled (it will always show for every course).

### DIFF
--- a/app/views/shared/_sidebar.html.erb
+++ b/app/views/shared/_sidebar.html.erb
@@ -203,7 +203,7 @@
     <%# Modul %>
     <% path = display_course_path(lecture) %>
     <% active_class = get_class_for_path(path) %>
-    <li class="sidebar-item nav-link <%= active_class %>
+    <li class="sidebar-item nav-link <%= active_class %>">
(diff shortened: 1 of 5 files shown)


Splines:
sorry, branched this off the other feature branch instead of `mampf-next`, will fix this later

app/views/shared/_sidebar.html.erb
```diff
@@ -204,7 +204,6 @@
     <% path = display_course_path(lecture) %>
     <% active_class = get_class_for_path(path) %>
     <li class="sidebar-item nav-link <%= active_class %>
```
fosterfarrell9, line 206: At the end of the line, a `">` is missing, which prevents the link in the next line from being displayed properly.

fosterfarrell9 (changes requested):

app/views/shared/_sidebar.html.erb
Splines: Oh, good catch

fosterfarrell9 (approved):

Issue #445, opened 2023-04-02
Switch back to official release of Commontator gem


fosterfarrell9:
Currently we are using a [fork](https://github.com/MaMpf-HD/commontator) of the [commontator gem](https://github.com/lml/commontator) because of this [issue](https://github.com/lml/commontator/issues/200) in commontator, which prevents that our customized commontator/comments_controller is being used, resulting in issue #389. The change was done in #390. Since then, a workaround has been proposed in [here](https://github.com/lml/commontator/issues/200#issuecomment-1231456146). We could try if it works out for us and if yes, change back to the official release.

PR #475, opened 2023-04-24
Fix decoloring of comments icon

### LINKED ISSUE #474: Decoloring of commentsIcon does not work
If you resolve all comments on the `/main/comments` page by clicking the corresponding button (alternatively, the last remaining comment), the `#commentsIcon` keeps its yellow color instead of becoming white again. The reason is that this is handled in `app/views/readers/update_all.coffee` (resp. `app/views/readers/update_all.coffee`) via 

```
$('#commentsIcon').removeClass('text-primary').addClass('text-light')
```

but the coloring is no longer given by the `text-primary` class in the new navbar. So, the fix should be pretty easy.

Splines:
Fixes #474


Splines:
In addition to `app/views/readers/update_all.coffee` and `app/views/readers/update.coffee`, I also changed these lines in `app/views/comments/create.js.erb`:

```erb
<% if @update_icon %>
$('#commentsIcon').addClass('new-comment');
<% end %>
```

While doing this, I was wondering if we really want the comments icon to be colored at all when you submit your own comment. Isn't it a bit weird for a user to then click on the icon just to notice it was their own comment 3 seconds ago?

fosterfarrell9:
> While doing this, I was wondering if we really want the comments icon to be colored at all when you submit your own comment. Isn't it a bit weird for a user to then click on the icon just to notice it was their own comment 3 seconds ago?

We could discuss this. I mean, for me personally, I like to have this coloring even for my own comments, because it gives me some more visual feedback that I did something that other users will notice. Of course, others might disagree with that view.
By the way, for any changes in that direction, #445 should be resolved first, because the code responsible for selecting who gets a coloring is inside the `comments_controller.rb` in the commontator gem which we currently self-host. It would be better to have this back as a custom controller inside our own repository.

PR #477, opened 2023-04-26
Use main as deployment branch (instead of production)

Splines:
Switch deployment: now from `main` branch, so that we get rid of the `production` branch.


Splines:
Will open a new PR as this one is branched off the wrong branch and it takes too long to look up the commands to change this 😅

PR #480, opened 2023-04-27
Add display text for external medium links

### LINKED ISSUE #479: Different representation of external medium links
For a medium, editors can add an external link which is displayed to users as follows when they click on the details for a medium:

image

An editor noted that a small comment would be very helpful to identify what this link is about instead of just having "Click here" written in the button. A small text field should be sufficient for this, e.g. no full-fledged WYSIWYG editor is necessary to enter the text. The text-box should be added here in the edit menu of a module.
image

We should also consider adding exactly this text as a tooltip when users hover of the link button of the respective medium.
image

Splines:
Closes #479

new feature, see #479 for more details

* **Other information**:

## Preview
image

image

image

### DIFF
--- a/app/views/media/_external_link_card.html.erb
+++ b/app/views/media/_external_link_card.html.erb
@@ -5,9 +5,10 @@
     </h5>
   </div>
   <div class="card-body text-center">
-    <%= link_to t('click_here'),
+    <%= link_to external_link_description_not_empty(medium),
                 medium.external_reference_link,
-                class: "btn btn-info",
+                class: "btn btn-info break-text-any-where",
+                title: medium.external_reference_link,
                 target: :_blank %>
   </div>
 </div>
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -329,3 +329,7 @@
 .tab-content > .active {
   visibility: visible;
 }
+
+.break-text-any-where {
+  word-break: break-all;
+}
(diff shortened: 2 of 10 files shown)


app/views/media/_external_link_card.html.erb
```diff
@@ -5,9 +5,10 @@
     </h5>
   </div>
   <div class="card-body text-center">
-    <%= link_to t('click_here'),
+    <%= link_to external_link_description_not_empty(medium),
                 medium.external_reference_link,
```
fosterfarrell9, line 9: I am not sure if it is visually the most pleasing to have a very large button that visually dominates the whole page if the description text is quite long, but I also do not have an alternative idea. Mabye just plain the text and then a simple "Link"-Button?
button

fosterfarrell9 (approved):

app/views/media/_external_link_card.html.erb
Splines: I thought the display text was just intended for some short text, e.g. "LA Bosch". That's why the input field is also just a small field and not a whole paragraph. The `word-break: break-all;` is therefore just for the edge case, so that the button does not stick out of the outer div. (Note that edge case also means when the display text is empty, then we show the url itself).

While I don't think that this button is overly pretty for long texts, I also don't really like the option of a "Link" button next to the display text as I feel like links should show the thing they point to directly on them and not somewhere else.

PR #494, opened 2023-05-08
Update rubocop version

Splines:
Update rubocop version for use with the [**Ruby LSP** VS Code extension](https://github.com/Shopify/vscode-ruby-lsp) (Language server that works hand in hand with a ruby formatter of your choice, in our case rubocop, and automatically detects rubocop settings).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -201,6 +201,9 @@ Style/RedundantFreeze:
 Style/SymbolProc:
   Enabled: true
 
+Style/SymbolArray:
+  Enabled: false
+
 # Use `foo {}` not `foo{}`.
 Layout/SpaceBeforeBlockBraces:
   Enabled: true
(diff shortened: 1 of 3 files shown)


fosterfarrell9 (changes requested):
I am confused that `Gemfile.lock` is unchanged. It still lists

`    rubocop (0.93.1)`

and other old rubocop dependencies. Did you run 

`bundle update rubocop`

inside the docker container? It should update the `Gemfile.lock`.
See also [this article](https://www.dragonastronauts.com/gemfile-lock-and-git/) and the links there on why it is recommended to include `Gemfile.lock` in the repository (it is by default in a new Rails app; it seems that the consensus is to include it, and hence, have it up to date).

fosterfarrell9 (approved):

PR #502, opened 2023-05-11
Fix flickering for quiz previews

### LINKED ISSUE #501: Hovering over items produces flickering
https://github.com/MaMpf-HD/mampf/assets/37160523/e9454a1b-b3a5-4b14-8326-ef8a96df993c

Splines:
Closes #501

### DIFF
--- a/app/views/media/fill_quizzable_preview.coffee
+++ b/app/views/media/fill_quizzable_preview.coffee
@@ -1,7 +1,5 @@
 # https://stackoverflow.com/a/36227664/
-getScrollTop = ->
-  return Math.max(document.body.scrollTop, document.documentElement.scrollTop)
-oldScrollPos = getScrollTop()
+oldScrollPos = Math.max(document.body.scrollTop, document.documentElement.scrollTop)
 
 $('#mediumPreview').empty()
   .append '<%= j render partial: "quizzes/edit/vertex_status",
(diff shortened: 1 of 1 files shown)


Splines:
https://github.com/MaMpf-HD/mampf/assets/37160523/c6e55998-7398-40e5-b7ab-567e42336823

This first commit [6e96e48](https://github.com/MaMpf-HD/mampf/pull/502/commits/6e96e48545ba2493886cd05df4407af8dbc3f25a) should fix the infinite flickering by restoring the old scroll position once we've appended the new quiz preview to the DOM.

However, there is still room for improvement as can be seen in the video. One inherent limitation is the following: when we have a large quiz preview and the user scrolls down and then hovers over another item where the quiz preview is shorter, we *cannot* keep the header with "Description", "Assigned to" etc. fixed as there is just no space left on the bottom of the page once the bigger element is removed. So, this can be a quick fix, but we might think of alternative ways to represent the quiz preview when hovered.

Splines:
@fosterfarrell9 Can you confirm that there is no infinite flickering anymore? (but there might still be a small "flickering" going on as seen in the video above)

fosterfarrell9:
I can confirm that there is no more infinite flickering.

Splines:
Great, so the issue is resolved. Yet, from a usability standpoint, it is still far from optimal as the content jumping around disrupts the user flow on the page. But this is probably something for another PR and in my opinion should be assigned a low priority as the page itself is working.

Issue #496, opened 2023-05-10
Wrong media types in quiz editor 


fosterfarrell9:
The search form in the quiz editor which is used for selection of questions/remarks in a quiz now has an "all" media button. When this is selected, media of all types (not only questions and reamrks) are searched, but hovering over the results (of wrong type) then results in an exception:

```
 undefined method `locale_with_inheritance' for nil:NilClass I18n.locale = @quizzable.locale_with_inheritance 

/usr/src/app/app/controllers/media_controller.rb:488:in `fill_quizzable_preview'
```
The "all" button should be removed in this form, and exception handling should be improved.

fosterfarrell9:
Also, searching Questions as a non-admin editor or teacher will not give any results; and in the media search in the admin area, you cannot even select "questions as type. These are all regressions introduced in #420.

fosterfarrell9:
Yes, this can be closed via #506.

PR #504, opened 2023-05-13
Hotfix: Restore intended behaviour in media search

fosterfarrell9:
As described in #496, PR #420 has introduced some unwanted behaviour (that also leads to exceptions) when the media search was made accesible to generic users via the start page. In this PR the following things have been changed

- In the 'Import Content' tab in the 'Create vertex' dialogue for quizzes, there is no more radio button that allows you to select all types of media.  You can only choose between Questions and Remarks. (This restores previous behaviour)
- In the 'Media Search' Tab that is available to editors in the admin area, it is now also possible to select media of type Questions/Remark/Erdbeere. (This restores previous behaviour)
- In the media search form on the start page, checking the 'all types' radio button now means checking all  types of media actually listed in this generic search form, which excludes e.g. Questions. For generic users these types have been filtered out anyway if you selected 'all', but for admins they were not, creating an inconsistency. I think that the exclusion of certain types makes sense in that form, because the results are rendered in a way that is not suitable for huge amounts of hits. If you are an editor/admin and want to search Questions, you can use the search form  in the admin area. Maybe it would make sense to add a helpdesk in this accordion fold for editors that explains that this media search is kind of restricted  and that it is recommended to use the media search in the admin area. (This removes an inconsistency)

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -224,7 +224,7 @@ class MediaController < ApplicationController
     results = search.results
     @total = search.total
 
-    # in the case of a search with tag_operator 'or', we 
+    # in the case of a search with tag_operator 'or', we
     # execute two searches and merge the results, where media
     # with the selected tags are now shown at the front of the list
     if search_params[:tag_operator] == "or" and search_params[:all_tags] == "0" and search_params[:fulltext].size >= 2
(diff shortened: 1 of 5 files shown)


fosterfarrell9:
I am not sure if this really falls in the 'hotfix' category despite the exceptions it triggers; I mean, nobody has complained about it. So we could also rebase to `mampf-next` if that fits better.

fosterfarrell9:
I think this is better as a fix to be merged into `mampf-next` instead of a hotfix to `main`, so I close this in favor of #506.

Issue #476, opened 2023-04-26
Improve user management


Splines:
[Low priority ] Improve the admin view for user management.

See [here](https://3.basecamp.com/3978947/buckets/7070405/todos/5185103909).

Splines:
Hey @zebleck, I've seen that you worked on the first version for better user management [here](https://github.com/MaMpf-HD/mampf/commits/better-usermanagement). Would it be possible for you to maybe open a PR for this explaining the current status and where to tie in?

PR #505, opened 2023-05-15
Refactor user controller and views

fosterfarrell9:
This PR just cleans up the user controller and views a little, getting rid of a variable that is no longer needed because data are now retrieved via ajax. Because of the many possible changes via #476, this is not a thorough refactorization, only some obvious stuff is changed.

### DIFF
--- a/app/views/users/_modal.html.erb
+++ b/app/views/users/_modal.html.erb
@@ -23,7 +23,7 @@
       <div class="modal-body"
            id="generic-users-modal-content">
         <%= render partial: "users/generic_users",
-                   locals: { generic_users: generic_users } %>
+                   locals: { generic_users_count: generic_users_count } %>
       </div>
       <div class="modal-footer">
         <div class="row">
--- a/app/views/users/list.json.jbuilder
+++ b/app/views/users/list.json.jbuilder
@@ -1,4 +1,0 @@
-json.array! @users do |user|
-  json.id user.id
-  json.text user.tutorial_info
-end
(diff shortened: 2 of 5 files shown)


app/views/users/_modal.html.erb
```diff
@@ -23,7 +23,7 @@
       <div class="modal-body"
            id="generic-users-modal-content">
         <%= render partial: "users/generic_users",
-                   locals: { generic_users: generic_users } %>
+                   locals: { generic_users_count: generic_users_count } %>
```
Splines, line 26: Why don't we need the "@" here? `@generic_users_count`

Splines (changes requested):

app/views/users/_modal.html.erb
fosterfarrell9: This is because the `users/_modal.html.erb` partial gets the `generic_users_count` variable passed from `users/index.html.erb`. Of course, we could use the instance variable `@generic_users_count` here and would not need to pass it down via `users/index.html.erb`, but it is considered bad practice to use instance variables in partials. Instead, it is recommended to pass variables to partials. See e.g. [here](https://andycroll.com/ruby/dont-use-instance-variables-in-partials/) or [here](https://stackoverflow.com/questions/2503838/rails-should-partials-be-aware-of-instance-variables).

app/views/users/list.json.jbuilder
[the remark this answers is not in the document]
fosterfarrell9: Yes. I thought I might slip this in here, because it fits under the refactorization topic of the PR. Of, course we might put this in a separate PR.

app/views/users/_modal.html.erb
Splines: Ah ok, good to know 👍

Splines (approved):

PR #506, opened 2023-05-15
Fix: Restore media search behaviour

fosterfarrell9:
As described in #496, PR #420 has introduced some unwanted behaviour (that also leads to exceptions) when the media search was made accesible to generic users via the start page. In this PR the following things have been changed

- In the 'Import Content' tab in the 'Create vertex' dialogue for quizzes, there is no more radio button that allows you to select all types of media.  You can only choose between Questions and Remarks. (This restores previous behaviour)
- In the 'Media Search' Tab that is available to editors in the admin area, it is now also possible to select media of type Questions/Remark/Erdbeere. (This restores previous behaviour)
- In the media search form on the start page, checking the 'all types' radio button now means checking all  types of media actually listed in this generic search form, which excludes e.g. Questions. For generic users these types have been filtered out anyway if you selected 'all', but for admins they were not, creating an inconsistency. I think that the exclusion of certain types makes sense in that form, because the results are rendered in a way that is not suitable for huge amounts of hits. If you are an editor/admin and want to search Questions, you can use the search form  in the admin area. Maybe it would make sense to add a helpdesk in this accordion fold for editors that explains that this media search is kind of restricted  and that it is recommended to use the media search in the admin area. (This removes an inconsistency)

### DIFF
--- a/app/helpers/media_helper.rb
+++ b/app/helpers/media_helper.rb
@@ -109,7 +109,7 @@ module MediaHelper
     return add_prompt(Medium.select_quizzables) if purpose == 'quiz'
     return Medium.select_question if purpose == 'clicker'
     return add_prompt(Medium.select_importables) if purpose == 'import'
-    return add_prompt(Medium.select_generic) if !current_user.admin?
+    return add_prompt(Medium.select_generic) if !current_user.admin_or_editor?
 
     add_prompt(Medium.select_sorts)
   end
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -202,7 +202,7 @@ class Medium < ApplicationRecord
   end
 
   def self.advanced_sorts
-    ['RandomQuiz', 'Question', 'Remark', 'Erdbeere']
+    ['Question', 'Remark', 'Erdbeere']
   end
 
   def self.generic_sorts
@@ -282,7 +282,16 @@ class Medium < ApplicationRecord
   # returns search results for the media search with search_params provided
   # by the controller
   def self.search_by(search_params, page)
-    search_params[:types] = [] if search_params[:all_types] == '1'
+    # If the search is initiated from the start page, you can only get
+    # generic media sorts as results even if the 'all' radio button
+    # is seleted
+    if search_params[:all_types] == '1'
+      search_params[:types] = if search_params[:from] == 'start'
+                                Medium.generic_sorts
+                              else
+                                []
+                              end
+    end
     search_params[:teachable_ids] = TeachableParser.new(search_params)
                                                    .teachables_as_strings
     search_params[:editor_ids] =
@@ -295,11 +304,12 @@ class Medium < ApplicationRecord
     if search_params[:all_tags] == '1' && search_params[:tag_operator] == 'and'
       search_params[:tag_ids] = Tag.pluck(:id)
     end
-    admin = User.find_by(id: search_params[:user_id])&.admin?
+    user = User.find_by(id: search_params[:user_id])
     search = Sunspot.new_search(Medium)
     search.build do
       with(:sort, search_params[:types])
-      without(:sort, Medium.advanced_sorts) unless admin
+      without(:sort, 'RandomQuiz')
+      without(:sort, Medium.advanced_sorts) unless user&.admin_or_editor?
       with(:editor_ids, search_params[:editor_ids])
       with(:teachable_compact, search_params[:teachable_ids])
       with(:term_id,
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -287,10 +287,10 @@ class Medium < ApplicationRecord
     # is seleted
     if search_params[:all_types] == '1'
       search_params[:types] = if search_params[:from] == 'start'
-                                Medium.generic_sorts
-                              else
-                                []
-                              end
+        Medium.generic_sorts
+      else
+        []
+      end
     end
     search_params[:teachable_ids] = TeachableParser.new(search_params)
                                                    .teachables_as_strings
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -680,7 +680,7 @@ class MediaController < ApplicationController
                     :teachable_inheritance, :fulltext, :per,
                     :clicker, :purpose, :answers_count,
                     :results_as_list, :all_terms, :all_teachers,
-                    :lecture_option, :user_id,
+                    :lecture_option, :user_id, :from,
                     types: [],
                     teachable_ids: [],
                     tag_ids: [],
--- a/app/views/main/start/_media_search.html.erb
+++ b/app/views/main/start/_media_search.html.erb
@@ -125,7 +125,7 @@
                             '2',
                             checked: false,
                             class: 'custom-control-input',
-                            data: { type: 'toggle', 
+                            data: { type: 'toggle',
                                     id: 'search_media_lectures'} %>
         <%= f.label :lecture_option,
                     t('search.media.lecture_options.own_selection'),
@@ -162,6 +162,7 @@
   <%= f.hidden_field :purpose, value: purpose %>
   <%= f.hidden_field :results_as_list,
                      value: 'false' %>
+  <%= f.hidden_field :from, value: 'start' %>
   <div class="row mb-3">
     <div class="col-12 text-center">
       <%= f.submit t('basics.search'),
--- a/app/views/media/catalog/_search_form.html.erb
+++ b/app/views/media/catalog/_search_form.html.erb
@@ -18,16 +18,20 @@
       <% if purpose == 'clicker' %>
         <%= f.hidden_field :types, value: ['Question'] %>
       <% end %>
-      <div class="custom-control custom-checkbox mb-2">
-        <%= f.check_box :all_types,
-                        id: 'search_all_media_types',
-                        class: 'custom-control-input',
-                        checked: purpose.in?(['media', 'clicker']),
-                        data: { id: 'search_media_types'} %>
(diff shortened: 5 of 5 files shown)


app/models/medium.rb
```diff
@@ -282,7 +282,16 @@ def self.lecture_search_option
   # returns search results for the media search with search_params provided
   # by the controller
   def self.search_by(search_params, page)
-    search_params[:types] = [] if search_params[:all_types] == '1'
+    # If the search is initiated from the start page, you can only get
+    # generic media sorts as results even if the 'all' radio button
```
Splines, line 286: "types" instead of "sorts" (to avoid confusion with "ordering")

EDIT: Oh, I've seen "sort" is used a lot in the code, might be better to leave it like this then to stay consistent.

app/models/medium.rb
```diff
@@ -282,7 +282,16 @@ def self.lecture_search_option
...
+    # is seleted
+    if search_params[:all_types] == '1'
```
Splines, line 288: we can't pass booleans as params in ruby on rails?

app/helpers/media_helper.rb
```diff
@@ -109,7 +109,7 @@ def media_sorts_select(purpose)
     return add_prompt(Medium.select_quizzables) if purpose == 'quiz'
     return Medium.select_question if purpose == 'clicker'
     return add_prompt(Medium.select_importables) if purpose == 'import'
-    return add_prompt(Medium.select_generic) if !current_user.admin?
+    return add_prompt(Medium.select_generic) if !current_user.admin_or_editor?
```
Splines, line 112: rubocop suggest this:

```ruby
unless current_user.admin_or_editor?
      return y(Medium.select_generic)
end
```

It suggested
```ruby
return add_prompt(Medium.select_generic) unless current_user.admin_or_editor?
```
first, but the line was too long then, so it favored the former.

app/models/medium.rb
```diff
@@ -295,11 +304,12 @@ def self.search_by(search_params, page)
...
       search_params[:tag_ids] = Tag.pluck(:id)
     end
-    admin = User.find_by(id: search_params[:user_id])&.admin?
+    user = User.find_by(id: search_params[:user_id])
     search = Sunspot.new_search(Medium)
     search.build do
       with(:sort, search_params[:types])
-      without(:sort, Medium.advanced_sorts) unless admin
+      without(:sort, 'RandomQuiz')
+      without(:sort, Medium.advanced_sorts) unless user&.admin_or_editor?
```
Splines, line 312: That's why the "hidden" field in the HTML is no security issue since we filter "advanced types" out here, right?

app/models/medium.rb
```diff
@@ -202,7 +202,7 @@ def self.select_sorts
   end
 
   def self.advanced_sorts
-    ['RandomQuiz', 'Question', 'Remark', 'Erdbeere']
+    ['Question', 'Remark', 'Erdbeere']
```
Splines, line 205: > In the 'Media Search' Tab that is available to editors in the admin area, it is now also possible to select media of type  Questions/Remark/Erdbeere. (This restores previous behaviour)

Yet, this has been possible beforehand as well, right? It's just that it is now also possible for _editors_, not just _admins_, right?

app/models/medium.rb
```diff
@@ -295,11 +304,12 @@ def self.search_by(search_params, page)
     if search_params[:all_tags] == '1' && search_params[:tag_operator] == 'and'
       search_params[:tag_ids] = Tag.pluck(:id)
     end
-    admin = User.find_by(id: search_params[:user_id])&.admin?
+    user = User.find_by(id: search_params[:user_id])
```
Splines, line 307: So I can pretend I'm another user by sending the corresponding user id in the search params? Where is this mysterious `search_params` variable declared? Is this provided by Ruby (on Rails) by parsing the URL? I cannot find it in the documentation. Is it sanitized?

app/models/medium.rb
```diff
@@ -295,11 +304,12 @@ def self.search_by(search_params, page)
...
     search = Sunspot.new_search(Medium)
     search.build do
       with(:sort, search_params[:types])
-      without(:sort, Medium.advanced_sorts) unless admin
+      without(:sort, 'RandomQuiz')
```
Splines, line 311: So nobody can sort for "random quizzes" anymore?

app/models/medium.rb
fosterfarrell9: A checkbox generated by the Rails form helper e.g. like `<%= f.check_box :all_types, ...%>` will by default return `'1'` for checked and `'0'` for unchecked, see [here](https://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-check_box). If you were to save the result in some `ActiveRecord` model, this would automatically translate to `true`/`false`. But here, `search_params` is just the raw hash of params that survived the whitelisting of some keys and some further manipulation going on in the media controller and is then passed to the medium model. Therefore, this value still `'1'` or `'0'`. I agree that this is realized far from optimal, as you can also see in the media controller: There is way too much business logic about the search happening there. I think that one way out would be to write a plain ruby service model that receives the params hash, deals with business logic that is right now sitting in the media controller, and in particular does things like converting '1', '0' into booleans. Such a model could then be passed to `Medium#search_by` (where again, some logic happening in that method could be made into a method in this service model). But that would possibly be something for a different PR, I think. We have previously used plain ruby models (or PORO for Plain Old Ruby) to some success in refactorization: I think `TeachableParser`, `QuizGraph` and `UserCleaner` are instances of this. There are several texts on that like[ this one](https://www.honeybadger.io/blog/refactor-ruby-rails-service-object/) or [this one](https://scottdomes.medium.com/service-objects-in-rails-75ca74214b77).

app/models/medium.rb
fosterfarrell9: Yes, even if you were to manipulate the form to set `all_types` to '1', you would not get these "advanced types" because of this line(now that I write, I indeed note that type and sort are pretty mixed throughout this code).

app/models/medium.rb
fosterfarrell9: It was possible before #420 for editors and admins to select "advanced sorts". In the current code in production, only admins can do that (by the line above that I removed in `media_helper.rb` and which is responsible for providing the sorts you can select in the dropdown). This change here is to make the distinction in `medium.rb` in lines 311 and 312 possible: 'Random Quizzes' are alway to be filtered out (I will explain this below), while the other sorts are filtered out on some condition.

app/models/medium.rb
fosterfarrell9: The search forms route to the media_controller, you can see this in this line

```
form_with url: media_search_path
```

in the media search partials. In the media controller, there is a search_params method which does filtering ([strong parameters](https://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters) as it is called in Rails). In this method, `search_params[:user_id]` is set via

```
      params[:search][:user_id] = current_user.id
```
so you should be able to alter that as a user. In the above mentioned service model I suggested, one could pass a user (the current user) for initialization.

app/models/medium.rb
fosterfarrell9: Nobody should be able to search Random Quizzes, not even admins. This restores previous behaviour. A 'RandomQuiz' is just a temporary guy: It is a quiz created by a random generator when a user starts a self test with random questions selected by choosing some tags. It is allowed to spend one day in the database until (see `medium.rb`)

```
  scope :expired, -> {
                    where(sort: 'RandomQuiz').where('created_at < ?', 1.day.ago)
                  }
```

and `lib/tasks/cleanup.rake` should make it go away... and hm, I thought there would be a cron job that executes this rake task (at least in production) periodically, but I can't seem to find it. That is one thing we should investigate (separate from this PR). When that rake tast was written we did not yet have `sidekiq-cron` running; with the current codebase a better solution than a rake task in a usual cron job would be a worker that is periodically scheduled via sidekiq. We have such things already happening, see `config/schedule.yml` and `app/workers`.

app/models/medium.rb
```diff
@@ -282,7 +282,16 @@ def self.lecture_search_option
...
   def self.search_by(search_params, page)
-    search_params[:types] = [] if search_params[:all_types] == '1'
+    # If the search is initiated from the start page, you can only get
+    # generic media sorts as results even if the 'all' radio button
+    # is seleted
+    if search_params[:all_types] == '1'
+      search_params[:types] = if search_params[:from] == 'start'
+        Medium.generic_sorts
+      else
+        []
```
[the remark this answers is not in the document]
Splines: ok, then what about

```ruby
search_params[:types] =
  if search_params[:from] == 'start'
    Medium.generic_sorts
  else
    []
  end
```

as seen [here](https://stackoverflow.com/a/13863289). In my opinion, this make it more clear that the variable is assigned a value based on an if-else-block.

app/models/medium.rb
Splines: Thanks for the links to the two blog posts. I agree that this should be something for another PR.

app/models/medium.rb
Splines: Ah ok, that's the reason I couldn't find a `search_params = ` statement in the codebase (except for some tests). But where is `params[:search]` mapped to the variable `search_params`? Is this done internally by rails? I guess `search_params[:user_id]` is performing a call to the `search_params()` method in the media controller, getting back and object and accessing the key `user_id` there. But why don't we have to use `search_params[:search][:user_id]` then?

Splines (approved):

app/models/medium.rb
fosterfarrell9: I'll explain it in our meeting, okay?

PR #509, opened 2023-05-20
Use default version of commontator gem instead of own fork

### LINKED ISSUE #445: Switch back to official release of Commontator gem

fosterfarrell9:
In this PR we switch back to the current version of the commontator gem hosted on github, no longer using our own fork. This fixes #445.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -91,9 +91,7 @@ gem "sidekiq-cron", "~> 1.1"
 gem "faraday", "~> 1.8"
 gem "globalize"
 gem "globalize-accessors"
-gem "commontator",
-  git: "https://github.com/MaMpf-HD/commontator.git",
-  branch: "main"
+gem "commontator"
 gem "acts_as_votable"
 gem "sprockets-rails",
   git: "https://github.com/rails/sprockets-rails",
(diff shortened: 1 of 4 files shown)


fosterfarrell9:
Note that the linter fails because of several style violations in the original commontator comments controller from the gem istelf, which is only slightly modified by us. I let rubocop autocorrect several of these, but I am not sure how to proceed with the rest - do we correct them or do we leave them? There is the possibility that the original file gets an update when the gem updates (although this is rather unlikely given that the last change in this file was 4 years ago).

Splines:
> There is the possibility that the original file gets an update when the gem updates

I'm not sure I understand what you mean here. After merging this PR, don't we have a custom controller which allows us to hook into the relevant parts of Commontator? If so, why should our custom controller be touched when the commontator gem is updated in the future and they change something internally in their library?

In the case we just copied a controller file of theirs: isn't there another way to only change what we really want to change, so that we don't depend on their implementation details of the library?

Splines (approved):
looks fine now and is working properly

Splines:
remember to **squash and merge** instead of creating a merge commit ;)

PR #515, opened 2023-06-12
Disable coloring of comments icon for comments created by user itself

fosterfarrell9:

It has been lamented by some users that the new comment icon is also getting colored when a user itself creates a new comment, and this comment is also listed as unread for this user. This PR changes this behaviour: The new comment icon is no longer colored in this case, and the created comment is not listed as unread.


app/views/commontator/comments/create.js.erb
```diff
@@ -29,10 +29,6 @@
   ) %>");
 <% end %>
 
-<% if @update_icon %>
-$('#commentsIcon').addClass('new-comment');
-<% end %>
```
Splines, line 34: Where was the `@update_icon` variable declared beforehand? I cannot find any occurrences, neither in our codebase, nor in that of `commontator`.

Splines:
With this new behavior, we now get that the new comments icon is not colored anymore when you've just published a comment (which is great I think). Yet, you still see your own comment in the comments list (`/main/comments`). Do we we want that? I have no strong preference but tend towards not seeing your own comment in the comments list (since we also disabled the indicator color on the comments icon, so users might get confused why the comments icon is not colored even though the list shows a comment).

app/views/commontator/comments/create.js.erb
fosterfarrell9: I removed the definition of @update_icon from the comments_controller.

app/controllers/commontator/comments_controller.rb
```diff
@@ -198,8 +198,13 @@ def update_unread_status
       return unless medium.released.in?(['all', 'users', 'subscribers'])
 
       relevant_users = medium.teachable.media_scope.users
-      relevant_users.where(unread_comments: false)
+      relevant_users.where.not(id: current_user.id)
+                    .where(unread_comments: false)
                     .update_all(unread_comments: true)
-      @update_icon = relevant_users.exists?(current_user.id)
```
fosterfarrell9, line 203: This is the original occurence of @update_icon

app/controllers/commontator/comments_controller.rb
```diff
@@ -198,8 +198,13 @@ def update_unread_status
...
+
+      # make sure that the generated comment is marked as read by the creator
+      # so it is not listed as a new comment in the comment index
+      Reader.find_or_create_by(user: current_user, thread: @commontator_thread)
+            .touch
```
Splines, line 208: `touch()`: "Please note that no validation is performed" ~ from [here](https://apidock.com/rails/ActiveRecord/Persistence/touch).

Is there another method that _does_ perform validation? Or why do we skip it? Instead, couldn't we use `save()`, which by default performs validation?

Splines:
> Yet, you still see your own comment in the comments list (/main/comments).

> I cannot reproduce that (was it related to the bug I introduced?) Note that comments are grouped for media. So, if someone else made a new comment on the same medium, it will show in the list. If you can still reproduce it, I would need more info.

With a fresh database, I cannot reproduce this issue anymore, it was probably related to the old logic when you created new reader instances.

Splines:
> We should discuss whether adding a uniqueness constraint in the reader model on the thread/user columns makes sense in order to have an additional saftey net.

I think this would be really helpful. It could be done in a new PR where we also update the usage of Rails' persistance methods, so that methods _with_ validations are used.

fosterfarrell9:
I rewrote the code a little since the last review. In particular I adressed the following issue that occured in the last reviewed version:
If two users A and B who have cleared all comments before wrote a comment at around the same time for the same medium (during that time not seeing whether new comments are popping up), the user who posted their comment last , say B, would get a yellow `commentsIcon` (because the flag was set by A's comment), but when clicking the icon, no comments would appear (because the thread was marked as read by B's post). In the new version, it is checked whether new comments by other users have been posted. If this is the case, the commentsIcon is colored for B, and the thread appears in the new comments list as unread.
Please check if I did not introduce some new issue by this ;-).

Splines:
Thanks, I've played around with it and it works, code looks good too.

There is still one little (workflow) issue in the following scenario:
_Scenario_: Users A and B write comments simultaneously on the same medium. A posts the comment first. Then B posts the comment. Since we haven't implemented something like websockets, person B will only be notified of the comment of person A once B has published his/her comment. Then suddenly, the comments icon is colored for person B. Person B clicks on it and finds that the "Latest post" is by him/herself - which is correct. Still, it seems that the icon was only colored because of B's comment, so person B might just dismiss the comment directly and therefore miss the comment of person A.

In my opinion, this is really an edge case and not easily fixable (and "Latest post" is still correct), so we might just totally ignore it.

fosterfarrell9:
I think we should ignore this edge case here, since formally, everything is correct. (We might an explanation of this for users in the docs in #311, though).

Splines (approved):

Splines:
Remember to **"squash and merge"**.

PR #517, opened 2023-06-14
Fix comment preview button position and its popover

Splines:
Due to commontator taking precedence for some CSS values, we had to overwrite it to reset the padding of the "toggle preview" switch for new comments to the bootstrap defaults.

Also fixed the helpdesk text not showing up next to that switch.


app/views/media/_destination_cards.html.erb
```diff
@@ -125,7 +125,7 @@
                             </div>
                           </div>
                         <div class="col-2 mt-2 text-center">
-                          <div class="form-check custom-switch">
+                          <div class="form-check form-switch">
```
Splines, line 128: Not necessarily related to this PR, however this is such a small change and goes along with replacing "custom-switch" with "form-switch"

fosterfarrell9: Fine with me.

fosterfarrell9 (approved):

Issue #454, opened 2023-04-11
Webpacker has been retired / Upgrade frontend bundling


Splines:
From the Webpacker [Readme](https://github.com/rails/webpacker):

> Webpacker has served the Rails community for over five years as a bridge to compiled and bundled JavaScript. This bridge is no longer needed for most people in most situations following [the release of Rails 7](https://rubyonrails.org/2021/12/15/Rails-7-fulfilling-a-vision). We now have [three great default answers to JavaScript in 2021+](https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b), and thus we will no longer be evolving Webpacker in an official Rails capacity.

> Finally, you can continue to use Webpacker as-is. We will continue to address security issues on the Ruby side of the gem according to [the normal maintenance schedule of Rails](https://guides.rubyonrails.org/maintenance_policy.html#security-issues). But we will not be updating the gem to include newer versions of the JavaScript libraries. This pertains to the v5 edition of this gem that was included by default with previous versions of Rails.

So, while we can except some more security updates in the short term, we should definitely **adopt one of Rails 7 ["new answers to JavaScript"](https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b) in the long run**, to still get security updates (in the distant future) and also new JS features.

> So that's the story for JavaScript in Rails 7 and beyond. A default path with Hotwire and import maps, an alternate path using a thin integration with one of the popular JavaScript bundlers. [...]

## A quote on the old Asset Pipeline (Sprockets) vs. new approaches

> "The problem with Sprockets (the old assets pipeline) is that it did not integrate with front-end package distribution systems like NPM or Yarn and relied on the community to make gemified versions of these packages. If you wanted to use something that was not available as a gem you would have to download it manually or integrate Webpack or some other package manager anyways. It was pretty great back when rails was one of the only frameworks to ship with an assets pipeline but is completely obsolete today." ~ from [here](https://stackoverflow.com/questions/64197591/webpack-vs-asset-pipeline-in-rails-6)

## Further resources
- [Rails 7 will have three great answers to JavaScript in 2021+](https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b)
- [Rails 7 replaces Webpacker with importmapped Hotwire as default JavaScript setup](https://blog.saeloun.com/2021/11/16/replacing-webpacker-with-importmapped)
[...]

PR #526, opened 2023-07-04
Replace datetimepicker

Splines:
The previously used JQuery datetimepicker was imported like this

```js
import css from 'jquery-datetimepicker/build/jquery.datetimepicker.min.css'
import myLib from 'imports-loader?imports=default%20jquery%20$!./../../../node_modules/jquery-datetimepicker/build/jquery.datetimepicker.full.min.js'
```

This dirty fix to direct webpack to the node modules folder has already caused too much pain since this was of course very fragile and thus failed often and seemingly in an undeterministic way. A quick fix was to change the import path, load the page again, then change the path back, so that webpack(er) picks up the changes. This has to find an end.

**Therefore, this PR replaces the previously used [JQuery datetimepicker](https://github.com/xdan/datetimepicker) with the [Tempus Dominus Datetimepicker](https://getdatepicker.com/)**.

image
image

## What to look for in reviews

- [x]  Check that every occurrence of the datetimepicker is working. You can find the datetimepickers here:
    - [x] at the `/lectures/:id/edit` endpoint under the "Homework" section
    - [x] at the `/media/:id/edit` endpoint inside the publish modal. Note that when you create an exercise, there are actually two datetimepickers in this modal! check both: only one datetimepicker (when trying to publish something different than an exercise) and both datetimepickers (when publishing an exercise)
- [x] Errors should be shown if we try to enter a datetime string by hand and the format is not the requested one. The error should go away if a correct date is selected.
- [x] Datetimepicker should show up when clicking in the text input field or when using the tab key to tab into this field. The Datetimepicker should also show up when clicking on the calender button. It should go away when any date is selected. It should not go away when the time is changed.
- [x] Validation of the form when clicking on the submit button should still work, e.g. when no date is selected at all.
- [x] Users should not be able to misuse the endpoint and make a request with an invalid date (the request should then fail). This can only be checked by making a custom request in apps like Postman etc. This should already be the case as the endpoints were not changed.
- [x] No remnants of the old JQuery Datetimepicker should be left in the codebase.
- [x] In the console, there should be no error: neither for the old JQuery datetimepicker, nor for the new Tempus Dominus datetimepicker.
- [x] In the console, there should be no warnings at all. Not even related to missing sourcemaps or debug info.

## Further notes
- We dropped [`Moment.js`](https://momentjs.com/) as library for formatting datetime strings. No frontend datetimeformatting is done anymore. The backend formatting is not affected by this.
- The `javascript/packs/application.sass` file is now an empty file. We cannot delete it right now as that would also mean we'd have to change the webpacker configuration which is done in another PR addressing #454.

### DIFF
--- a/app/assets/javascripts/datetimepicker.js
+++ b/app/assets/javascripts/datetimepicker.js
@@ -1,0 +1,12 @@
+// see https://getdatepicker.com
+console.log('trying to import datetimepicker');
+$('#assignment-picker').tempusDominus({
+    display: {
+        sideBySide: true
+    },
+    localization: {
+        startOfTheWeek: 1,
+        format: 'yyyy-MM-dd HH:mm',
+        hourCycle: 'h23'
+    }
+});
--- a/app/assets/javascripts/datetimepicker.js
+++ b/app/assets/javascripts/datetimepicker.js
@@ -1,12 +1,14 @@
 // see https://getdatepicker.com
 var datetimePicker = new tempusDominus.TempusDominus(
-    document.getElementById('assignment-picker'),
+    // currently, we only support one datetimepicker on a page
+    document.getElementsByClassName('td-picker')[0],
     {
         display: {
             sideBySide: true,
         },
         localization: {
             startOfTheWeek: 1,
+            // choose format to be compliant with backend time format
             format: 'yyyy-MM-dd HH:mm',
             hourCycle: 'h23',
         }
@@ -20,15 +22,27 @@ datetimePicker.dates.parseInput = (input) => {
     try {
         return datetimePicker.dates.oldParseInput(input);
     } catch (err) {
+        console.log($('.td-error'));
+        const errorMsg = $('.td-error').data('td-invalid-date');
+        $('.td-error').text(errorMsg).show();
         datetimePicker.dates.clear();
     }
 };
 
-var isButtonInvokingFocus = false;
+datetimePicker.subscribe(tempusDominus.Namespace.events.change, (e) => {
+    // see https://getdatepicker.com/6/namespace/events.html#change
+    if (e.isValid && !e.isClear) {
+        $('.td-error').empty();
+    }
+});
+
+
 
 // Show datetimepicker when user clicks in text field next to button
 // or when input field receives focus
-$('#assignment-picker-input').on('click focusin', (e) => {
+var isButtonInvokingFocus = false;
+
+$('.td-input').on('click focusin', (e) => {
     try {
         if (!isButtonInvokingFocus) {
             datetimePicker.show();
@@ -39,13 +53,13 @@ $('#assignment-picker-input').on('click focusin', (e) => {
     }
 });
 
-$('#assignment-picker-button').on('click', () => {
+$('.td-picker-button').on('click', () => {
     isButtonInvokingFocus = true;
-    $('#assignment-picker-input').focus();
+    $('.td-input').focus();
 });
 
 // Hide datetimepicker when input field loses focus
-$('#assignment-picker-input').blur((e) => {
+$('.td-input').blur((e) => {
     if (!e.relatedTarget) {
         return;
     }
--- a/app/assets/javascripts/datetimepicker.js
+++ b/app/assets/javascripts/datetimepicker.js
@@ -1,70 +1,91 @@
 $(document).ready(function () {
 
-    // see https://getdatepicker.com
-    var datetimePicker = new tempusDominus.TempusDominus(
-        // currently, we only support one datetimepicker on a page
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/datetimepicker.js
```diff
@@ -0,0 +1,103 @@
+$(document).ready(function () {
+
```
fosterfarrell9, line 2: This should probably be `$(document).on("turbolinks:load",` and so on as we are using turbolinks.

fosterfarrell9 (changes requested):
I played around with it and it seems to work very nicely! I noticed the following things:

- The new datetimepicker does not yet seem to be idempotent (possibly because of the use of `$(document).ready` instead of the turbolinks load event. You can see this by clicking the back and forward button. Playing around in this way, you can produce alive datetimepickers in the middle of the window:
Screenshot 2023-07-12 201826

- Creating absurd inputs sometimes results in a javascript exception as in here:
Screenshot 2023-07-12 201425

- If several error messages appear, they are appended after each other, but not separated in some way:
Screenshot 2023-07-12 2016002

app/assets/javascripts/datetimepicker.js
Splines: No, the turbolink event is already fired when the js is loaded dynamically (which is the case as we use `javascript_include_tag`).

> Any <script src="xxx"> label added to document dynamically will be loaded asynchronously,
and the 'turbolinks:load' event is always triggered before the script finish loading.
That's why 'turbolinks:load' event in the new script not work .
~ from [here](https://github.com/turbolinks/turbolinks/issues/326#issuecomment-339541823)

Instead, I fixed the issue by adding an additional event listener for the turbolink load event that gets fired when the page is visited again via back and forth navigation in the browser. We then look for stale datetimepickers and simply remove them from the DOM.

Splines:
- Dead datetimepickers (turbolink) -> fixed ✅
- absurd inputs, javascript exception -> should be fixed ✅
- several error messages -> ❓ cannot reproduce how did you get both error messages to show at the same time?

app/assets/javascripts/datetimepicker.js
fosterfarrell9: Ah, that is good to know.

fosterfarrell9:
> several error messages -> ❓ cannot reproduce; how did you get both error messages to show at the same time?

I can produce this message by using the datetimepicker for an existing homework: I modify the date like 2022-08-01 07:07 by removing a dash to 202208-01 07:07 and then click on the "Save" button.

Playing around with this I discovered another small issue: If two homeworks are edited at the same time and such an error is created for one them, hitting the save button for this one will lead to errors being displayed for the other one as well (in the example in the screenshot I hit the save button in the third row, but error messages probably destined for the third row were being displayed in the first row)
 
Screenshot 2023-07-13 002725

Splines:
> I can produce this message by using the datetimepicker for an existing homework: I modify the date like 2022-08-01 07:07 by removing a dash to 202208-01 07:07 and then click on the "Save" button.

Oh, good that you do such thorough testing here ;)
Now fixed ✅

> If two homeworks are edited at the same time...

This is actually not related to this PR, you found an already existing bug. Note that other components are affected by this as well, e.g. try to select a wrong deletion date. The reason is in `update.coffee` and related files. Here, we use `$('#assignment-deadline-error')`, but should use `#assignment-deadline-error-<%= @assignment.id %>` instead. However, this means, we also have to use the ids in the HTML code which is currently not the case (it is the case for the `id`s of the respective elements, just not for the divs showing the error). In my opinion, this should be a separate PR.

fosterfarrell9:
I agree, this is a separate PR.
Everything is working very nicely now, so I will aprove. Before that, I have just one more comment (or rather a question). The turbolinks issue was fixed by adding an event listener to the `turbolinks:load` event which removes all stale datetimepickers. From my understanding of the turbolinks workflow, see also [here](https://github.com/turbolinks/turbolinks#preparing-the-page-to-be-cached) or [here](https://github.com/turbolinks/turbolinks/issues/30#issuecomment-195803051), shouldn't it be the `turbolinks:before-cache` event (even if the end result ist the same)? We have also used this in our codebase multiple times in this way.

Splines:
> You can use this event to reset forms, collapse expanded UI elements, or tear down any third-party widgets so the page is ready to be displayed again.

As this is perfectly describing our use case, I've replaced it with `turbolinks:before-cache`, thanks for noting.

Splines:
It'd be great if you could also check the checkboxes in my first comment of this page (hope that you can access these checkboxes and check/uncheck them, even if you're not the person that posted the comment).

fosterfarrell9 (approved):

PR #527, opened 2023-07-04
Add feedback button

### LINKED ISSUE #466: Add button to give feedback
Currently, the only way to give feedback for MaMpf is to look in the footer, see the MaMpf email and write to it. I think we'd receive more feedback if we would make this option more prominent or even have our own "feature" to give feedback.

For example, we could introduce a new button on the right-hand side of the top navbar (or anywhere else) that will open a modal with a text-field. Optionally, users can include their email address in there, so that we can reply. We will then receive a mail with the feedback and the `Reply-to` field of the mail is automaticall set if the user entered their email address, so that we can reply (if necessary).

 This way, users can give feedback in MaMpf without ever leaving the platform and without having to open an email client.

Splines:
New feedback button for users to give feedback regarding MaMpf. Closes #466 

~~⚠ **Merge #537 beforehand, then bring in mampf-next into this branch.**~~

## TODOs
- [x] Prevent sending feedback with empty message
- [x] Make sure everything works when title is empty (title is optional)
- [x] Implement better user flow: close modal and clear message upon successful sending, otherwise show flash error message.
- [x] Set `FEEDBACK_EMAIL` env variable in production! -> won't be reflected in this PR, manual in production environment
- [x] Test email delivery in experimental.

## Preview
image
image
image

## For reviewers

- [x] Check internationalization: everything should work fine in German as well as English including
    - all the texts in the modal
    - success/error toast messages
- [x] Check user flow
    - good path: write feedback and submit, check in [`Mailcatcher`](http://localhost:1080/) if email arrived
    - error paths: write feedback with no characters, in `app/controllers/FeedbacksController` inside the `create` method, add: `@feedback_success = false` right before `respond_to(&:js)` to simulate unsuccessful saving of message to database, simulate bad internet connection, other error paths?
- [x] Code quality & linting!
- [x] Check that "Allow us to contact you via email" is correctly reflected in `Reply-to` of mail (only if checkbox is ticked should we be able to contact the user by replying to the mail, otherwise `Reply-to` field should be empty)
- [x] Clean state: after having successfully submitted feedback, modal should close and next time have texts cleared. If error: modal should stay open and text should not be cleared. Text should also not be cleared when user writes feedback, closes modal (without submitting), then opens it again.
- [x] Mobile: test that feedback submit button is still visible in navbar.
- [x] Own ideas to test this feature: `...` (specify)


Splines:
Hey @Frodo161, this PR is now ready for your review ;) Please see the `For reviewers` section in my upmost comment here. You can tick off the checklist there and also edit my comment.

Frodo161 (commented):
**Problems:**

- "Bitte verwenden Sie zumindest 10 Zeichen" is in German even if the language is set to English.
- If one uses "@feedback_sucess = false" as described in the PR, nothing happens (only the modal won't close on save). Especially, there is no error message (should there be one?). The mail arrives in the mail catcher as expected.
- More generally, I didn't get to see any of the error messages that are written down in the locale files.

**Code:**

- Maybe, instead of writing _authorize_resource except: [:create]_ in the FeedbacksController, you can add an ability file for feedback.

Splines:
@Frodo161 When I change the code in `app/controllers/feedbacks_controller.rb` to the following:

```ruby
def create
    feedback = Feedback.new(feedback_params)
    feedback.user_id = current_user.id
    @feedback_success = feedback.save

    if @feedback_success
      FeedbackMailer.with(feedback: feedback).new_user_feedback_email.deliver_later
    end

    @feedback_success = false

    respond_to(&:js)
end
```

I do see the error message:

image

Wondering why you don't see it, mmh.

Splines:
> "Bitte verwenden Sie zumindest 10 Zeichen" is in German even if the language is set to English.

Good point, I just fixed this and also specified an upper limit of `10 000` chars.

Splines:
> Maybe, instead of writing authorize_resource except: [:create] in the FeedbacksController, you can add an ability file for feedback.

@Frodo161 The thing is: right now there is just the `create` method in the `FeedbacksController`. So I could basically get rid of this statement altogether and don't need any ability as any MaMpf user should be able to give Feeback. However, I wrote this just as precaution, when methods might be added there in the future to prevent them from automatically being unauthorized. Either we leave this in there, or if you want to, I could also get rid of this line.

Note: it'd be awesome if you could (in the future) make comments like these directly next to the respective line (see `Files changed` tab in GitHub), so it's easier to discuss. And the conversation regarding this particular thing can then be closed when finished (which prevents cluttering this thread). Yet another advantage: I don't have to tag you since you wrote the comment and I can directly reply to you.

Frodo161:
I tried again and now I see the error message, but only if I scroll down, because it's hidden by the cards on the main page (see image).

Bildschirmfoto vom 2023-08-26 15-27-01

Splines:
> I tried again and now I see the error message, but only if I scroll down, because it's hidden by the cards on the main page (see image).

@Frodo161 Ah ok, yeah I had that problem too in the beginning and wasted several hours trying to fix this. Then, I noticed that something with Bootstrap did not work and some variables of it were not properly imported (the `z-index` for Toast messages).

Try to:
- Delete the `public/packs` folder locally and restart MaMpf locally.
- Remove the `mampf` and `webpacker` containers including their respective volumes and start again:
```
cd ./docker/compose/
docker compose ps
docker compose rm -s mampf
docker compose rm -s webpacker
docker volume ls
docker volume rm development_public
docker compose up -d
```
(`-s` is shorthand for `--stop`)

- If that doesn't help, also completely rebuild the `mampf` and `webpacker` images:
```
docker compose up --build mampf
docker compose up --build webpacker
```

Hopefully, that will work out. In production, this is not a problem since the `public` folder is replaced in its entirety with new content (or should be, I'm currently on it in #529 ;))

Splines:
Ok, I can reproduce problem with the wrong `z-index`, will have to fix this on my side. Probably had this cached somehow locally.

Splines:
Wow, I found the reason, it was a bug in Bootstrap and fixed in release [`5.2.2`](https://github.com/twbs/bootstrap/releases/tag/v5.2.2) (we had `5.2.1`), see Bootstrap-PR [#37118](https://github.com/twbs/bootstrap/pull/37118).

I've upgraded Bootstrap to version `5.3.1` via `bundle update bootstrap`. Note that the lockfile also indicates some other packages were updated that Bootstrap depends on. A conservative run with `bundle update bootstrap --conservative` did not work (containers did not start due to dependency errors).

Run `bundle install`, so that you fetch the new bootstrap version locally. This should probably be enough:
```
docker compose rm -s nginx
docker compose rm -s mampf
docker compose rm -s webpacker
docker volume remove development_public
docker compose up -d
```

If this doesn't work, you may have to rebuild the whole mampf and webpacker images:
```
docker compose rm -s mampf
docker compose rm -s webpacker
docker compose build mampf webpacker
docker compose up -d
```

Frodo161:
Nice! Now it works.

I still have to take a look at the HTML part of the code, but everything else in this PR seems to be fine for me.

config/initializers/default_setting.rb
Frodo161, line 1: Is the change from ENV[] to ENV.fetch(_, nil) just for a better code styling? If you pass "nil" as a second parameter, it should be equivalent, isn't it?

Splines: Actually, Rubocop did that automatically for me and I didn't even notice (it runs automatically whenever I save a file) 😅
See the [suggestion in the docs](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FetchEnvVar). 

> Suggests `ENV.fetch` for the replacement of `ENV[]`. `ENV[]` silently fails and returns `nil` when the environment variable is unset, which may cause unexpected behaviors when the developer forgets to set it. On the other hand, `ENV.fetch` raises KeyError or returns the explicitly specified default value.

I went over this and removed the `nil` default values, so that `ENV.fetch` will raise a KeyError when the variable is not set which is good as we don't want this to silently go unnoticed. There are actually more such occurrences (`ENV[...]`) in the codebase, so it would be a good idea to address the rest in a separate PR.

Splines:
Since your last review, I added the functionality to submit the form via `Ctrl + Enter`, which should only work when the form is actually opened.

Frodo161:
The `Ctrl + Enter` works in my local MaMpf installation. I also didn't find anything else in the code.

One more thing I had in mind: Should we maybe change the "Save" button into a "Send" button?

config/initializers/default_setting.rb
Splines: I actually found a problem now with this approach, as described in #542, so I reverted this change in commit 1261706a10f94.

Splines:
I've renamed the button as suggested. From my side, everything in this PR is now ready. @fosterfarrell9, it'd be great if you could also take a look at this PR.

Frodo161 (dismissed):

fosterfarrell9 (changes requested):
I noticed the following issues:

- An exception is thrown in the browser console if you access the `/administration` path, e.g. via the navbar (probably because there is no feedback button in the admin navbar)
- Because there is a foreign key to `user_id` in the `feedbacks` table, deletion of users who have created feedbacks will result in a rollback, so it will fail (I could reproduce this, so this is  not purely thoretical). This can probably be fixed by adding the line
`has_many :feedbacks, dependent: :destroy`
to `user.rb` (this will trigger a destruction of all the user's comments before the user's database entry is deleted).

app/views/feedback_mailer/new_user_feedback_email.text.erb
```diff
@@ -0,0 +1,13 @@
...
+# Feedback
+<%= @feedback.feedback %>
+
+
+-----
+<% if @feedback.can_contact %>
+Reply to this mail to contact the user.
+<% else %>
+User did not give permission to contact them regarding their feedback, so we cannot reply to this mail.
+<% end %>
```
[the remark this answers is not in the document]
Splines: I intentionally did not localize this. Do we need to? It's just a mail for us internally. And the language in which we should reply to the mail can be inferred by the language the user used to write the feedback. What advantage would it then have if strings like "Reply to this mail to contact the user" are also in German if the user speaks German?

Splines:
@fosterfarrell9 Thanks for your good observations. I committed fixes for all of them (expect for localization). I fixed the delete on cascade issue by specifying in the `CreateFeedbacks` migration:

```rb
t.references :user, null: false, foreign_key: { on_delete: :cascade }
```

After dropping the database, reseeding it and posting a comment as student5, I was able to successfully do:

```rb
User.where(name: "student5")[0].destroy
```

In the log, you then find this line

```rb
User Destroy (1.6ms)  DELETE FROM "users" WHERE "users"."id" = $1  [["id", 9]]
```
During the deletion, the PostgreSQL database will automatically cascade for us based on our `on_delete: :cascade` statement.

app/views/feedback_mailer/new_user_feedback_email.text.erb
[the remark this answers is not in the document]
fosterfarrell9: I guess I was just surprised when I tested it that I got a mail where parts are in english (whereas my development setup and my production setup is set on German and I do not see English in many places except maybe exception messages). But I can live with the current solution.

db/schema.rb
```diff
@@ -906,6 +916,7 @@
   add_foreign_key "commontator_subscriptions", "commontator_threads", column: "thread_id", on_update: :cascade, on_delete: :cascade
   add_foreign_key "course_self_joins", "courses"
   add_foreign_key "divisions", "programs"
+  add_foreign_key "feedbacks", "users", on_delete: :cascade
```
fosterfarrell9, line 919: Technically, that is certainly a correct solution. I am just a little worried that we are now using different methods that achieve the same in our `user` model and mix them. I did not count it, but in `user.rb` you will find lots of relations that have the `dependent: :destroy` flag set. So these are all at the same place. By delegating the cascading for `feedbacks` to the database, we now open up a second place where behavior for such dependencies is stored. Maybe there should at least be a comment in user.rb that there is also a relation to feedbacks, but that deletion is handled on a database levels (alternatively, add `has_many :feedbacks, dependent: :destroy` to `user.rb`, then the cascading in the migration is not necessary). Note that there a lot of articles in the net about `dependent: :destroy` vs '`on_delete: :cascade)`'.

Splines: Oh good point. I agree we should respect business logic when deleting an active record and have the additional Rails layer, i.e. favor `has_many :feedbacks over dependent: :destroy` with the latter working directly on the database. I've just changed it in 07ba509b828c.

fosterfarrell9 (approved):

fosterfarrell9 (dismissed):
Looks good to me. Before merging into production: Has this

- [x]  Set FEEDBACK_EMAIL env variable in production! -> won't be reflected in this PR, manual in production environment

been done yet?

Splines:
Will merge without approval as Denis is on vacation.

PR #528, opened 2023-07-12
Fix styling of small dot indicating current navbar item

Splines:
In the Apple Safari browser, the dot under the currently active navbar-item was not placed at the correct position. This is how it should be:

image

⚠ Before merging: let Apple user actually test this on mampf-experimental.


fosterfarrell9 (approved):

Splines:
Fix was verified on an iPhone and MacBook by fellow students 👍

PR #529, opened 2023-07-14
[Delete later] Fix/webpacker assets not being served

Splines:
Ever since the `application.js` file was altered in #523, we had problems in our test environments with assets not being served correctly (but not in localhost). Our docker setup did not include the `public` volume for mampf workers but only for the master container. So, within the last years, we basically accidentally always created two public folders in different containers and only one of them was served.

As the `application.js` file was altered, it was only available in the worker container where the `assets:precompile` task is executed. However, the static assets are only served from the specified `public` volume. The fix mounts the volume for the `public` folder into the worker docker container where the `assets:precompile` task is executed, so that the file can be served from any container that has also mounted the volume.

I originally tried this which did not work:
[this medium post](https://medium.com/@rubalps/rails-cloudfront-webpack-823558c71b58) and [this comment](https://github.com/rails/webpacker/issues/1651#issuecomment-513482825).

### DIFF
--- a/app/assets/javascripts/datetimepicker.js
+++ b/app/assets/javascripts/datetimepicker.js
@@ -22,6 +22,24 @@ function startInitialization() {
     });
 }
 
+function getDateTimePickerIcons() {
+    // At the moment: continue to use FontAwesome 5 icons
+    // see https://getdatepicker.com/6/plugins/fa5.html
+    // see https://github.com/Eonasdan/tempus-dominus/blob/master/dist/plugins/fa-five.js
+    return {
+        type: 'icons',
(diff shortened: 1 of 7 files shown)


Splines:
@tynsh I've requested a review from you: just if you've got time and want to take a look as you have originally introduced the docker setup.

Splines:
We solved this problem in the server script that triggers the docker files, so this is not needed anymore and was actually not a good solution in the first place since the public volume does not have to be available to the workers and so it shouldn't. Instead `nginx` is in play here and does the magic serving the public files.

PR #512, opened 2023-05-25
Init new annotation tool

Splines:
See feature overview in [this comment](https://github.com/MaMpf-HD/mampf/pull/512#issuecomment-1627722606) down below.

PR #533, opened 2023-08-23
Add annotation tool & migrate to JS in Thyme Player

Frodo161:
This branch adds the feature to add annotations to the video timeline. They can be used for giving feedback to the teacher, for getting help, for publishing something directly in the video comments or - simply - to make personal notes.

By default students can create annotations on the video timeline. Click on the button next to the speed control to open a dialogue.

image

Enter the text for your annotation (you can use Latex!), choose a color and select the category. You can publish the content directly as a comment and - if the teacher enabled this feature - make your annotation visible for the teacher. If the teacher entered an "emergency-link" (see below) and if you choose "content" as category and then "definition" the dialogue will give you a link for further help.

Markers will appear on the video timeline once you click on "Save". If you click on one of them their content will be shown in the interactive area on the right.

To see the annotations of students (which are marked as "visible for teacher") the teacher has to activate the emergency button on the lecture edit page (comments).

If it is (not) activated there is still the posibility to (de)activate it for each single video. By default the option "inherit from lecture" is selected.

Go again to the lecture edit page (preferences this time). There you can enter links (or select a lecture) which is shown to the students as "emergency-link", if they have problems with understanding a definition.

To see the feedback of the student, teachers now have the "feedback player" which can be accessed on the media edit page (where the video was uploaded). In this very simple player, the teacher sees all the annotations (that are marked as "visible for teacher") together with a heatmap. The teacher can (de)select the categories of annotations that are shown to him at the bottom of the video.

Mistake annotations are slightly bigger than other annotations and are not included in the calculation of the heatmap!


Splines:
See my first review on the closed branch here: #512

app/views/annotations/_form.html.erb
```diff
@@ -0,0 +1,151 @@
...
+    <br>
+
+    <!--- VISIBLE FOR TEACHER? - CHECKBOX -->
+    <% annotations_status = Medium.find_by(id: @medium_id).get_annotations_status %>
+    <% if annotations_status == 1 %>
+      <%= f.label :visible_for_teacher, class: "checkbox inline" do %>
+        <span, title="<%= t('admin.annotation.visible_for_teacher_tooltip') %>"><b>
+          <%= t('admin.annotation.visible_for_teacher') %> &nbsp
+        </b></span>
+        <%= f.check_box :visible_for_teacher %>
```
Splines, line 114: Checkbox not at the same height as label:

image

Splines (changes requested):
Pretty cool to play around with the new feature. Some first remarks referring to the UI.

Note that I've copied this first review from [here](https://github.com/MaMpf-HD/mampf/pull/512#pullrequestreview-1591375234). Now, we have everything on this new PR and you can safely ignore the old one #512.

app/assets/stylesheets/thyme.scss
```diff
@@ -341,3 +418,234 @@ figure {
...
+  margin-left: 10%;
+  margin-right: 10%;
+  margin-bottom: 15px;
+  margin-top: auto;
+}
+
+
+/* A collection of all colors.
+Be aware that color changes must also be written in the
+AnnotationHelper! */
```
Splines, line 507: No, please avoid these things where we have duplicated code and then even a comment telling: when you touch this, you also have to touch this. See my proposal in the `_form.html.erb` file. Short form: keep it here, remove it from the helper.

app/views/annotations/_form.html.erb
```diff
@@ -0,0 +1,151 @@
...
+
+    <br>
+
+    <!--- COLOR SELECTION -->
+    <b><%= t('admin.annotation.color') %></b>
+    <div id="annotation-color-picker">
+      <% for i in 1..15 do %>
+        <%= f.radio_button :color,
+                           annotation_color(i),
+                           id: "annotation_color#{i}" %>
```
Splines, line 44: Why do we have the `annotation_color(int)` in the `AnnotationsHelper` in the first place? The only usage I see is here. Why not add a class here, so that we only need the map of colors defined in `thyme.scss`? (where I feel it is perfectly placed as this is frontend not backend related)

app/helpers/annotations_helper.rb
Splines, line 1: maybe get rid of this file altogether, see my comment in the `_form.thml.erb` file for the color selection.

app/assets/stylesheets/thyme.scss
```diff
@@ -341,3 +418,234 @@ figure {
...
+#annotation-color-picker {
+  input[type="radio"] {
+    display: none;
+    &:checked + label {
+      span { transform: scale(1.25); }
+      @each $name, $value in $annotation_colors {
+        .#{$name} {
+          border: 2px solid black;
+        }
+      }
```
Splines, line 539: Why have multiple CSS classes that do the exact same thing? Instead: define a class like `annotation-color-activated` and assign it accordingly to the currently selected element.

app/controllers/media_controller.rb
```diff
@@ -515,6 +515,17 @@ def fill_reassign_modal
     @no_rights = params[:rights] == 'none'
   end
 
+  def check_annotation_visibility
+    medium = Medium.find_by_id(params[:id])
+    isPermitted = medium.annotations_visible?(current_user)
```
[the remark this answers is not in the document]
Splines: Then you can probably do something like

```ruby
render json: { isPermitted: is_permitted }
```

in the next line.

app/assets/stylesheets/thyme.scss
```diff
@@ -341,3 +418,234 @@ figure {
...
+    }
+    span {
+      display: block;
+      width: 100%;
+      height: 100%;
+      transition: transform .2s ease-in-out;
+      @each $name, $value in $annotation_colors {
+        &.#{$name} {
+          background: $value;
+        }
```
Splines, line 562: same thing as above

app/javascript/packs/application.js
Splines, line 1: remove from this PR

app/assets/stylesheets/thyme.scss
```diff
@@ -341,3 +418,234 @@ figure {
...
+  annotation_color12: #924129,
+  annotation_color13: #444444,
+  annotation_color14: #999999,
+  annotation_color15: #eeeeee,
+);
+
+* {
+  box-sizing: border-box;
+}
+
```
Splines, line 529: Isn't this the default? Even if it isn't, please don't use `*`, as this will refer to every HTML element on every possible MaMpf page due to the bundling.

app/assets/stylesheets/thyme.scss
```diff
@@ -341,3 +418,234 @@ figure {
...
+        &.#{$name} {
+          background: $value;
+        }
+      }
+    }
+  }
+}
+
+
+/* THYME FEEDBACK */
```
Splines, line 569: Ofentimes, when putting a comment like this acting as a visual separator, one knows that this is probably a good breaking point to put the rest in a separate file. I feel this is a very good example: thyme-related CSS that concerns only the feedback player should probably go in a separate file like `tyhme_feedback_player.scss`.

app/views/media/feedback.html.erb
Splines, line 1: maybe rename to `feedback_player` as mentioned somewhere else

config/locales/de.yml
```diff
@@ -322,6 +322,54 @@ de:
...
+      next_annotation: 'nächste Annotation'
+      further_help: 'Du brauchst weitere Hilfe? Dann klicke auf:'
+      warning_message:
+        publishing: >
+          Jeder, der diese Vorlesung abonniert hat, wird deinen Kommentar
+          lesen können, sobald du auf "Speichern" klickst.
+          Falls du das nicht möchtest, entferne bitte wieder den Haken
+          bei "Als Kommentar veröffentlichen?".
+        mistake: >
+          Bevor du diesen Kommentar abschickst, überprüfe außerdem bitte,
```
Splines, line 355: We have consistently capitalized `du` so far. Note this is not the only line where you've used `du` instead of `Du`. Watch out for simple search and replace, there's lots of variants like `Deinen`, `Dich` etc.

db/migrate/20230117124508_add_annotations_status_to_medium.rb
Splines, line 1: We should only need database migrations in subsequent code changes, not in the initial PR that introduces the feature. You could manually add the changes from these migrations to the `schema.rb`, so that there are no new migrations at all in this PR.

app/controllers/annotations_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+    
+    render json: annots
+  end
+
+  def near_mistake_annotations
+    annots = Annotation.where(medium: params[:mediumId])
+    radius = params[:radius].to_i
+    timestamp = params[:timestamp].to_i
+    validAnnots = []
+    for a in annots
```
Splines, line 90: Luckily, we don't have to pay for characters 😅 I'd strongly prefer `for annotation in annotations` here. One-char-variable names only result in confusion in my experience. Code is more read than written, so it should be clear immediately what we mean.

app/controllers/annotations_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+      if a.category == "mistake" &&
+         (a.timestamp.total_seconds - timestamp).abs() < radius &&
+         a.public_comment_id != nil
+        validAnnots.push(a)
+      end
+    end
+    c = validAnnots.size()
```
Splines, line 97: Why do you store all annotations in an array, but in the end only return the size of that array. Instead: use a counter variable.

Splines: Furthermore: the method should be renamed to make clear that we are only intersted in the number not the actual annotations here, e.g. `num_nearby_mistake_annotations` (note also `nearby` instead of `near`).

Splines (changes requested):
I'm by far not yet finished with the code review, here some more remarks as intermediate result. Please make sure you have `rubocop` locally installed, so that it can show you some code suggestions. Otherwise, it's tedious that I have to open my editor, see what rubocop says, and copy it in here as comment. Note it's only responsible for Ruby files, not `.erb` files, for those you can install an ERB linter (which is not working very well for me in VSCode, will have to experiment a bit more with that).

Locally, I've slightly changed the `.rubocop.yml` file, see it [here](https://gist.github.com/Splines/2fa569dcfec1422eaa74e8e9d1e64b1e). It just adds some more excludes, so that one can focus on relevant things.

app/views/annotations/_form.html.erb
```diff
@@ -0,0 +1,151 @@
...
+
+    <%= f.label :visible_for_teacher, class: "checkbox inline" do %>
+      <span><b><%= t('admin.annotation.post_as_comment') %> &nbsp</b></span>
+      <%= f.check_box :post_as_comment %>
+    <% end %>
+
+    <br>
+
+    <div class="modal-footer">
+      <%= f.submit t('buttons.save'), class: "btn btn-primary", id: "submit-button" %>
```
Splines, line 129: choose more specific id since `submit-button` is probably not unique (many components have a submit button)

app/views/annotations/_form.html.erb
```diff
@@ -0,0 +1,151 @@
...
+      </div>
+       <!--- for prefilling the edit form -->
+      <div id="annotation_subtext">
+        <%= @annotation.subtext %>
+      </div>
+    </div>
+
+    <!--- COMMENT TEXT AREA -->
+    <b><%= t('admin.annotation.comment') %></b>
+    <%= f.text_area :comment, class: 'form-control' %>
```
Splines, line 22: Ensure no empty comments are stored. See [here](https://github.com/MaMpf-HD/mampf/pull/527/files#diff-40587fc8ef28ff3ce7ef7f060ae9ecd81d7dfa7697b881968c4c54d55844caa9R24-R29). Make sure to also validate in backend, not only frontend.

app/assets/javascripts/thyme.js
```diff
@@ -0,0 +1,1092 @@
...
+  }
+
+  // Update annotations after submitting the annotations form
+  $(document).on('click', '#submit-button', function() {
+    /* NOTE:
+       Updating might take some time on the backend,
+       so I added a slight delay.
+       I couldn't think of an easy way to let the script
+       wait for the update to complete (as with the delete button),
+       but it might be possible! */
```
Splines, line 684: You can send a JS file to the frontend when the controller is finished. This file could contain `updateMarkers()`. See how I realized something similar for the feedback button:
- I have a `create` method [here](https://github.com/MaMpf-HD/mampf/pull/527/files#diff-2b0bba658ec28b34a9c6bea0cacecc10022b929dd8aed5df15d92e1d93856e86R4-R7).
- Then, correspondingly to the name of the controller method, a `create.js.erb` file [here](https://github.com/MaMpf-HD/mampf/pull/527/files#diff-cee5fcef04cf78d4b0883370c679f768eb2d1ae4b8a4561674ecefd9dfb414a3) that is sent to the frontend.

app/assets/javascripts/thyme.js
Splines, line 1: This file was already very long (more than 700 lines of code). Try to put the new annotation-related code into another `.js` file. If `thyme.js` and the annotation tool share functions, they can also be outsourced to yet another separate file.

app/assets/javascripts/thyme.js
```diff
@@ -0,0 +1,1092 @@
...
+  if (times.length == 0) {
+    return;
+  }
+  var i = 0;
+  while (i < times.length) {
+    if (times[i] > seconds) {
+      return times[i];
+    }
+    ++i;
+  }
```
Splines, line 34: Use case for a `for` loop.

Splines: No sorry, this was already the case beforehand in the coffeescript. We only consider new changes here, so ignore this. See the review comment for a proposal.

Splines (changes requested):
Right now, some parts a hard to review in this PR due to the conversion of coffeescript to JS (which I indeed asked you to do). However, this PR would be more manageable if just the *annotation-tool-related coffeescript lines* would be converted to JS. Note that I say "lines", not "files". For example `thyme.js` is pretty big right now and GitHub shows me that a lot of new lines have been added. It is difficult to distinguish between lines that are really new regarding the annotation tool or lines that are just "new" as they were converted by the online decaffeinater.

### My proposal is as follows:
- Keep the converted JS files somewhere on your PC locally for a later PR (or just save the commit hash right now, so we can jump back to these file later on).
- ⚠Outsource your annotation-tool specific javascript code to new `.js` files, e.g. `thyme_annotations.js`. Restore the old `.coffeescript` files like `thyme.coffeescript`.
- In a future PR named "Decaffeinate project (Coffeescript to JavaScript)", we can then bundle the annoying "get rid of CoffeeScript" changes together and fix things that the decaffeinater does not perfect manually, but in many files at once. This "de-clutters" this PR which should only consider things related to the annotation tool and not some old stuff from other files.

I know this is a bit extra-work, but I think it's worth it.

Note I also fell prey to decaffeination, e.g. [here](https://github.com/MaMpf-HD/mampf/pull/534/files#diff-fb6efe10767d0182a352ba6fefefbd182ae57091321bbffcc0a5b659eb4d317d), however there it's really just one file with 24 lines of code, not huge files.

app/controllers/media_controller.rb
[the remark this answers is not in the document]
Frodo161: `render json: { isPermitted: is_permitted }` didn't work and I didn't find anything else. For the moment, I'd leave it as it is.

app/controllers/media_controller.rb
```diff
@@ -515,6 +515,17 @@ def fill_reassign_modal
...
+    render json: isPermitted
+  end
+  
+  def feedback
```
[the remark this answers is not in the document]
Frodo161: I like `feedback` a little bit more because it's more consistent with the other players (it's not `play_player` or `enrich_player`). Because this is only the name of the method in the media controller (and the feedback button doesn't interfere here), I think there shouldn't be any confusion.

app/models/medium.rb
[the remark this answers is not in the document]
Frodo161: You're definitely right! But let's postpone this to a future branch ;).

app/views/annotations/_form.html.erb
Frodo161: Wouldn't you say that in this case it's a feature to have the possibility of a blank comment field?
Maybe, sometimes users only want a marker on their video timeline without text. But I'm open for other opinions!

app/views/annotations/_form.html.erb
Frodo161: Ok, that's strange (in my browser they are aligned). Also, in your case, the color picker doesn't fit in one line (which it should). I'll take a closer look on it!

config/locales/de.yml
Frodo161: Good catch! Thanks to your comment, I also found a type in multiple_close_annotations_2 ;).

db/migrate/20230117124508_add_annotations_status_to_medium.rb
Splines: Edit: Sorry, at the time I was obviously not too aware of how these migrations work in detail. So actually we need of course at least one new migration file. I probably meant to rather have just one migration instead of multiple ones. E.g. when you init a table and then later have a migration that adds another column, we could just have that as one migration where all necessary columns are added in the first place.

app/assets/stylesheets/thyme.scss
Frodo161: Actually, removing these lines doesn't change anything as far as I see it (I thought, we needed it for a better behaviour of the color boxes, but eventually this is not the case).

app/models/medium.rb
```diff
@@ -1125,7 +1125,22 @@ def subscribed_users
...
   end
-
+  
+  def get_annotations_status
+    if annotations_status != 0
+      return annotations_status
+    else
+      return lecture.annotations_status
+    end
+  end
```
fosterfarrell9, line 1135: It would probably be more rubyesque to write this as

```
def get_annotations_status
  return annotations_status if annotations_status != 0
  lecture.annotation_status
end
```

app/controllers/annotations_controller.rb
```diff
@@ -141,4 +143,19 @@ def comment_html(medium, total_seconds, comment)
...
 
+    def update_comments
+      for annotation in Annotation.where(medium: params[:mediumId])
+        # find comment
+        unless annotation.public_comment_id == nil
+          comment = Commontator::Comment.find_by(id: annotation.public_comment_id).body
+
+          # remove "Thymestamp: H:MM:SS" at the end of the string
+          index = comment.rindex("\nThymestamp")
+          comment = comment[0 .. index - 1]
```
Splines, line 154: What if the "Thymestamp" part was removed by the user (by editing the comment)? What if the user inserted text after the "Thymestamp"?

Splines: Instead, maybe search for the whole string using a regular expression that allows for a variable "H:MM:SS" part. However, we should definitely include a case when no match was found. In this case, we should probably not strip anything from the comment.

Generally, this feels like a pain point here since with the right data model, this string editing would not even be necessary, e.g. if a comment in the database had an annotation linked to it and would show the respective thymestamp in a separate HTML paragraph-element or something. However, this would include having to customize the commontator logic and UI representation which would result in another pain point :( Probably, this way is better then for right now. But again, this shows how intricate this comment-annotation relationship really is. Maybe we can improve this in the future all along a new idea: #540.

Splines: By postponing this: do you mean completely dropping the "post as comments" feature for right now or just the part that editing the comments does not sync up with the annotation? If only the latter, it's an inconsistency we have to be aware of and users will probably notice it as a "bug". But I agree that the state of this, as of right now, is not really satisfying and this kind of string manipulation would only lead to more errors in the long run.

app/assets/javascripts/thyme/annotations/annotation.js
```diff
@@ -80,48 +80,9 @@ class Annotation {
...
-    switch (this.category) {
-      case "note":
-        return "#44ee11"; //green
-      case "content":
-        return "#eeee00"; //yellow
-      case "mistake":
-        return "#ff0000"; //red
-      case "presentation":
-        return "#ff9933"; //orange
-    }
```
Splines: Yeah 🎉, now this is what I call the benefits of encapsulation. Really cool approach with the "Enum" ;)

Splines (changes requested):
A few remarks regarding the new enum approach

app/assets/javascripts/thyme/thyme_feedback.js
```diff
@@ -34,34 +34,35 @@ $(document).on('turbolinks:load', function() {
...
+    category = allCategories[i];
+    annotationCategoryToggles[i] = new AnnotationCategoryToggle(
+      'toggle-' + category.name + '-annotations', category, heatmap
+    );
+    if (category === Category.MISTAKE) {
+      // exclude mistake annotations from heatmap
+      annotationCategoryToggles[i].heatmap = null;
+    }
+    annotationCategoryToggles[i].add();
+  }
```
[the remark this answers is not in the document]
Frodo161: I wanted to have the same indices in `allCategories` and `annotationCategoryToggles`. I guess, it will have the correct ordering with a for-of loop, but I'm a little bit careful.

app/controllers/annotations_controller.rb
Frodo161: I meant the latter one, i.e. you can post an annotation as a comment, but once posted the comment has nothing to do with the annotation anymore and both can be edited independently.

On the other side: If someone is not actively trying to mess up, the string manipulation won't do any harm (once, I've added regex and filter the string-not-found case).

app/assets/javascripts/thyme/thyme_feedback.js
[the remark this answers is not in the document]
Splines: sorry, I didn't see you had two arrays here, so a for loop with an index is actually needed. Another alternative would be to iterate over `allCategories.entries()` as seen [here](https://stackoverflow.com/a/51683785/9655481), but a for loop with in index does the exact same thing, so it's fine.

app/controllers/annotations_controller.rb
Splines: It could do harm depending on how you handle certain cases, e.g. when a user says

```
I really dig this new "Thymestamp 0:04:35" feature.
```

which could then become

```
I really dig this new "" feature.
```

which is semantically altering what the user was saying 😂 But of course this is an extreme edge-case (we should still have in mind though). Or what about having multiple "Thymestamp" strings in one comment, what do you do in this case?

Splines: @fosterfarrell9 What do you think? Do you prefer

- having the "post as comment" feature right now, but without any link between comment and annotation. Maybe add a tooltip text next to the "post as comment" checkbox and make clear that editing the comment or the annotation will not affect the other side)
- having the "post as comment" feature right now with some degree of linkage between comments and annotations. In one direction (edit annotation -> update comment) full link, in other direction (edit comment -> update annotation) not at all or "hacky" with this string manipulation.
- not having the "post as comment" feature for right now and find a better solution in the future

I tend to prefer the first option where annotations can be posted as comments but editing annotations and comments is completely isolated (until a better solution, maybe with hooks into the commontator gem, is found)

Frodo161: > which is semantically altering what the user was saying 😂 But of course this is an extreme edge-case (we should still have in mind though). Or what about having multiple "Thymestamp" strings in one comment, what do you do in this case?
I only catched the last occurance of "Thymestamp". So, normally: If you edit the comment, you leave the thymestamp as it is. If you edit the annotation, you won't see the thymestamp.

I'm not sure yet, which option I'd prefer. From the code side, it's pretty ugly that users have the possibility to cause unwanted behaviour. On the other side, this is not going to happen in practise, so maybe, it's better to accept this not-so-nice solution until we have something better.

docker/development/.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.4
```
Splines, line 1: why update in this PR?

db/schema.rb
```diff
@@ -272,6 +298,9 @@
     t.integer "submission_max_team_size"
     t.integer "submission_grace_period", default: 15
     t.boolean "legacy_seminar", default: false
+    t.integer "annotations_status"
+    t.integer "emergency_link_status", default: 0
+    t.text "emergency_link"
```
Splines, line 303: "emergency" even made it to the database ;) remove emergency branding (see comment above).

Splines (changes requested):
Yet another review ;) Please make sure you start your methods with a verb and not a noun or adjective for better readability and since this is convention. See the respective comments.

app/assets/javascripts/application.js
```diff
@@ -49,10 +49,52 @@
...
 //= require upload
 //= require users
 //= require vertices
 //= require watchlists
 //= require turbolinks
+
+/*
+ * THYME RELATED SCRIPTS
+ * (warning: the order of the scripts is important
+ * - do not switch to alphabetical order!)
```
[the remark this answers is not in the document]
Frodo161: Yes, e.g. we have
`//= require thyme/components/component
//= require thyme/components/add_item_button`
because AddItemButton extends Component.

app/assets/javascripts/thyme/components/annotations_toggle.js
```diff
@@ -0,0 +1,53 @@
...
+  /*
+    Auxiliary method
+  */
+  show() {
+    $('#volume-controls').css('left', '66%');
+    $('#speed-control').css('left', '77%');
+    $('#emergency-button').css('left', '86%');
+    thymeAttributes.hideControlBarThreshold.x = 960;
+    this.div.show();
+    //updateControlBarType(); TODO
```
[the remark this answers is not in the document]
Frodo161: Good catch!

I think, it's not necessary to update the control bar type here (any more), because the updateControlBarType() method is called at the very end of the thyme player script (when everything is loaded, and it should stay like this way).

app/assets/javascripts/thyme/components/ia_button.js
```diff
@@ -0,0 +1,75 @@
+class IaButton extends Component {
```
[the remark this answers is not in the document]
Frodo161: IA always means "Interactive Area". Would a comment also be ok? (IA is a fixed expression in the thyme related JS and I got used to it.)

app/assets/javascripts/application.js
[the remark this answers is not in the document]
Splines: wow, I read "do not switch *the* alphabetical order" 🙈

app/assets/javascripts/thyme/thyme_feedback.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+    return false;
+  }
+  const annotationArea = new AnnotationArea(false, colorFunc, isValid);
+  thymeAttributes.annotationArea = annotationArea;
+  function strokeColorFunc(annotation) {
+    return annotation.category === Category.MISTAKE ? 'darkred' : 'black';
+  }
+  function sizeFunc(annotation) {
+    return annotation.category === Category.MISTAKE ? true : false;
```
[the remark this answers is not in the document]
Frodo161: In this case, I thought it was a good idea to break with the convention, because the function definition here is like a variable definition, i.e. we're only defining a specific behaviour here to give it to the annotation manager as parameter. For me, it visually helps to see, what we're doing. But if you want me to insert newlines anyway, I'll do it.

app/assets/javascripts/thyme/thyme_player.js
```diff
@@ -0,0 +1,212 @@
+$(document).on('turbolinks:load', function() {
+  /*
+    VIDEO INITIALIZATION
+   */
+  // exit script if the current page has no thyme player
+  const thymeContainer = document.getElementById('thyme-container');
+  if (thymeContainer === null || $('#video').get(0) === null) {
+    return;
+  }
```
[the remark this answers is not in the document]
Frodo161: I'm still not sure if we really need the `$('#video').get(0) === null` part (@fosterfarrell9, do you still know, why you included this statement?).
The `thymeContainer === null` is there to exit the script if we're not on the thyme page, so here we don't want to have a warning.

app/assets/javascripts/thyme/thyme_player.js
```diff
@@ -0,0 +1,212 @@
...
+    const annotationId = Number(document.getElementById('annotation_id').textContent);
+    $.ajax(Routes.annotation_path(annotationId), {
+      type: 'DELETE',
+      dataType: 'json',
+      data: {
+        annotationId: annotationId
+      },
+      success: function() {
+        annotationManager.updateAnnotations();
+        // close and open again = show normal IA
```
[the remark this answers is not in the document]
Frodo161: Interactive Area - the area on the right side of the player :D.

app/assets/javascripts/thyme/utility.js
```diff
@@ -0,0 +1,146 @@
...
+    Converts a given integer between 0 and 255 into a hexadecimal, s.t. e.g. "15" becomes "0f"
+    (instead of just "f") -> needed for correct format.
+   */
+  toHexaDecimal: function(int) {
+    if (int > 15) {
+      return int.toString(16);
+    } else {
+      return "0" + int.toString(16);
+    }
+  },
```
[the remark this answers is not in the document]
Frodo161: That's nice!

As all functions in the utility object are written down as `doSth: function() { }`, I'd prefer to write it down like this for consitency:
`toHexaDecimal: function(int) {`
 ` 	return int.toString(16).padStart(2, '0');`
 ` },`
But I'm open minded for your solution :).

app/controllers/annotations_controller.rb
```diff
@@ -0,0 +1,161 @@
...
+
+  def num_nearby_mistake_annotations
+    annotations = Annotation.where(medium: params[:mediumId])
+    radius = params[:radius].to_i
+    timestamp = params[:timestamp].to_i
+    counter = 0
+    for annotation in annotations
+      if annotation.category == "mistake" &&
+         (annotation.timestamp.total_seconds - timestamp).abs() < radius &&
+         annotation.public_comment_id != nil
```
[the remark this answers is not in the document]
Frodo161: The "post as comment" checkbox is checked by default for mistake annotations, but it's possible to unclick these.

The `num_nearby_mistake_annotations` method must only count posted annotations (if e.g. 10 students mark a mistake for themselves, but no one posts it, it's ok for the 11th student to post it's annotation and therefore he shouldn't be warned in this case).

We _could_ rename `num_nearby_mistake_annotations` into `num_nearby_posted_mistake_annotations`. But please, let's find a shorter name :D.

app/controllers/lectures_controller.rb
```diff
@@ -23,6 +23,15 @@ def edit
                               Time.parse(ENV['RAILS_CACHE_ID'])].max)
       eager_load_stuff
     end
+
+    # emergency link -> prefill form
+    status = @lecture.emergency_link_status
+    link = @lecture.emergency_link
+    if status == "lecture_link"
```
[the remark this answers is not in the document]
Frodo161: You mean, use
`status = @lecture.emergency_link_status_for_database`
and
`if status == Lecture.emergency_link_statuses[:lecture_link]`?
Or is there a better way?

app/controllers/lectures_controller.rb
```diff
@@ -230,6 +248,12 @@ def search_examples
 
   def close_comments
     @lecture.close_comments!(current_user)
+    # disable emergency button
```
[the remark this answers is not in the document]
Frodo161: I guess, calling the feature "emergency button" is deprecated by now. Still, I'd like to call the button, which opens the annotation modal, "emergency button" (even though I see, it's most of the time not used for emergencies). Do you think, this is a good idea, or do you also want me to rename this button?

app/views/annotations/edit.js.erb
```diff
@@ -0,0 +1,240 @@
...
+    var rad = 60; // annotations that are inside this radius (in seconds) are considered as "near". 
+    $.ajax(Routes.num_nearby_mistake_annotations_path(), {
+      type: 'GET',
+      dataType: 'json',
+      data: {
+        mediumId: medId,
+        timestamp: video.currentTime,
+        radius: rad,
+      },
+      success: function(c) {
```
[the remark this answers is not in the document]
Frodo161: The number of nearby mistake annotations. I replaced it by `count`.

app/views/annotations/edit.js.erb
```diff
@@ -0,0 +1,240 @@
...
+        content();
+        break;
+      case "mistake":
+        mistake();
+        break;
+      case "presentation":
+        presentation();
+        break;
+    }
+  }
```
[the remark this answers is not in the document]
Frodo161: I didn't find a way to share code between the asset JS and the view JS. If you have something, let me know!

docker/development/.ruby-version
Frodo161: I can't remember having changed anything in the docker folder. This file should also be removed from this PR.

db/schema.rb
Frodo161: But in this case the branding "emergency" actually makes sense. What would be your suggestion for renaming it?

app/views/annotations/edit.js.erb
```diff
@@ -0,0 +1,240 @@
...
+  } else {
+    previewCSS(false, 100, 100, 0);
+  }
+});
+
+// disable post comment checkbox if annotation was already posted
+if (<%= @posted %>) {
+  postComment(true);
+  postAsComment.disabled = true;
+}
```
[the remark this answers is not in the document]
Frodo161: Yes, this is also checked in the update method of the annotations controller, so no annotation can be posted twice (even if someone manipulates the form in the frontend such that the checkbox is activated again).

app/views/annotations/edit.js.erb
```diff
@@ -0,0 +1,240 @@
...
+        left: '\\(',
+        right: '\\)',
+        display: false
+      }, {
+        left: '\\[',
+        right: '\\]',
+        display: true
+      }
+    ],
+    throwOnError: false
```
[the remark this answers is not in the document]
Frodo161: I copy-pasted this part of the code from the "old" thyme player, so that would again be a question for @fosterfarrell9.

app/assets/javascripts/thyme/thyme_player.js
```diff
@@ -0,0 +1,212 @@
...
+
+  /*
+    INTERACTIVE AREA
+   */
+  iaButton.toHide = [$('#caption'), annotationArea];
+  iaButton.add();
+  (new IaCloseButton('ia-close', iaButton)).add();
+  //TODO
+  //const interactiveArea = new InteractiveArea();
+  //thymeAttributes.interactiveArea = interactiveArea;
```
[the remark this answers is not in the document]
Frodo161: Good catch! I wanted to create an InteractiveArea class, because I also have created an AnnotationArea class and in theory this would lead to a better code structure. However, in practise the interactive area is far more complicated than the annotation area and as far as I tried, the separated class only made things more complicated.
In future, I think it would improve the code structure if we think again about this carefully, but I'd like to postpone this to a future release.

app/assets/javascripts/thyme/thyme_player.js
```diff
@@ -0,0 +1,212 @@
...
+    /* NOTE:
+       Updating might take some time on the backend,
+       so I added a slight delay.
+       I couldn't think of an easy way to let the script
+       wait for the update to complete (as with the delete button),
+       but it might be possible! */
+    setTimeout(function() {
+      annotationManager.updateAnnotations();
+    }, 500);
+  });
```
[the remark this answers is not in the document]
Frodo161: I don't fully understand, how you would do it. I need something that triggers the updateAnnotations() method, once the backend has updated the corresponding annotation. How can I send a success response from the update method of the AnnotationsController to thyme_player.js?

app/assets/javascripts/thyme/display_manager.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // mediaQuery listener for very small screens
+    const match_verysmall_x = window.matchMedia("screen and (max-width: " +
+      thymeAttributes.hideControlBarThreshold.x + "px)");
+    match_verysmall_x.addListener(function(result) {
+      if (result.matches) {
+        dm.smallDisplay();
+      }
+    });
```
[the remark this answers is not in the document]
Frodo161: Actually, resizing is still very buggy right now. I try to fix it in general and then come back to your comment.

fosterfarrell9:
> @fosterfarrell9 What do you think? Do you prefer
>
>having the "post as comment" feature right now, but without any link between comment and annotation. Maybe add a tooltip >text next to the "post as comment" checkbox and make clear that editing the comment or the annotation will not affect the >other side)

I think this option ist okay with me. By the way, there is a blogpost [here](https://medium.com/indent-labs/extending-thredded-or-any-rails-gem-with-activerecord-callbacks-aa0b5aa4e630) that could serve as a blueprint on how to connect comments and annotations (in the nontrivial direction). See also in their [repository](https://github.com/indentlabs/notebook). The only drawback ist that they are still on Rails 6.2. In Rails 7, the whole autoloading process was [changed](https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html), so their solution might not work for us.

fosterfarrell9:
I realized a linking between the `Commontator::Comment` and `Annotation` models using the above mentioned blueprint in [this](https://github.com/MaMpf-HD/mampf/tree/annotation-tool-experiments) branch, it seems to work nicely. I added some syntactic sugar on both models (in particular creating a `public_comment` method on `Annotation` and an `annotation` method on `Comment`) but did not fill out all details. Of course, this does not yet resolve the "thymestamp" problem. For that I would suggest using the `config.javascript_proc` in `config/initializers/commontator.rb`. On pages that display coments, we could put the timestamps in some data attributes and display them by some javascript code in this proc (we already use this proc for TeX). Then the comment body would always be the same as the annotation body, and the "thymestamp" would go away when editing the comment.

docker/development/.ruby-version
Splines: I removed it.

pdfcomprezzor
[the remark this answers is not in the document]
Splines: Fixed with the help of [this](https://stackoverflow.com/a/23668025/9655481) and [this](https://stackoverflow.com/a/64967374/9655481) and [this](https://stackoverflow.com/a/50877349/9655481).

app/assets/javascripts/thyme/thyme_feedback.js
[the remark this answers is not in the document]
Splines: Once we have the linter installed, it's going to enforce this rule anyways ;) So let's already do it. I don't see the advantage of not having the newline here.

app/assets/javascripts/thyme/components/ia_button.js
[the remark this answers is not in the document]
Splines: Ok, you've included the comment, so I guess it's fine 👍

app/assets/javascripts/thyme/display_manager.js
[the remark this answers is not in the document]
Splines: What is the current state of affairs here?

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Splines: > so here we don't want to have a warning.

But isn't `thymeContainer === null` a case we don't ever expect to happen? So basically the early return here is just a failsafe for our code. A warning or even better an error in the browser would then be appropriate as it indicates that something really went wrong and somehow the `thymeContainer` is `null` although it shouldn't be this way.

Did I perceive this correctly?

app/assets/javascripts/thyme/thyme_player.js
```diff
@@ -0,0 +1,212 @@
...
+    alert($('body').data('badbrowser'));
+    $('#caption').hide();
+    $('#annotation-caption').hide();
+    $('#video-controlBar').hide();
+    video.style.width = '100%';
+    video.controls = true;
+    resizeContainer();
+    window.onresize = resizeContainer;
+    return;
+  }
```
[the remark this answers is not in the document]
Splines: Browser detection generally is considered bad practice, see the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent).

> Serving different Web pages or services to different browsers is usually a bad idea. The Web is meant to be accessible to everyone, regardless of which browser or device they're using. There are ways to develop your website to progressively enhance itself based on the availability of features rather than by targeting specific browsers.

Also note that the `navigator.userAgent` strings constantly changes. E.g. as of October 2023, the string for Edge is actually not "Edge", but "Edg" (see [here](https://stackoverflow.com/a/77234353/9655481) and I can confirm this on my machine).  Also see [this answer](https://stackoverflow.com/a/21712356/9655481) for how often the UserAgent string for Internet Explorer (IE) changed with time.

If one really wanted to detect the browser, one should use `navigator.userAgentData` and not the "legacy" `userAgent`, see [here](https://learn.microsoft.com/en-us/microsoft-edge/web-platform/user-agent-guidance).

---

All in all, I feel we should not bother with non-compliant browsers. If somebody is still using Internet Explorer these days to access MaMpf, we can't help it. For now, what about getting rid at least of the `/Edge/.test(navigator.userAgent)` part and file a future PR to get rid of IE support completely? If you agree, I could create a new issue so that we don't forget it.

app/assets/javascripts/thyme/utility.js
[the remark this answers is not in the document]
Splines: Sure, the way you adapted it is fine. But I stumbled on the `int` as variable name as in many languages this is actually a reserved keyword (not in JS, but still, it might lead to confusion). So why not call it `integer`, `intValue` or something along these lines. Of course, this one is very nitpicky :)

db/schema.rb
```diff
@@ -10,11 +10,27 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_04_27_124337) do
+ActiveRecord::Schema[7.0].define(version: 2023_10_01_065435) do
```
Splines, line 13: image

I've just purged all my Docker data, initialized Mampf with the seed data and went to the lecture edit page. For `Lineare Algebra 2, WS 2022/23`, no option was selected. Likewise for any other lecture I have tested. Running `rake db:migrate` did not help either.

More concerning though: when I clicked on "Yes", this change was not reflected and users could still not make their annotations visible for the teacher.

Splines: In the database, `annotation_status` has the value `nil`. We probably want a default value there.

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Frodo161: The point is, that all code of the JS files is executed on every MaMpf page. So we need a condition for the JS to exit the script if we're not on the Thyme Player page. This part of the script looks at the HTML of the page and searches for an element with the right ID 'thyme-container'. If this ID doesn't exist on the given page, it should exit the JS (which is a wanted behaviour and therefore shouldn't return an error message).

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Frodo161: But this is a JS file from the view. My action has to contact a JS file from the asset folder.

A hidden HTML element would be a possible (although maybe not the best) solution for that problem, because it can serve as a bridge between view JS and asset JS.

db/schema.rb
Frodo161: You're right, I should definitely add a default value to the schema.
I'm not entirely sure, but from the description of your problem, I would assume, that you didn't set a value on the medium level: Currently, if no value is set on the medium level, it also deactivates the feature by default, even if a value is set on the lecture level.

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Frodo161: Ok, in that case let me just remove the `$('#video').get(0) === null` part.

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Frodo161: I updated it now in the way, I tried to describe above.

db/migrate/20231021100348_change_annotations_status_default_in_media_and_lectures.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeAnnotationsStatusDefaultInMediaAndLectures < ActiveRecord::Migration[7.0]
+  def change
+    change_column_default :media, :annotations_status, 0
+    change_column_default :lectures, :annotations_status, -1
```
Splines: What does it mean if the annotation status is `0` or `-1`? The only case I see the value `-1` is in the `close_comments` method in the lectures controller. How is `0` different from `-1`?

db/migrate/20221015114100_create_annotations.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_table :annotations do |t|
+      t.references :medium, null: false, foreign_key: true
+      t.references :user, null: false, foreign_key: true
+      t.text :timestamp, null: false
+      t.text :comment
+      t.integer :category
+      t.boolean :visible_for_teacher
+      t.string :color
+
```
Splines, line 13: You did not specify `null: false` for some of these vales, so is the rest of the code working fine if we have a `null` value somewhere in here?

Splines (changes requested):
Some more comments. Note I merged together some annotation-related migrations. For example: if you add a key (in this case: "subtext") to the annotation table, then later remove it, why not get rid of it entirely from the codebase? Similar for cases where you later added a default value: this can be added directly to the "Create table" migration.

Of course, if this branch was already merged, you must not alter any migrations. But in the current state, we can do anything we want to the newly added migrations which makes dealing with them in the future easier.

Please make sure I've merged together the migrations correctly (this was done manually). Here are the changes: 526bc087b59c35f0d

db/migrate/20231021100348_change_annotations_status_default_in_media_and_lectures.rb
Splines: If we only needed a binary state for "on/off", then a boolean would suffice, right? That's why I assume that `-1` has a special meaning which I did not find out yet.

db/schema.rb
Splines: Yeah, you're right, in this case it does make sense if it's just referring to the one link leading to further resources 👍

app/assets/javascripts/thyme/annotations/annotation.js
```diff
@@ -0,0 +1,104 @@
+/**
+  This class helps to represent an annotation in JavaScript.
+*/
+class Annotation {
+
+  constructor(json) {
+    // We only save attributes that are needed in the thyme related JavaScripts in the asset pipeline!
```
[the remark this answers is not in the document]
Splines: you resolved this conversation but didn't change anything or provided a comment

app/assets/javascripts/thyme/display_manager.js
```diff
@@ -0,0 +1,130 @@
...
+
+    if (window.matchMedia("screen and (max-device-width: " +
+        thymeAttributes.hideControlBarThreshold.x + "px)").matches ||
+        window.matchMedia("screen and (max-device-height: " +
+        thymeAttributes.hideControlBarThreshold.y + "px)").matches) {
+      dm.smallDisplay();
+    }
+
+    // mediaQuery listener for very small screens
+    const match_verysmall_x = window.matchMedia("screen and (max-width: " +
```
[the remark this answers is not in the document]
Splines: not resolved yet

app/assets/javascripts/thyme/heatmap.js
```diff
@@ -0,0 +1,118 @@
...
+    /* for any visible annotation, this array contains its color (needed for the calculation
+       of the heatmap color) */
+    let colors = [];
+
+    /*
+       data calculation
+    */
+    for (const a of thymeAttributes.annotations) {
+      const valid = this.#isValidCategory(a.category);
+      if (valid === true) {
```
Splines, line 51: Replace with `if (valid)`, same a few lines below

app/assets/javascripts/thyme/key_shortcuts.js
```diff
@@ -0,0 +1,112 @@
...
+      Arrow Down - previous chapter
+      i - toggle interactive area
+  */
+  addPlayerShortcuts() {
+    const iaButton = document.getElementById('ia-active');
+    const nextChapterButton = document.getElementById('next-chapter');
+    const previousChapterButton = document.getElementById('previous-chapter');
+
+    window.addEventListener('keydown', function(evt) {
+      if (thymeAttributes.lockKeyListeners === true) {
```
Splines, line 62: get rid of `=== true` (multiple occurrences throughout this PR) if you can be sure that the underlying variable can only hold `boolean` values.

Aside: To test for non-existence use `if (!myProperty)` but make sure you don't need the distinction between `undefined` and `null`. E.g. search for `=== null` in this PR to apply this change.

app/assets/javascripts/thyme/metadata_manager.js
```diff
@@ -0,0 +1,221 @@
...
+      // if the metadata cue changes, highlight all current media and scroll
+      // them into view
+      $(metaTrack).on('cuechange', function() {
+        let j = 0;
+        const time = video.currentTime;
+        $('#' + metadataListId + ' li').removeClass('current');
+        while (j < this.activeCues.length) {
+          const activeStart = this.activeCues[j].startTime;
+          let metalink;
+          if (metalink = document.getElementById('m-' + activeStart)) {
```
Splines, line 207: Prefer to assign data to the variable prior to the `if` statement. Then only do: `if (metalink)`

app/controllers/annotations_controller.rb
```diff
@@ -0,0 +1,160 @@
...
+    end
+    @total_seconds = @annotation.timestamp.total_seconds
+    @medium_id = @annotation.medium_id
+    @posted = (@annotation.public_comment_id != nil)
+  end
+
+  def update
+    @annotation = Annotation.find(params[:id])
+    comment_id = @annotation.public_comment_id
+    if (comment_id == nil)
```
Splines, line 31: Please make sure to use **rubocop** for stylistic hints (I get many more suggestions for this file for example). For this specific line, rubocop suggests to use `if comment_id.nil?`. Also: no parentheses around `if` statements in Ruby.

image

app/controllers/annotations_controller.rb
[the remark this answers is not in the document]
Splines: Of course, `num_nearby_posted_mistake_annotations` is long, but in this case I find it perfectly valid as this makes it crystal clear what the method is really doing. E.g. you read something like "ajax get num nearby posted mistake annotations", this is almost like spoken language. So, it'd be great if you could indeed rename it ;)

---

Note these lines also make for a very good candidate for rubocop. It shortened it to this statement:

```rb
annotations.each do |annotation|
      next unless annotation.category == 'mistake' &&
                  (annotation.timestamp.total_seconds - timestamp).abs() < radius &&
                  !annotation.public_comment_id.nil?

      counter += 1
end
```

Splines (changes requested):
Some more stylistic things.

app/controllers/lectures_controller.rb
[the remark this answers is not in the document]
Splines: I'm not sure what you mean by `emergency_link_status_for_database`. I meant the statement
```rb
if status == "lecture_link"
```

should be replaced by what you provided in your comment:
```rb
if status == Lecture.emergency_link_statuses[:lecture_link]
```

This way, we will encounter an error if the string for `lecture_link` is changed later on and this code occurrency is overlooked during the refactor. Otherwise (with just the string), it might not be noticed and the check just fails since the `status` is not that specific string anymore.

app/controllers/lectures_controller.rb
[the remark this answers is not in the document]
Splines: I'd prefer to have it called "Annotation button" as the button is referring to the "Annotation tool" and not an "Emergency tool". For the link where one can find additional information, I'm fine with "emergency link" as described in another thread.

app/views/annotations/edit.js.erb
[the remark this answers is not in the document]
Splines: `Category.NOTE.name` works for me

fosterfarrell9:
> Note these lines also make for a very good candidate for rubocop. It shortened it to this statement:
> 
> annotations.each do |annotation|
>       next unless annotation.category == 'mistake' &&
>                   (annotation.timestamp.total_seconds - timestamp).abs() < radius &&
>                   !annotation.public_comment_id.nil?
> 
>       counter += 1
> end

It would probably even more rubyesque if we would use Ruby's own `count` method with a block:

```
counter = annotations.to_a.count do |annotation|
   annotation.category == 'mistake' &&
        (annotation.timestamp.total_seconds - timestamp).abs() < radius &&
        !annotation.public_comment_id.nil?
end
```

app/assets/javascripts/thyme/components/annotation_category_toggle.js
```diff
@@ -16,6 +16,9 @@ class AnnotationCategoryToggle extends Component {
     const category = this.category;
     const check = document.getElementById(this.element.id + '-check');
     const heatmap = this.heatmap;
+    if (heatmap != null) {
```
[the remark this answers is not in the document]
Splines: you marked this as resolved but did not provide a comment or changed the file

Splines (commented):
Comments referring to 80924242d and afadcff6

app/assets/javascripts/thyme/annotations/annotation.js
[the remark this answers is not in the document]
Frodo161: Strange. I could swear, I deleted "in the asset pipeline" before I resolved this conversation.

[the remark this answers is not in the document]
Frodo161: I could wear, that I changed this. Thanks for noticing!

app/assets/javascripts/thyme/components/annotation_category_toggle.js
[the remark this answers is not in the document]
Frodo161: It seems that I changed the `if (heatmap != null)` in line 25 and thought it was the line you referenced. Good catch, I'm going to change the other one as well!

app/assets/javascripts/thyme/key_shortcuts.js
Frodo161: Ok, I changed this everywhere. I have to admit, that I tend to use the `==` operator to check if an object is null rather than `===` for having the `undefined` case included (maybe even in some situation where it can be ensured that this doesn't happen). Still, for now I only replaced the appearances of `=== null`.

app/controllers/lectures_controller.rb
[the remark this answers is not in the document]
Frodo161: The thing is, `Lecture.emergency_link_statuses[:lecture_link]` returns the integer associated to `lecture_link` while `@lecture.emergency_link_status` returns the string representation. So it doesn't make sense to compare them. Using `emergency_link_status_for_database` also returns an integer such that we can now compare both values.

app/views/annotations/_form.html.erb
Frodo161: Now, in my browser the text next to the checkboxes is top aligned :P:
Bildschirmfoto vom 2023-10-27 19-34-23

app/views/annotations/edit.js.erb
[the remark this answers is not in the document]
Frodo161: Nice! I replaced it in all switch-case-statements. If you think, there are some other passages in this script where one can improve the code structure with the enum, let me know (I took a look at the whole script again but e.g. I think it's ok having functions named after the categories as we are in a view script and want very different behaviour for each category).

db/migrate/20221015114100_create_annotations.rb
Frodo161: I added a constraint for `category`. Also `color` shouldn't be nil - but here I added a more restrictive regex constraint that also filters strings which don't represent a hexadecimal color. Therefore a non-null constraint might be overkill (still, I can add one if you like).
All the other field actually have to be `nil` sometimes, so I'll won't change anything here.

Splines: - It'd be great, if you could add these kinds of migrations directly do the original migration as described in [this comment](https://github.com/MaMpf-HD/mampf/pull/533#pullrequestreview-1691457370). Note that I've changed this now with the new commit: 63c15eaf25629
- What about the `subcategory` integer. Is it valid for an annotation to not have a subcategory?
- Can you point me to the exact code where you filter the strings to check if it is a hexadecimal color? Is this done in the backend? I'd still advocate to have `null: false` here as well. Otherwise, what does it mean if an annotation in the database has `nil` for its color?
- Why doesn't `visible_for_teacher` have to be non-null?

app/assets/javascripts/thyme/display_manager.js
```diff
@@ -38,48 +38,50 @@ class DisplayManager {
 
   // Check screen size and trigger the right method
   updateControlBarType() {
+    const dm = this;
+
```
Splines, line 42: I've reverted this in 759d99c368dd9, why did you change it back? I don't think we need to rename `this` since `this` is very common in many object-oriented programming languages (e.g. Java/C++, in Rust/Python the keyword is called `self`). It would only confuse to refer to `this` under another name.

Frodo161: The resize()-function didn't work properly anymore. The problem was that in the scope of the listeners down below `this` refered to the element wo which the listener was attached to and not to the display manager. Using a different reference for the current display manager solves this problem.

db/migrate/20221015114100_create_annotations.rb
Frodo161: - `subcategory` must only be selected if you select `content` as category (actually I should include that condition into the backend). So for each other category, it's ok for that field to be nil.
- The String check is done in the `is_valid_color`-method in the `AnnotationsController` (used by create and update), l. 148. Actually, in this method I already filter out the nil-case, so we'll never be in the situation that we have a nil value in the db. But I'll add the additional restriction - doesn't do any harm and maybe, we'll change something later and then it's better to have it ;).
- I decided to allow `visible_for_teacher` to be `nil` (and use this in basically the same way as `false`) because this part of the feature can be deactivated (originally, I wanted to differ between "user is not allowed to choose an option" and "user is actively saying no"). One could add a non-nil restriction but then has to pass `false` as default value (either in the form or in the controller).

app/assets/javascripts/thyme/display_manager.js
Splines: Note that I updated the `updateControlBarType()` function in 8cd43416be24 to reduce its complexity.

app/assets/javascripts/thyme/display_manager.js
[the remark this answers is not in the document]
Splines: I've reduced the complexity of this in 8cd43416be24.

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Splines: Your change did not work for me in a chromium-based browser since the event you were using is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/MutationEvent). I've rewritten the logic as proposed without any additional HTML elements, see commit 8fcc2389.

app/views/annotations/_form.html.erb
Splines: That's weird. Are checkboxes and labels aligned [in this example](https://getbootstrap.com/docs/5.3/forms/checks-radios/#checks) in your browser?

Splines:
@Frodo161 I've merged the ESLint changes into this branch. Until the ESLint branch is not merged into `mampf-next`, you will see many unrelated commits here but they will disappear as soon as the ESLint branch is merged tomorrow or in a few days.

## ESLint
I resolved most warnings and errors in ESLint, but some are still open and I leave them to you as you wrote the respective code.  See for example the [summary page of the last run](https://github.com/MaMpf-HD/mampf/actions/runs/7400999983?pr=533) for all warnings and errors. All of them have to be resolved. The remaining ESLint errors cannot be autofixed. (Many RuboCop errors can though, see below).

See [this wiki entry](https://github.com/MaMpf-HD/mampf/wiki/Code) on how to setup RuboCop and ESLint. You can then run `yarn run eslint .` in your terminal. If you're using VSCode, you can even Ctrl + click on the warnings/errors to jump directly to the location in the code.

Note you have to pull this branch to obtain the `.vscode/settings.json` file with many prepopulated and hopefully helpful settings for the editor and its plugins. But of course, you can do everything solely from the command line as well if you want to (but might take a lot longer). One setting is for example that files are auto-linted whenever you save them, so no code style violations are accidentally introduced. And even if, our pipeline will check for this again.

## RuboCop

Please also resolve the **RuboCop** warnings and errors. You can use the command I've used [here](https://github.com/MaMpf-HD/mampf/pull/566):
```
bundle exec rubocop --autocorrect-all --disable-uncorrectable # used later
```

It should only change things related to this branch as we've already merged the PR that makes our *entire codebase* compliant to our RuboCop rules. **Please make sure you commit the *safe* autocorrections separately from the *unsafe* autocorrections.** Again, if you run this inside a VSCode terminal, you can directly jump to the respective line of code by ctrl + clicking. Note that with the `--disble-uncorrectable` option, rubocop will add `rubocop:todo` comments that you will have to resolve. You can find them easily by viewing the locally modified files in the Git tab. Or search for "rubocop:todo" in the entire codebase. However, the last option will already yield 16 results not related to the annotation tool. 

---

If you've got any questions, feel free to ask. Especially if you encounter scenarios were you think RuboCop might not give the best result. Maybe [another RuboCop rule](https://docs.rubocop.org/rubocop-rails/cops_rails.html) is needed in this case.

app/controllers/lectures_controller.rb
```diff
@@ -343,12 +343,10 @@ def lecture_params
...
                         :emergency_link_status]
       if action_name == "update" && current_user.can_update_personell?(@lecture)
-        allowed_params.concat([:teacher_id, { editor_ids: [] }])
-      end
-      if action_name == "create"
-        allowed_params.concat([:course_id, :teacher_id, { editor_ids: [] }])
+        allowed_params.push(:teacher_id, { editor_ids: [] })
       end
       allowed_params.push(:course_id, :teacher_id, { editor_ids: [] }) if action_name == "create"
+      allowed_params.push(:course_id, :teacher_id, { editor_ids: [] }) if action_name == "create"
```
Splines, line 349: You duplicated the line here

Splines (changes requested):

db/migrate/20240123180500_add_annotation_related_fields.rb
```diff
@@ -0,0 +1,13 @@
+class AddAnnotationRelatedFields < ActiveRecord::Migration[7.0]
+  def change
+    # Annotations status
+    add_column :media, :annotations_status, :integer, default: 0, null: false
+    add_column :lectures, :annotations_status, :integer, default: -1, null: false
+
+    # Emergency Link
+    change_table :users, bulk: true do |t|
```
Splines, line 8: I've just cleaned up the migrations by merging them in fewer files. In the previously existing file with the class `AddEmergencyLinkStatusAndEmergencyLinkToLecture`, I found this line. Why do you change the `users` table here and not the `lectures` table?

Splines (changes requested):

db/migrate/20240123180500_add_annotation_related_fields.rb
Splines: I just fixed this in 783e73e7 as we got server errors otherwise since `emergency_link` was missing in the lectures table. Please check that commit if I did it correctly.

Splines:
How the UI is looking as of 66b6891

image

app/controllers/lectures_controller.rb
Frodo161: Don't know, how this happened (this line is not even related to the Annotation feature). I deleted it.

db/migrate/20240123180500_add_annotation_related_fields.rb
Frodo161: Yes, it must be lectures, not users. The table should now be correct.

Splines:
Demo of the feedback player as of commit 7f15b0c

mampf-feedback-player

app/controllers/lectures_controller.rb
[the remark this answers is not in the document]
Splines: done in 57ba6e69b6

app/views/lectures/edit.html.erb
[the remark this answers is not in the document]
Splines: fixed in 4cf369248cbdd42

app/assets/stylesheets/thyme.scss
```diff
@@ -341,3 +418,234 @@ figure {
...
+    }
+    .switch {
+      vertical-align: middle;
+    }
+  }
+  #annotation-modal-preview {
+    height: 543px;
+    width: 35%;
+    word-wrap: break-word;
+    overflow-y: scroll;
```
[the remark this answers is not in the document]
Splines: fixed in later commits, we also have a `annotations.scss` now

app/views/annotations/_form.html.erb
Splines: We did everything as stated in the docs [here](https://getbootstrap.com/docs/5.3/forms/checks-radios/#checks), so I don't know why your browser is rendering it differently. I hope this is already fixed in newer comments. I cannot reproduce it in Chromium-based browsers (Brave & Edge). Will resolve it as to move on, if this is still an issue, please open a GitHub issue ;)

app/assets/javascripts/thyme/thyme_player.js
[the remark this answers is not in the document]
Splines: ok will resolve this conversation. If we find bugs later on, we might recall that we removed this statement 😅

Splines (approved):
Finally, after a long ride, I approve this PR.

To be honest, +4,677 code additions and -1,116 code deletions have shown me why huge pull requests like this one are frowned upon. It's just not manageable. Even with a new feature, there's always a way to break things down into smaller pieces. And there's no need for one PR to equal one release, e.g. one strategy could have been to create smaller PRs and merge them into the `annotation-tool` branch one step at a time, then later have a big `annotation-tool` to `dev` PR, but you have already gone through all the code in smaller PRs.

Hindsight is easier than foresight, of course. But I hope we learn our lessons from this PR. It's a sign if even the GitHub UI takes years to load the Git diff in the `Files changed` tab.

--- 

Despite all this: the new annotation feature is great and I'm really thrilled to get it shipped soon. With such a beast of a PR, I wasn't able to review every single line of code. The bigger picture is fine with me, but I wouldn't stake my life on the annotation tool not including any bugs. For sure, there will be some, but I'm sure we can tackle them quickly. The new Ruby and JS linters also assure code quality and looking at the code, the new modularized thyme player is really a big step forward away from coffeescript, thanks a lot @Frodo161 for your efforts!

I will merge this PR soon into `dev`, the last stage before we hit production ;)

Issue #524, opened 2023-07-01
Hitting the "New comment" button has no effect


fosterfarrell9:
I am currently running into the issue that sometimes hitting the "New comment" button does not work. In the browser console you can see the following error message:

```
VM2455:1 Uncaught SyntaxError: Identifier 'popoverTriggerList' has already been declared
    at processResponse (rails-ujs.source.js:288:1)
    at rails-ujs.source.js:216:1
    at xhr.onreadystatechange (rails-ujs.source.js:271:1)
```
After page reload, the button works.

PR #534, opened 2023-08-23
Outsource initialization of Bootstrap popovers (Fix "New comment" button)

Frodo161:
This branch fixes issue #524: Open the show_comment page. Hit the "new comment" button, then click "abort". Hitting the "new comment" button again won't have any effect now.

The problem is that two variables in the edit javascript file are declared as constants while hitting the button twice causes them to be redefined.

Changing the variable declaration from "const" to "var" won't cause any problems in this situation and fixes the bug.


Splines:
@Frodo161 You're change certainly gets the job done. However, the code you were touching is actually duplicated code. Bootstrap [requires](https://getbootstrap.com/docs/5.3/components/popovers/#enable-popovers) to manually initialize popovers. When we dynamically add content to the page, we have to reinitialize this. I outsourced this behavior to a new function in `app/assets/javascripts/bootstrap_popovers.js` which is called from the respective places, so that we don't have the same code all over the place.

If you are fine with this, please make sure to test that popovers work everywhere, e.g. test popovers at some places, but especially in places where content is loaded dynamically, e.g. when you add a new chapter and a modal opens for that operation that has some popovers (helptexts) in it.

app/views/quizzes/proceed.coffee
```diff
@@ -79,7 +79,6 @@ displayNext = ->
   <% else %>
   renderNext('<%= j render partial: "quizzes/quiz_round",
                            locals: { hidden: true } %>')
-  $('[data-bs-toggle="popover"]').popover()
```
Frodo161, line 82: I'm not shure, why you didn't insert an "initBootstrapPopovers()" here.
I tried a quiz with and without this additional line, but I didn't see that it changed anything in the behavior. So probably, this is fine as it is, but maybe double check this.

app/assets/javascripts/upload.coffee
```diff
@@ -439,8 +438,6 @@ directUpload provides an interface to upload (multiple) files to an endpoint
 ###
 @result = undefined
 @userManuscriptUpload = (fileInput) ->
-  # update helpdesk
-  $('[data-bs-toggle="popover"]').popover()
```
Frodo161, line 443: Why didn't you add an "initBootstrapPopovers()" here?

app/views/quizzes/proceed.coffee
Splines: forgot to add it here, fixed it (even though you didn't notice a change when deleting the line, but it doesn't harm to have it here as well)

app/assets/javascripts/upload.coffee
Splines: Oh, my bad, I forgot it, fixed it.

Frodo161 (commented):
I didn't see anything in the code (except the two missing lines you already added).

In my local MaMpf environment I tested some popovers (chapters, sections, lessons, comments and some more) and everything worked.

Splines (approved):

PR #535, opened 2023-08-26
Upgrade Puma version

Splines:
We jump up two major versions! See the release changes here:
- [4 to 5](https://github.com/puma/puma/blob/master/5.0-Upgrade.md)
- [5 to 6](https://github.com/puma/puma/blob/master/6.0-Upgrade.md)

_This is to prevent [this security issue](https://github.com/puma/puma/security/advisories/GHSA-68xg-gqqm-vgj8)._

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ gem "activerecord-nulldb-adapter"
 # Use sqlite3 as the database for Active Record
 gem "sqlite3", "~> 1.4"
 # Use Puma as the app server
-gem "puma", "~> 5"
+gem "puma", "< 7"
 # Use SCSS for stylesheets
 gem "sass-rails", ">= 6"
 # Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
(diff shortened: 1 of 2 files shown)


Frodo161 (commented):
In my local installation of MaMpf I didn't get any errors from these changes.

On https://github.com/puma/puma/blob/master/6.0-Upgrade.md (at the bottom of the page) they give a checklist with changed behavior. Actually, most variables aren't even used in MaMpf. Maybe double check the ones I found (but I guess, it won't cause any problems as it is right now):

- WEB_CONCURRENCY in config/puma.rb
- remote_addr in docker/nginx.tmpl

Splines:
Thanks.

- [x] `WEB_CONCURRENCY`: I deleted the stale comments from 6 years ago as we don't use this variable.
- [x] `remote_addr`: This is a coincidence. We don't use the [`remote_addr` config](https://github.com/puma/puma/blob/c2f73304655711219fed45fd083963174735a734/lib/puma/dsl.rb#L1021) of Puma but instead the variable [`remote_addr`](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_remote_addr) from Nginx, so this should be fine.

Splines:
Note I forgot to reflect the latest update in the lockfile, so you might want to test again (run `bundle install`) beforehand. Sorry for this inconvenience.

Frodo161 (approved):
I didn't get any errors when testing MaMpf in my local environment again.

PR #537, opened 2023-08-27
Upgrade bootstrap to v5.3.1

Splines:
**Update `bootstrap` version from `5.2.1` to `5.3.1`**, so that #527 works correctly (`z-index` did not work properly, see comment [over there](https://github.com/MaMpf-HD/mampf/pull/527#issuecomment-1694698091)).

Command used:
```
bundle update bootstrap
```

Note that `bundle update bootstrap --conservative` did not work (containers did not start due to dependency errors).

See migration guide [here](https://getbootstrap.com/docs/5.3/migration/) for what has changed in Bootstrap.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/rails/sprockets-rails
-  revision: 73e7351abff3506f6dca6b2da8abedfd5c7c0d77
+  revision: 065cbe83989c44019eca7161782ed4fdb6473517
   branch: master
   specs:
     sprockets-rails (3.4.2)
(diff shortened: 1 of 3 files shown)


Splines:
> I played around with this branch in my local installation (especially, I've sent a few mails) and everything worked.
> 
> In the migration guide of bootstrap they say that `navbar-dark` is deprecated. I found this one in "app/views/administration/_navbar.html.erb". It doesn't cause any problems in my local MaMpf, but maybe double check this.

Good catch. These were the last two occurrences of `navbar-dark` in our codebase and were removed in #536, so that's not an issue anymore.

Frodo161 (approved):

PR #543, opened 2023-10-16
Add aperiodic monotiles to landing page

Splines:
A fresh and interactive landing page look for the next releases (starting with 1.9.0 ~~#546~~ #561 ) using the newly found aperiodic monotiles, see the [official web page](https://cs.uwaterloo.ca/~csk/hat/) and the [repo with the BSD3-licensed code](https://github.com/isohedral/hatviz).

image

Different colors (which I prefer as the overall tone is more balanced):

image

## Issues

- Matrix multiplication is implemented in JS, that's why it is a bit slow. Try to hit Reset, then 2x "Monotile". The pattern will then not fill up the whole screen but is far more responsive. We could also have it initialized this way...
image

- On tablets, we cannot scroll the website anymore as the dragging of the canvas is somehow taking precedence.
- I've changed one color to red. If you don't like it, we can change it back to grey as it is [here](https://cs.uwaterloo.ca/~csk/hat/app.html).


Splines:
@fosterfarrell9 I've added the aperiodic monotiles (ein stein) as interactive background. Currently available on [experimental](https://mampf-experimental.mathi.uni-heidelberg.de/) for you to try out. More infos in the PR.

fosterfarrell9:
Is it possible to make the tiles bigger? I think that the current size makes it look a little restless/"unruhig".

Splines:
@fosterfarrell9 I've made the tiles bigger and darkened the colors to make the background less busy/jarring. Due to the bigger tiles, the numer of tiles generated are reduced which also makes the page load faster.
I've also fixed the issue on mobile devices, it should work on all devices now (buttons are disabled on small devices and you cannot drag the background, otherwise one could not scroll the site anymore). I've also implemented a small logic to disable the "monotile" button as to not freeze browsers when generating too many tiles.

image

image

app/assets/javascripts/monotile/hat.js
```diff
@@ -0,0 +1,466 @@
...
+let monotile_btn;
+let reset_btn;
+
+const colors = {
+	'H1': [0, 137, 212],
+	'H': [148, 205, 235],
+	'T': [251, 251, 251],
+	'P': [250, 250, 250],
+	'F': [191, 191, 191]
+};
```
fosterfarrell9, line 37: I have shown the new landing screen to several people. There were some complaints about the dark grey. I experimented myself a little with the colors and arrived at these values:
```javascript
const colors = {
	'H1': [0, 137, 212],
	'H': [148, 205, 235],
	'T': [251, 251, 251],
	'P': [229, 234, 241],
	'F': [230, 255, 255]
};
```
Are they better? I do not know, we would have to ask some more people.

Splines: I don't like the 'F' color where you just lowered the red value (starting off of white), which results in a blue-greenish look that is too greenish in my eyes.
image

If the argument was that the grey is too prominent, we could just replace it with a more light grey, so that the blue tiles attract more attention:

```js
const colors = {
	'H1': [0, 137, 212],
	'H': [148, 205, 235],
	'T': [251, 251, 251],
	'P': [250, 250, 250],
	'F': [220, 220, 220]
};
```

It looks like this which I think is a workable solution:

image


Other than that, I thought one opinion is that the background shouldn't be too bright and not too colorful because MaMpf "is sober and should represent clarity and focus" [you or Hendrik Kasten]?. That's why I switched to the grey in the first place to have more subdued colors (even though I'd love to see the red color, but I agree that it is too much).

What were the complaints about the grey? This color is even is built-into the Mathematikon as facing concrete everywhere, so it perfectly represents the mathematical faculty ;) Of course, color choice is very subjective in the end.

fosterfarrell9: I like the colors of the previous comment, will approve if you use them.

fosterfarrell9 (dismissed):

Splines:
@fosterfarrell9 Requesting a review just for the last two commits where I merged the `lint/eslint-all` branch into this branch. You probably want to review #575 first (e.g. maybe some rules will change due to your review and then we'd have to update here accordingly).

fosterfarrell9 (approved):

Issue #513, opened 2023-06-05
Bulk upload creates corrupt file attachments if it encounters a rollback


fosterfarrell9:
Occasionally, we are seeing these exceptions in production:

```
 :type option required

/usr/local/bundle/gems/actionpack-7.0.4.3/lib/action_controller/metal/data_streaming.rb:121:in `send_file_headers!'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/action_controller/metal/data_streaming.rb:73:in `send_file'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/action_controller/metal/instrumentation.rb:30:in `block in send_file'
/usr/local/bundle/gems/activesupport-7.0.4.3/lib/active_support/notifications.rb:206:in `block in instrument'
/usr/local/bundle/gems/activesupport-7.0.4.3/lib/active_support/notifications/instrumenter.rb:24:in `instrument'
/usr/local/bundle/gems/activesupport-7.0.4.3/lib/active_support/notifications.rb:206:in `instrument'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/action_controller/metal/instrumentation.rb:28:in `send_file'
/usr/src/app/app/controllers/submissions_controller.rb:156:in `show_correction'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/abstract_controller/base.rb:215:in `process_action'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/action_controller/metal/rendering.rb:53:in `process_action'
/usr/local/bundle/gems/actionpack-7.0.4.3/lib/abstract_controller/callbacks.rb:234:in `block in process_action'
/usr/local/bundle/gems/activesupport-7.0.4.3/lib/active_support/callbacks.rb:118:in `block in run_callbacks'
/usr/local/bundle/gems/actiontext-7.0.4.3/lib/action_text/rendering.rb:20:in `with_renderer'
/usr/local/bundle/gems/actiontext-7.0.4.3/lib/action_text/engine.rb:69:in `block (4 levels) in <class:Engine>'
/usr/local/bundle/gems/activesupport-7.0.4.3/lib/active_support/callbacks.rb:127:in `instance_exec'
[...]

PR #552, opened 2023-10-29
Prevent upload of empty submission corrections

Splines:
This is a temporary fix for #513 by preventing empty file uploads.

- During correction upload
- During bulk correction upload
- Not checked: size during submission uploads by users

### Where in the codebase

To test it in the UI, search for the string":type option"in the codebase. Once you've found the comment, set the size to `0` manually by adding

```js
data.metadata.size = 0
```

above. Then, go ahead and upload any file in [respective view](http://localhost:3000/lectures/1/tutorials). You should see a warning and not be able to complete upload.

### Backend

~~Note: this is only a UI fix, not checking this on the backend, users could still upload empty files via POST. We might still want to ship this to prevent basic errors in the frontend during upload and to maybe find the reason for #513.~~
We are also checking for the minimum file size in the backend. To test it, change

```js
if data.metadata.size == 0
```

 to 

```js
if false
```

Then, users are able to upload the empty file in the frontend, but when they click on "Save", it will fail (with the same error message), as the backend validation fails.

### Dummy file

Use the following dummy file as an empty file to test with: [empty-file.pdf](https://github.com/MaMpf-HD/mampf/files/13206767/empty-file.pdf)


fosterfarrell9 (approved):

app/assets/javascripts/upload.coffee
```diff
@@ -291,6 +306,16 @@ bulkCorrectionUpload = (fileInput) ->
...
+
+        # Disable upload if uploaded (zip)-file is empty
+        # Temporary fix for ":type option required" error in the
+        # app/controllers/submissions_controller.rb -> show_correction
+        if data.metadata.size == 0
+          console.log('empty file encountered during bulk upload')
+          alert($('#bulk-uploadButton-button-actual').data('tr-failure-empty-file'))
+          $('#bulk-uploadButton-button-actual').hide()
+          return
+
```
fosterfarrell9, line 318: I played around with some sample files and observed the following. First of all, bulk upload means uploading several files at once using a name pattern that MaMpf recognizes (e.g. keep the names from the bulk download). In my tests, I bulk-uploaded corrections for two students. It appears that `data.metadata` contains only information on the file that was uploaded last. So, in my tests, if the zero-size file had a name that came alphabetically before the name of the non-zero-size file, then the UI did not catch the zero-size. (It did if the zero-size file came last). Of course, the server-sided validation caught the zero-size file when the UI did not.

Splines: Found the bug. The `directUpload` only passed the last uploaded file to the `successCallback`. I fixed this but didn't want to break any other usage of `directUpload`; this is why I've introduced a second parameter for this method which is only used in the callback inside `bulkCorrectionUpload`. This is not really best practice, but if I wanted to fix this more thoroughly, I'd have to go through more parts of the file, which is nothing I like to do in Coffeescript. But we should keep this in mind, convert the `upload.coffee` to JS and greatly clean up, it's really a mess with lots of hardcoded strings, intermingled functionality and lots of states variables throughout the file.

**With the fix in the latest commit, please try to upload files again**. You should also get a message that tells you which file is empty. (If multiple files are empty, you get this info just for one file but can then progressively get rid of the empty files).

Also, the "Cancel" button of the bulk upload does not work, which is something for another PR.

fosterfarrell9 (approved):
I did some testing in my local setup, and it seems to be working nicely.

PR #560, opened 2023-11-12
Sensible configuration for rubocop

fosterfarrell9:
The aim of this PR is to figure out a sensible configuration for rubocop in our project. Until now, we just used the rubocop config file of Rails. Looking at the details of what happens there, you see that many of these settings do not make sense for us.
In a first attempt to improve this, I incorporated Shopify's rubocop settings which are [well documented](https://github.com/Shopify/ruby-style-guide) and made some minor twists for settings where we have (until now) used different defaults. Note that in the shopify settings all `Metrics/` methods of rubocop are disabled.
I put this first attempt up for discussion. If the general approach is considered okay, we might discuss where we deviate from shopify's config.


fosterfarrell9:
This is now ready for review. I added a [Style Guide Page](https://github.com/MaMpf-HD/mampf/wiki/Style-Guide) in the wiki.

.rubocop.yml
```diff
@@ -1,341 +1,9 @@
...
-Metrics/ModuleLength:
-  Enabled: false
-
-Metrics/ClassLength:
-  Enabled: false
-
-Metrics/MethodLength:
-  Exclude:
-    - 'app/abilities/**/*'
+  EnforcedStyle: never
```
Splines, line 8: From the [RuboCop Docs](https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration):

> Pay attention to the comments above each entry. They can reveal configuration parameters such as EnforcedStyle, which can be used to modify the behavior of a cop instead of disabling it completely.

We want to get rid of it completely and not even have warnings about this, right? Then, I'd favor: `Enabled: false`.

.rubocop.yml
```diff
@@ -1,341 +1,9 @@
...
-Layout/EmptyLinesAroundModuleBody:
-  Enabled: true
-
-Layout/ExtraSpacing:
-  Enabled: true
+inherit_gem:
+  rubocop-shopify: rubocop.yml
 
 Layout/LineLength:
   Max: 80
```
Splines, line 5: Of course, this limit is [highly controversial](https://stackoverflow.com/questions/110928/is-there-a-valid-reason-for-enforcing-a-maximum-width-of-80-characters-in-a-code).

I strongly recommend `80` characters as a **soft limit** and we should include that in our style guide, e.g. in the Wiki. In VSCode you can also display a grey border at 80 chars.

image

However, in my past dev experience, I remember lines that were slightly longer than 80 chars and I was really upset that a linter forced me to break a statement to two lines. Sometimes this can actually make a statement less readable (given that the line was just a bit above 80 chars and not something like 120 chars or even more).

This is probably why they even got rid of the 80-column warning in the Linux kernel [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144). Instead, they raised the limit to `100` chars which I think is more reasonable as a hard limit.

Splines (changes requested):
Some comments.

Thanks for the [Wiki entry](https://github.com/MaMpf-HD/mampf/wiki/Style-Guide). I changed the link there to point to the [webpage](https://ruby-style-guide.shopify.dev/) that shows the rendered markdown files (and not just the GitHub Markdown which does not support many things they are using, e.g. colors for headers, nice code formatting and so on).

.rubocop.yml
Splines: This is also what Shopify uses to disable cops.

.rubocop.yml
```diff
@@ -1,341 +1,9 @@
...
-  Enabled: false
-
-Metrics/MethodLength:
-  Exclude:
-    - 'app/abilities/**/*'
+  EnforcedStyle: never
 
-Metrics/AbcSize:
-  Exclude:
-    - 'app/abilities/**/*'
```
Splines, line 341: We currently have _two_ empty newlines at the end of the `.rubocop.yml` file. Remove one of those newlines.

Splines (changes requested):

Splines:
For a future PR that aims to align all Ruby files in our entire codebase to the new Rubocop style guides, [this](https://stackoverflow.com/a/63842073/) could be a very useful comment.

.rubocop.yml
Splines: Ah no, my bad, the `EnforcedStyle: never` will even ensure that RuboCop will complain when we put `frozen_string_literal` somewhere. With `Enabled: false`, we don't get any  warning for it whatsoever.

You can therefore resolve this conversation.

Gemfile
```diff
@@ -119,9 +119,7 @@ group :development, :docker_development do
   gem "spring"
   gem "spring-watcher-listen", "~> 2.0.0"
   gem "rubocop", "~> 1.50", require: false
-  gem "rubocop-packaging", require: false
-  gem "rubocop-performance", require: false
-  gem "rubocop-rails", require: false
+  gem "rubocop-shopify", require: false
```
Splines, line 122: Use a version specifier here. Otherwise, we might suddenly have breaking changes.

```suggestion
  gem "rubocop-shopify", "~> 2.14", require: false
```

Gemfile.lock
```diff
@@ -508,15 +508,8 @@ GEM
...
-    rubocop-packaging (0.5.1)
-      rubocop (>= 0.89, < 2.0)
-    rubocop-performance (1.10.2)
-      rubocop (>= 0.90.0, < 2.0)
-      rubocop-ast (>= 0.4.0)
-    rubocop-rails (2.9.1)
-      activesupport (>= 4.2.0)
-      rack (>= 1.1)
-      rubocop (>= 0.90.0, < 2.0)
+    rubocop-shopify (2.13.0)
```
Splines, line 511: Newest version is `2.14`, not `2.13`.

Splines (changes requested):

Gemfile
Splines: Note that this will also update `rubocop` from `1.50.2` to `1.52.1`.

Splines: I think this is a good time to get some versions right. Let's update rubocop and rubocop-shopify together:

```rb
gem "rubocop", "~> 1.57", require: false
gem "rubocop-shopify", "~> 2.14", require: false
```

However, `bundle install` will then fail due to `erb_lint`. Let's deal with that in a future PR. For now, we just want to get Rubocop to work. Therefore, remove the line

```rb
gem "erb_lint", require: false
```

Then running `bundle install` should work just fine.

.rubocop.yml
Splines, line 1: What about [`Style/ClassMethodsDefinitions`](https://github.com/Shopify/ruby-style-guide/blob/v2.14.0/rubocop.yml#L443-L445), do we really want that?

It will enforce [this](https://docs.rubocop.org/rubocop/cops_style.html#enforcedstyle-self_class) affecting us 132 times in our codebase. I'd rather go with [`EnforcedStyle: def_self`](https://docs.rubocop.org/rubocop/cops_style.html#enforcedstyle-def_self-default).

Splines (changes requested):

Gemfile
fosterfarrell9: Done. Note that updating the rubocop version also triggered some updates of rubocop dependencies, see `Gemfile.lock`.

.rubocop.yml
fosterfarrell9: Agree and done.

fosterfarrell9: I also updated the wiki page correspondingly.

Splines, line 1: [`Style/CaseLikeIf`](https://docs.rubocop.org/rubocop/cops_style.html#minbranchescount-3-default) is disabled in Shopify. What do you think about adding a [`Severity: refactor`](https://docs.rubocop.org/rubocop/configuration.html#severity) to it?

Splines (changes requested):

.rubocop.yml
Splines: Ok thanks. I wonder: if we add more custom things to the `.rubocop.yml` file, maybe it's a bit tedious to keep it synced with the Wiki? Except for the link to the style guide, I do not see the benefit of listing every change we make in the Wiki. 

One could just open the `.rubocop.yml` file (maybe link to that in the Wiki?) or even run rubocop on the local code changes to directly see where one violates our styles. In the best case, the plugin for VSCode also adds blue squiggly lines to indicate style violations. In my setup, it even autocorrects these (if the autocorrections are safe) whenever I save the file.

If you argue that the Wiki is a place where we could add additional comments for *why* we chose to do something, we could also have that in the [`Details`](https://docs.rubocop.org/rubocop/configuration.html#details) section.

Splines, line 1: What's your opinion on [`Style/GuardClause`](https://docs.rubocop.org/rubocop/cops_style.html#styleguardclause) which is disabled in Shopify. I don't have a strong opinion on this but remember you once suggested to use `unless`, so you probably want it enabled?

Splines, line 1: I strongly vote to enable the [`Style/DoubleNegation`](https://docs.rubocop.org/rubocop/cops_style.html#styledoublenegation) cop.

Splines, line 1: Instead of disabling [`Style/EmptyMethod`](https://docs.rubocop.org/rubocop/cops_style.html#styleemptymethod), we should enforce our preferred style, which is [`EnforcedStyle: expanded`](https://docs.rubocop.org/rubocop/cops_style.html#enforcedstyle-expanded).

Splines, line 1: I strongly suggest enabling [`Style/NegatedUnless`](https://docs.rubocop.org/rubocop/cops_style.html#stylenegatedunless), otherwise you have to write down your own truth table to find out when the `if` statement is `true`.

Splines, line 1: I strongly suggest enabling [`Style/NegatedIfElseCondition`](https://docs.rubocop.org/rubocop/cops_style.html#stylenegatedifelsecondition).

Splines, line 1: I suggest enabling the cop [`Style/YodaCondition`](https://docs.rubocop.org/rubocop/cops_style.html#styleyodacondition) with the default `EnforcedStyle`.

Splines, line 1: I strongly suggest enabling the cop [`Style/TrailingMethodEndStatement`](https://docs.rubocop.org/rubocop/cops_style.html#styletrailingmethodendstatement). Why would one disable that? I see no scenario where that could be a useful thing to do.

Splines, line 1: I strongly suggest enabling the cop [`Style/TrailingMethodEndStatement`](https://docs.rubocop.org/rubocop/cops_style.html#styletrailingbodyonmethoddefinition).

Splines, line 1: I prefer to have [`Style/SwapValues`](https://docs.rubocop.org/rubocop/cops_style.html#styleswapvalues) enabled as this makes the intention of what is really done in the code clearer.

Splines (changes requested):

.rubocop.yml
Splines, line 1: I suggest not disabling any security-related RuboCop cop like [Shopify does](https://github.com/Shopify/ruby-style-guide/blob/v2.14.0/rubocop.yml#L403C10-L413). This is since I don't feel like I'm in the expert position to judge if we don't need this, so I'd prefer having this as an error that signifies to me I really have to watch out and should probably use a different way to do things (since smart people already figured this could be a potential security issue).

I even suggest writing the following to enable all security-related Rubocop cops.
```rb
Security:
  Enabled: true
```

Splines, line 1: I don't see a use-case for self-assignment, so I'd suggest to enable the cop [`Lint/SelfAssignment`](https://docs.rubocop.org/rubocop/cops_lint.html#lintselfassignment).

Splines, line 1: I suggest enabling the cop [`Lint/EmptyBlock`](https://docs.rubocop.org/rubocop/cops_lint.html#lintemptyblock).

Splines, line 1: I suggest enabling the cop [`Lint/DuplicateRequire`](https://docs.rubocop.org/rubocop/cops_lint.html#lintduplicaterequire).

Splines (changes requested):

Splines:
To be honest: meanwhile I feel that for almost every cop they disabled at Shopify, I want to enable it again and disable it only if we encounter a use-case in our *own* code base. Maybe a better strategy would be to go with the Rubocop defaults; it has many cops (which I suggested to enable in this PR) enabled *by default*.

I mean we could still disable some things such as many metric things that tell us about too long methods and the like. In the end, Rubocop should serve us and not vice-versa. When starting fresh, it might be very helpful to run the Rubocop autocorrect locally, see what things it proposes to change. One can then skim through the changes and if we don't like a change, we can deactivate it.

The command I use is:

```rb
```

Then take a look at the changes and also search for `rubocop:todo` for things that could not be autocorrected.

Splines:
@fosterfarrell9 I've tried to reduce our existing `.rubocop.yml` file by using the fact that RuboCop defaults are [already imported by default](https://github.com/rubocop/rubocop/blob/master/config/default.yml), so we only have to worry about specifying where we deviate from that. For this, I've opened a new draft PR #564 to try things out.

fosterfarrell9:
You are certainly right. There are even more things where shopify deviates from Robocop's default settings that I do not like. We should not use the shopify settings as a basis. I am therefore closing this in favor of #564.

PR #561, opened 2023-11-12
Continuous Release 1.9.0

Splines:
> [!warning]
> Ok, so I'm not sure why, but GitHub shows 4 commits up until "Add pgadmin..." that don't actually belong here. Instead, you can already find these commits on the main branch. This issue is exactly described [here](https://stackoverflow.com/q/16306012/9655481) on Stack Overflow and has been viewed over more than 270k times. It seems like it is a limitation of GitHub. However, `git log --oneline main..mampf-next` also shows these commits in the diff and I'm not sure what we did wrong to end up in this weird situation.
<br>1) I tried the [rebasing solution](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch/45286781#45286781) in #579 and it was working perfectly fine and removed the 4 commits. The only downside of this approach is that due to the rebasing all the new commits of mampf-next would have the same timestamp: the time of performing the rebase. The will actually even get reassigned a new commit id.
<br>2) If this first solution is not what we want, we could instead ignore the additional commits in this PR. The merge should work just fine. As we wanted to rename `mampf-next` to `dev` anyways, we could instead create this new branch `dev` from `main` after release 1.9.0 directly and ditch `mampf-next`. This way, we could be sure that `main` and `dev` match exactly and have the same commits on them.
<br>No matter what option we choose, we will have a clean slate again starting with the next continuous release. <sub>Due to the nature of merge commits though, `dev` might be one commit behind `main` as it doesn't include the merge commit itself.  See also [this SO answer](https://stackoverflow.com/a/70715962/) which explains this. That's not an issue though. One can avoid this by merging from main into dev directly after a continuous release such that dev can be fast-forwarded.</sub>

## Changes

Also see #565 for a background on the linting changes.
- #543
- #553
- #558 
- #559
- #560
- #562 
- #564 
- #566 
- #568 
- #569 
- #575 
- #577 
- #578
- #582

## For future releases?
- #574 

---

- [x] check `schema.rb` has correct verison


Splines:
Currently, all existing model tests pass except for `quiz_certificate_spec.rb`. Already the first test there ("has a valid factory") fails with this error:

```
ActiveRecord::InverseOfAssociationNotFoundError:
Could not find the inverse association for quiz (:quiz_certificate in Medium) Did you mean?  quiz_certificates
```

You can verify that by your own on #581 which has all the changes from here merged into it.

fosterfarrell9 (approved):

Splines:
For completeness: we've decided to go with option 2 of [this comment](https://github.com/MaMpf-HD/mampf/pull/561#issue-1989455951).

PR #564, opened 2023-11-15
Redo the Rubocop config

Splines:
Related:
- #560
- #562 

**See the new [`.rubocop.yml`](https://github.com/MaMpf-HD/mampf/blob/feature/redo-rubocop-config/.rubocop.yml) from this PR.**

[Automatically Generated Configuration](https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration). And some useful commands.

```
bundle exec rubocop --autocorrect (maybe also use option: --disable-uncorrectable)

rubocop --auto-gen-config --no-exclude-limit
[work through .rubocop_todo.yml]
rubocop --regenerate-todo
```

---

Some stats using the new config at commit e768bfc:

```sh
$ bundle install

# First: Safe autocorrects
$ bundle exec rubocop --autocorrect 
704 files inspected, 7952 offenses detected, 6814 offenses corrected, 721 more offenses can be corrected with `rubocop -A`

# Second: Unsafe autocorrects
$ bundle exec rubocop -A
704 files inspected, 1350 offenses detected, 938 offenses corrected

# Then: generate todo rubocop file
$ rubocop --auto-gen-config --no-exclude-limit
704 files inspected, 413 offenses detected
Created .rubocop_todo.yml.
```

This newly created todo file "only" contains 37 offenses that couldn't be autocorrected (although for some we get the comment they can be using unsafe autocorrection, but we've done that in the step beforehand?! -> don't know why this is the case). This file can be found [here](https://gist.github.com/Splines/509c1ddff374f2d1cbf2c4a86e53b795), maybe we want to disable some of the cops from there already...

(Note: of course one also has to watch our carefully when running rubocop with unsafe autocorrects as this could possibly modify the code in a way that is semantic and not just syntactic. A strategy for a future PR could therefore consist of running the safe autocorrects first and commit, then do a separate commit for just unsafe autocorrects. This will ease the review process. Finally, we have the rubocop todo file that we can either directly try to work off, or do it from time to time, e.g. in another PR).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,341 +1,80 @@
+# SEVERITY LEVELS
+# In the GitHub CI/CD, we can only distinguish between "error" and "warning".
+# In Rubocop this is much more granular: "info", "refactor", "convention",
+# "warning", "error" and "fatal".
+# From the docs: "The level is normally 'warning' for Lint and 'convention' for
+# all the others, but this can be changed in user configuration."
+# However, we don't want to set the severity for each cop individually, so instead
+# we set the fatal level as "convention" in the CI/CD,
+# i.e. severity: "convention" and up will be treated as an error on GitHub
+# and below as warning such that the check will still pass (with a warning).
+#
+# Short:
+# "convention" severity and up = Error on GitHub (CI/CD) -> check fails
+# "refactor" severity and below = Warning on GitHub (CI/CD) -> check passes
+#
+# see the docs on severity
+# https://docs.rubocop.org/rubocop/configuration.html#severity
+
+# Rubocop defaults are by default required/included:
+# https://github.com/rubocop/rubocop/blob/master/config/default.yml
 require:
-  - rubocop-packaging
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -1,341 +1,87 @@
...
-
-Style/HashTransformValues:
-  Enabled: true
-
-Style/RedundantBegin:
-  Enabled: true
+Style/StringLiteralsInInterpolation:
+  EnforcedStyle: double_quotes
 
 Style/RedundantReturn:
```
fosterfarrell9, line 76: I suggest adding the following two rules:

```
Style/WordArray:
  EnforcedStyle: brackets

Style/EmptyMethod:
  EnforcedStyle: expanded
```

Splines:
@fosterfarrell9 Apart from the comment you gave, do you think we can merge?

fosterfarrell9 (approved):

PR #566, opened 2023-11-25
Enforce new RuboCop cops for all Ruby files

Splines:
Rubocop version information:
```
rubocop 1.57.2 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.1.4) [x86_64-linux]
```

Commands used:
```sh
bundle exec rubocop --autocorrect-all --disable-uncorrectable # used later
```

## Notable manual changes
- Use `Time.current` instead of `Time.now`. [Ref](https://discuss.rubyonrails.org/t/time-now-vs-time-current-vs-datetime-now/75183)
- Use `update` instead of `update_all`. [Ref](https://stackoverflow.com/a/45056430/9655481)

## Notable manual changes (explicit) / Questions from my side
Current status (2023-12-03 15:30):
- 30 `rubocop:todo Rails/` cops left
- 1 `rubocop:todo Naming/` cop left -> easily fixable
- 3 `rubocop:todo Lint/` cops left
- 2 `rubocop:todo Metrics/` cops left -> probably won't fix in this PR (will stay as todo note)

Some open points or commits to take especially care of (please mark as checked):
- [x] `registrations_controller.rb` -> `check_registration_limit`
- [x] Disabled line length in multiple occurrences in `user_cleaner.rb` for long regex statements.
- [x] Deleted some lines in `Devise.setup` (see Basecamp)
- [x] `shrine.rb` -> added `submission_path`
- [x] `quizzes_controller.rb` -> `init_values` (`quiz_round_params[:save_probe]`)
- [x] Duplicate methods in `medium.rb´ (one still has to be fixed `is_restricted?`)
- [x] `submissions.rb`: are the methods at the bottom really supposed to be private class methods?
- [x] `db.rake` -> `ensure_format`
- [x] Why do we rescue the `StandardError` in `registrations_controller` as well as in `log_fail`? Please provide a comment there
- [x] Why do we have a `has_documents?` method in `assignments_helper` if we have the same method in `assignment.rb` (over there: renamed to `documents?`
- [x] `user.rb` -> The method `has_tutorials` was never used (I've made a search for that word), so I deleted it
- [x] 12 `Rails/HasManyOrHasOneDependent` todos -> won't fix on my own.
- [x] I fixed the `Rails/InverseOf` cops but am not 100% sure that this is the way it should work. I followed the examples from [here](https://docs.rubocop.org/rubocop-rails/cops_rails.html#examples-62) but have a feeling summarized well by [this comment](https://stackoverflow.com/questions/9296694/what-does-inverse-of-do-what-sql-does-it-generate#comment14954709_9297745).
- [ ] ~~12 `Rails/UniqueValidationWithoutIndex` -> won't fix on my own and probably not in this PR as it may require changing db columns, i.e. add new migrations, right?~~
- [x] In `course.rb` -> `question_ids_for_quiz`, there is an unused variable. Should we return it?
- [x] `submission.rb` -> `filename_for_bulk_download`
- [ ] Security fixes in [this commit](TODO)

## Notable unsafe corrections (incomplete list!)
Note that autocorrects running on old migrations were not committed.

- [x] `video_uploader.rb` -> `pp` changed to `Rails.logger.debug`
- [ ] ...

## Before merging
- [x] Search for unwanted `rubocop:enable` that don't have matching `rubocop:todo` or `rubocop:enable` block beforehand (rubocop can search for that, it is a cop on its own)
- [x] `bundle exec rubocop -A` should not show any warning/error any more (see also CI/CD pipeline)

## Todo
- [x] Enable [`MethodCallWithArgsParentheses`](https://docs.rubocop.org/rubocop/1.58/cops_style.html#stylemethodcallwithargsparentheses)


app/controllers/answers_controller.rb
```diff
@@ -9,7 +9,7 @@ def current_ability
 
   def new
     question = Question.find_by_id(params[:question_id])
-    @answer = Answer.new(value: true, question: question)
+    @answer = Answer.new(value: true, question:)
     authorize! :new, @answer
```
fosterfarrell9, line 13: What rule is it that replaces `@answer = Answer.new(value: true, question: question)`
by `@answer = Answer.new(value: true, question:)`? I did not even know that this was possible. I also think that this is bad for the readability of the code.

Splines: Oh, good point. This is due to [`EnforcedShorthandSyntax`](https://docs.rubocop.org/rubocop/1.57/cops_style.html#enforcedshorthandsyntax-never). We can set it to `never` in order to enforce that `question: question` is used instead of `question:`.

app/controllers/application_controller.rb
```diff
@@ -96,26 +94,26 @@ def store_user_location!
...
 
       study_participant = current_user.anonymized_id if current_user.study_participant
       # as of Rack 2.0.8, the session_id is wrapped in a class of its own
       # it is not a string anymore
       # see https://github.com/rack/rack/issues/1433
       InteractionSaver.perform_async(request.session_options[:id].public_id,
                                      request.original_fullpath,
-                                     request.referrer,
+                                     request.referer,
                                      study_participant)
```
fosterfarrell9, line 117: Fun fact: The corect spelling is actually referrer, but is not used anymore; [https://en.wikipedia.org/wiki/HTTP_referer#Etymology](https://en.wikipedia.org/wiki/HTTP_referer#Etymology)

Splines:
Hi @fosterfarrell9, may I ask you to fix the remaining `rubocop:todo` in the list (see above) and commit them individually (e.g. one commit for the three linting cop violations, but probably a few more commits for the Rails related cops, etc.)?

Also, please review my changes. Of course not in the "Files changed" tab as this is too tedious with this many files. But maybe you could go click on each commit that only carries out *manual* fixes and check what I've done there. Please mark the tasks as completed accordingly in the above list ;)

Note that the last commit now also passes the CI/CD pipeline. This means, there are no rubocop violations anymore that haven't been marked as todo or have been disabled 🎉 So, we only have the TODOs in the above list left to tackle.

Note that I've disabled the rules for the already existing migrations as I feel like the migrations should not be altered as they are kind of like an archive. But I didn't want to exclude the `db/` folder completely from rubocop as we might find rubocop useful for migrations we create in the future.

app/controllers/registrations_controller.rb
```diff
@@ -33,50 +33,57 @@ def create
...
-      end
+      timeframe = (ENV["MAMPF_REGISTRATION_TIMEFRAME"] || 15).to_i.minutes
+      threshold_date = DateTime.now - timeframe
+      new_users = User.where(
+        "users.confirmed_at is NULL and users.created_at > '#{threshold_date}'"
+      ).count
+
+      max_users = (ENV["MAMPF_MAX_REGISTRATION_PER_TIMEFRAME"] || 40).to_i
+      return if new_users > max_users
+
```
fosterfarrell9, line 81: Are you sure you want `>` this way around? The way I read the old code, the statements following this line were executed if `new_users > max_users` which would imply that in the guard clause we have to `return if new_users < max_users` (or `<=` if you want to exactly preserve the old behaviour).

Splines: Oh, I should've looked more carefully here. You are absolutely right. Just fixed it with commit cb2fa144ef.

Splines:
Note: the current `settings.json` that might be useful in conjunction with Rubocop is available [here](https://gist.github.com/Splines/aecf55785648a532aa11b707c3f246e4).

app/controllers/registrations_controller.rb
```diff
@@ -72,13 +72,14 @@ def after_sign_up_path_for(_resource)
     def check_registration_limit
       timeframe = (ENV["MAMPF_REGISTRATION_TIMEFRAME"] || 15).to_i.minutes
       threshold_date = DateTime.now - timeframe
-      new_users = User.where(
+      num_new_registrations = User.where(
         "users.confirmed_at is NULL and users.created_at > '#{threshold_date}'"
       ).count
```
fosterfarrell9, line 77: If we are very strict: This query violates sveral cops, but this is probably not recognized by RuboCop due to the complexity:
https://github.com/rubocop/rails-style-guide#avoid-interpolation
https://github.com/rubocop/rails-style-guide#hash-conditions
https://github.com/rubocop/rails-style-guide#where-ranges

fosterfarrell9: Perhaps something like
```
timeframe = (ENV["MAMPF_REGISTRATION_TIMEFRAME"] || 15).to_i.minutes.ago
num_new_rgistration = User.where(confirmed_at: nil, created_at: timeframe..).count
````
or similar would be better?

File: app/models/medium.rb [hunk 5]
```diff
@@ -726,7 +720,7 @@ def locked?
     released == "locked"
   end
 
-  def restricted? # rubocop:todo Lint/DuplicateMethods
+  def restricted?
```
fosterfarrell9, line 723: Can you confirm my observation that with the removal of the first `restricted?` method, the `restricted?` methods for courses and lessons are no longer needed (whereas the `restricted?` method for lectures is still in use).

app/models/medium.rb
```diff
@@ -393,12 +393,6 @@ def self.similar_courses(search_string)
     end
   end
 
-  def restricted?
-    return false unless teachable
-
-    teachable.restricted?
-  end
```
fosterfarrell9, line 399: Can you confirm my observation that with the removal of the first `restricted?` method, the `restricted?` methods for courses and lessons are no longer needed (whereas the `restricted?` method for lectures is still in use)?

fosterfarrell9:
When I start the container for this PR, I get two errors:

```
development-mampf-1        | NameError: undefined local variable or method `course' for Course:Class
development-mampf-1        | /usr/local/bundle/gems/activerecord-7.0.4.3/lib/active_record/dynamic_matchers.rb:22:in `method_missing'
development-mampf-1        | /usr/src/app/app/models/course.rb:18:in `<class:Course>'
development-mampf-1        | /usr/src/app/app/models/course.rb:2:in `<main>'
development-mampf-1        | /usr/local/bundle/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
development-mampf-1        | /usr/local/bundle/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/kernel.rb:30:in `require'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/helpers.rb:135:in `const_get'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/helpers.rb:135:in `cget'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:175:in `block in actual_eager_load_dir'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/helpers.rb:40:in `block in ls'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/helpers.rb:25:in `each'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/helpers.rb:25:in `ls'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:170:in `actual_eager_load_dir'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:17:in `block (2 levels) in eager_load'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:16:in `each'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:16:in `block in eager_load'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:10:in `synchronize'
development-mampf-1        | /usr/local/bundle/gems/zeitwerk-2.6.11/lib/zeitwerk/loader/eager_load.rb:10:in `eager_load'
development-mampf-1        | /usr/local/bundle/gems/railties-7.0.4.3/lib/rails/autoloaders.rb:32:in `each'
development-mampf-1        | /usr/local/bundle/gems/railties-7.0.4.3/lib/rails/application.rb:498:in `eager_load!'
development-mampf-1        | /usr/local/bundle/bundler/gems/sunspot-414a59413cb7/sunspot_rails/lib/sunspot/rails/tasks.rb:33:in `block (3 levels) in <main>'
development-mampf-1        | /usr/local/bundle/bundler/gems/sunspot-414a59413cb7/sunspot_rails/lib/sunspot/rails/tasks.rb:70:in `with_session'
development-mampf-1        | /usr/local/bundle/bundler/gems/sunspot-414a59413cb7/sunspot_rails/lib/sunspot/rails/tasks.rb:19:in `block (2 levels) in <main>'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/cli/exec.rb:58:in `load'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/cli/exec.rb:58:in `kernel_load'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/cli/exec.rb:23:in `run'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/cli.rb:483:in `exec'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/cli.rb:31:in `dispatch'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/cli.rb:25:in `start'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/exe/bundle:48:in `block in <top (required)>'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/lib/bundler/friendly_errors.rb:103:in `with_friendly_errors'
development-mampf-1        | /usr/local/bundle/gems/bundler-2.3.13/exe/bundle:36:in `<top (required)>'
development-mampf-1        | /usr/local/bundle/bin/bundle:25:in `load'
development-mampf-1        | /usr/local/bundle/bin/bundle:25:in `<main>'
development-mampf-1        | Tasks: TOP => sunspot:solr:reindex => sunspot:reindex
```

and

```
2023-12-07T15:01:08.439Z pid=55 tid=qpj WARN: NameError: uninitialized constant Net::IMAP
development-mampf-1        |
development-mampf-1        |     @imap = Net::IMAP.new(ENV.fetch("IMAPSERVER", nil), port: 993, ssl: true)
development-mampf-1        |                ^^^^^^
```

app/models/course.rb
```diff
@@ -15,7 +15,7 @@ class Course < ApplicationRecord
   # rubocop:todo Rails/HasManyOrHasOneDependent
   has_many :media, -> { order(position: :asc) },
            as: :teachable,
-           inverse_of: course
+           inverse_of: :course
```
fosterfarrell9, line 18: Note that this is (while syntactically correct now) still not correct. Calling course.media now gives an exception:

```
/usr/local/bundle/gems/activerecord-7.0.4.3/lib/active_record/reflection.rb:236:in `check_validity_of_inverse!': Could not find the inverse association for media (:course in Medium) (ActiveRecord::InverseOfAssociationNotFoundError)
```

fosterfarrell9: I changed it in the latest commit to `inverse_of: :teachable`.
This seems to work. See also [here](https://makandracards.com/makandra/47200-association-to-polymorphic-model-does-not-determine-inverse_of-automatically), although in contrast to what is written there, saving also works without the additional` inverse_of`.

app/helpers/application_helper.rb
```diff
@@ -291,10 +292,10 @@ def realization_path(realization)
   def first_course_independent?
     current_user.administrated_courses
                 .natural_sort_by(&:title)
-               &.first&.term_independent
+                &.first&.term_independent
   end
 
-  def get_announcements
```
fosterfarrell9, line 297: Since we include the `ApplicationHelper` in `lecture.rb`, renaming this class to `announcements` has the consequence that this method overrides the `announcements` method of the `lecture` model which is given by the `:has_many announcements` relation.

fosterfarrell9:
Note that we have an infinite trigger cascade going on now: Just open some tag edit page like `/tags/5/edit` and add an alias to the tag. You can see in the logs e.g. that a method like `touch_media` is called ad infinitum. I would suspect that the reason is because statements like `update_all(updated_at: Time.now)` within these have been replaced by `update(updated_at: Time.current)`. The `update_all` statement had been in place deliberately, because it does not trigger ActiveRecordCallbacks, see [here](https://apidock.com/rails/ActiveRecord/Relation/update_all) whereas `update` does, setting the trigger cascade in motion.
@Splines Is there way to specifically undo the `update_all` replacements?

app/models/tag.rb
```diff
@@ -323,27 +325,27 @@ def cytoscape_edge(related_tag)
...
-    super + '-' + I18n.locale.to_s
+    "#{super}-#{I18n.locale}"
   end
 
   def touch_lectures
-    Lecture.where(id: sections.map(&:lecture)
-                              .map(&:id)).update_all updated_at: Time.now
+    Lecture.where(id: sections.map { |section| section.lecture.id })
+           .update(updated_at: Time.current)
   end
```
fosterfarrell9, line 340: All changes from `update_all` to `update` are extremely problematic and introduce unwanted behaviour. The reason is that `update` triggers Active Record Callbacks whereas `update_all` does not, see [here](https://apidock.com/rails/ActiveRecord/Relation/update_all). If you use `update` here and in the other places where it replaced `update_all`, you will get infinite update loops. Just try `/tags/5/edit`
and add an alias to the tag. This loop is easily explained: The `touch_lectures` method now triggers updates for the associated lectures, which (among others) triggers the `touch_lessons` after save callback on the `lecture` model which then triggers the `touch_media` after save callback on the lesson model, and so on and so on. The `update_all` was placed deliberately in the places where it was before to prevent such cascades. There seems to be a `touch_all` method that [does exactly what we want](https://api.rubyonrails.org/v7.0.4/classes/ActiveRecord/Relation.html#method-i-touch_all). I would suggest we use this in all instances where we had an `update_all(updated_at: Time.now)` before and add `touch_all` to the list of allowed methods on Rails/SkipsModelValidations. Beforehand, we have to undo exactly these replacements of `update_all` to `update`. Is there a simple way to do this?

app/models/course.rb
```diff
@@ -14,7 +12,11 @@ class Course < ApplicationRecord
            after_remove: :touch_tag,
            after_add: :touch_tag
 
-  has_many :media, -> { order(position: :asc) }, as: :teachable
+  # rubocop:todo Rails/HasManyOrHasOneDependent
```
fosterfarrell9, line 15: I think we discussed that we want to disable this cop, right?

app/helpers/notifications_helper.rb
```diff
@@ -38,34 +36,34 @@ def notification_header(notification)
...
+      t("notifications.course_selection")
     elsif notification.lecture_announcement?
       announcement_notification_card_header(notifiable)
     else
-      link_to t('mampf_news.title'), news_path, class: 'text-dark'
+      link_to(t("mampf_news.title"), news_path, class: "text-dark")
     end
-    text.html_safe
+    text.html_safe # rubocop:todo Rails/OutputSafety
   end
```
fosterfarrell9, line 46: I am not sure what we are suppposed to do here. In three of the cases that happen, `text` will be a `link_to` (`medium_notification_card_header` and `announcement_notification_card_header` return such) which is safe from my understanding, and in the remaining case it is a string from our local file. Maybe we could do this?

```ruby
  def notification_header(notification)
    notifiable = notification.notifiable
    if notification.medium?
      medium_notification_card_header(notifiable)
    elsif notification.course? || notification.lecture?
      t("notifications.course_selection").html_safe
    elsif notification.lecture_announcement?
      announcement_notification_card_header(notifiable)
    else
      link_to(t("mampf_news.title"), news_path, class: "text-dark")
    end
  end
```

There is this [gist](https://gist.github.com/joekur/73779c40c481a2f8a44f) which might be helpful.

app/helpers/notifications_helper.rb
```diff
@@ -38,34 +36,34 @@ def notification_header(notification)
...
       course_notification_card_text(notifiable)
     elsif notification.lecture?
       lecture_notification_card_text(notifiable)
     else
-      t('notifications.new_announcement')
+      t("notifications.new_announcement")
     end
-    text.html_safe
+    text.html_safe # rubocop:todo Rails/OutputSafety
   end
```
fosterfarrell9, line 61: Similar to the above, only this time `course_notification_card_text` and `lecture_notification_card_text` are constructed from the locales, but this time with parameters. It is unclear to me if it would be okay add an html_safe within both of these methods (and  to both `t('...')` in `notification_text`.)

app/helpers/notifications_helper.rb
```diff
@@ -76,13 +74,13 @@ def notification_link(notification)
     else
       notifiable.details
```
fosterfarrell9, line 75: According to the above cited gist, possibly `"".html_safe()+notifiable.details` (and remove `text.html_safe`)?

app/helpers/talks_helper.rb
```diff
@@ -1,45 +1,45 @@
...
-                data: { toggle: 'tooltip' },
-                title: t('admin.talk.speakers')).html_safe
+                data: { toggle: "tooltip" },
+                title: t("admin.talk.speakers")).html_safe
   end
 
   def speaker_list_with_icon(talk)
-    (speaker_icon(talk) + speaker_list(talk)).html_safe
+    (speaker_icon(talk) + speaker_list(talk)).html_safe # rubocop:todo Rails/OutputSafety
   end
```
fosterfarrell9, line 36: Possibly add `"".html_safe +` inside the brackets and remove `html_safe` athe end?

fosterfarrell9 (changes requested):
I adressed the points you mentioned, for some I don't know what do myself. I still have to look at all changes in more detail in order to make sure I did not overlook anything. But maybe the current state is okay for a first feedback.

app/models/user.rb
```diff
@@ -34,41 +34,41 @@ class User < ApplicationRecord
   has_many :given_lectures,
            class_name: "Lecture",
            foreign_key: "teacher_id",
-           inverse_of: :user
```
fosterfarrell9: Note that for this change (and all the others in the same commit) I opened a rails console and tested: `user.given lectures` would throw an exception with the `inverse_of: :user` option, whereas everything worked well with `inverse_of: :teacher`.

app/models/medium.rb
Splines: Can confirm partially:

- `restricted?` method for lectures is still needed
- `restricted?` method for course is no longer needed (if you remove this, please also remove the respective test in `course_spec.rb`).
- I have not found a `restricted?` method in `lesson.rb`, it just delegates to the `restricted?` method of the lecture. If you mean that this method is not called on a lesson, you are right; out of the 15 occurrences of `restricted?` in the code, none was a call to the delegated `restricted?` method of lesson. However: is it somehow possible that a lesson is " in disguise" a lecture since a lecture has many lessons? If not, it'd be fine to even remove the `delegate :restricted?, to: :lecture` in `lesson.rb`.

app/models/course.rb
Splines: Oh ok, finally a use case where one writes something different than just the class name. In my head, I've updated the rule to the following mantra, is it correct?

```sh
If "as:" association is specified:
Write inverse_of: :<class that "as:" refers to>
return

If foreign_key is specified, e.g. foreign_key: classname_id
Write inverse_of: :<classname (without the _id part)>
return

If file-level class name ends with Join
Write inverse_of: <convert classname to small snake-case and append an "s" to "join" at the end>
return

Otherwise:
Write inverse_of: :<convert classname to small snake-case>
```

As far as I know, these rules are in accordance with what you've done in c3abe9f82 and 2a730ffe. I've also checked that all `inverse_of` occurrences in the codebase (21 in total) are in line with this ruleset.

I begin to understand why Ruby cannot do this automatically. There might be a lot more edge cases ;)

app/helpers/application_helper.rb
Splines: Oh, glad you caught this! I went through every method of the `application_helper.rb` and found no method that is overwritten in `lecture.rb`. Just a few things that shouldn't be a problem though:

- helper has `active(value)` method, in lecture we use `Term.active`
- helper has `lecture_media` method, we use this as variable name in lecture (but locally scoped, so should be fine), same for `course_media`

The `ApplicationHelper` is also included in `course.rb`, `item.rb`, `medium.rb` and `user.rb`. Shouldn't we also find duplicate naming not just in `lecture.rb` but in these other files as well? If yes, I could do this.

app/models/tag.rb
Splines: Good to know that this is so problematic. Here is what I've done to fix this in commit 461697476d5:

- In the diffs, search for "Rails/SkipsModelValidations": `git log -p -S "Rails/SkipsModelValidations"`
- Check the lines manually, go to the respective file and replace the method call (if the method call included the word `Time` and `updated_at:`)
- In VSCode search: manually check occurrences of "Time.now" and "Time.current" and "Time.zone.now" to make sure they do not occur in an `update` method call (in migrations it's ok). With this approach I changed a few more occurences: `comments_controller.rb`, `medium_controller.rb`, `answer.rb`, in `user.rb` we have `consented_at: Time.zone.now` which I left as is)

app/controllers/registrations_controller.rb
```diff
@@ -1,30 +1,30 @@
...
       request = Net::HTTP::Post.new(uri.request_uri, header)
       request.body = data.to_json
 
       # Send the request
       response = http.request(request)
       answer = JSON.parse(response.body)
       return true if answer["message"] == "verified"
-    rescue
+    rescue StandardError # rubocop:todo Lint/SuppressedException
     end
```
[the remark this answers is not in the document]
Splines: ok, will leave it as `rubocop:todo` then for our future selves to be reminded ;)

spec/cypress/app_commands/log_fail.rb
```diff
@@ -1,31 +1,29 @@
...
 
 if defined?(ActiveRecord::Base)
-  json_result['records'] =
+  json_result["records"] =
     ActiveRecord::Base.descendants
                       .each_with_object({}) do |record_class, records|
       records[record_class.to_s] = record_class.limit(100).map(&:attributes)
-    rescue
+    rescue StandardError # rubocop:todo Lint/SuppressedException
     end
```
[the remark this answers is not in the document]
Splines: ok, will leave it as `rubocop:todo` then for our future selves to be reminded ;)

app/models/course.rb
Splines: Yes, I've just done this in 82c572e39b.

app/helpers/notifications_helper.rb
Splines: I think we shouldn't use `html_safe` here. For example, `lecture_notification_card_text` is constructed from `lecture.course.title`. This is a database field. In my opinion, we should construct the code such that user input is always sanitized *before* it goes to the db. This way, we don't have to worry and sanitize all the time when we query the database. In the case of `lecture_notification_card_text` we even assume that the db field is sanitized as we are using the `_html` method  which is automatically marked as HTML safe by Rails ([see here](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations)).

However, there are lots of different opinions on this [here](https://stackoverflow.com/questions/1274558/when-to-filter-sanitize-data-before-database-insertion-or-before-display). And they have good arguments *not* to escape input and instead just filter it before it gets to the db. For example, when users can store HTML content and wand to edit it, then we shouldn't escape the tags...

What's the way we are doing this right now? Why do we use `html_safe` many times in different views, e.g.
`<%= chapter.details.html_safe %>` in `_chapter.html.erb`. Does that mean we have already sanitized the chapter details when the user submitted the form and we saved it to the db?

Edit: Ok, this [great post](https://stackoverflow.com/a/21886874/9655481) cleared things up to me. Briefly: as long as we use Active Record in the intended way, it's just fine and Rails does the job (as usual ;)). As a [comment over there](https://stackoverflow.com/questions/21886170/best-way-to-go-about-sanitizing-user-input-in-rails#comment136881268_21886874) I've provided some sources for the statements.

app/models/course.rb
fosterfarrell9: The mantra seems correct to me.

app/helpers/application_helper.rb
fosterfarrell9: > Did you mean "method" instead of "class"?

Yes.

app/models/tag.rb
fosterfarrell9: I commented on some lines in [https://github.com/MaMpf-HD/mampf/commit/461697476d51a8b23f05d879b533b048ed6ed7c5](https://github.com/MaMpf-HD/mampf/commit/461697476d51a8b23f05d879b533b048ed6ed7c5), for some reason they does not show up in the main dialogue.

app/models/medium.rb
fosterfarrell9: I removed the `restricted?` methods for course and lesson and the corresponding test in [https://github.com/MaMpf-HD/mampf/pull/566/commits/9fa990a30ccd1cb6348c1ed302060a77d5786a6c](https://github.com/MaMpf-HD/mampf/pull/566/commits/9fa990a30ccd1cb6348c1ed302060a77d5786a6c).

app/models/manuscript.rb
```diff
@@ -410,18 +410,18 @@ def version_info
...
   # sections in a manuscript chapter not represented in mampf
   def new_sections_in_chapter(chapter)
     sections = sections_in_chapter(chapter)
     sections.each_with_index
             .map do |s, i|
-              [s['mampf_section'], i + 1, s['description'], s['counter']]
+              [s["mampf_section"], i + 1, s["description"], s["counter"]]
             end
-            .select { |s| s.first.nil? }
+    sections.select { |s| s.first.nil? }
```
fosterfarrell9, line 424: This seems to be a breaking change: In the lines before this one, `sections` remains unchanged, so you use `sections` of l419 again, not the result of 423.

app/models/tag.rb
Splines: Fixed it in 58874f70b. I also replaced `touch_all` by `touch` in the media controller where you noted there is a chance we actually need to call the callbacks. Well, let's hope it's not the case. If it is, we know how to quickly solve it and can then write a comment over there such that this is no longer implicit knowledge.

app/models/manuscript.rb
Splines: oh, that's the one I fixed manually, and apparently not correctly ;) Here is the fix for the fix: c8addbf555

app/helpers/notifications_helper.rb
Splines: fixed in b26b11f2

app/helpers/notifications_helper.rb
Splines: see above (fixed in b26b11f2)

app/helpers/talks_helper.rb
Splines: not changed, see above)

app/models/talk.rb
```diff
@@ -111,6 +112,6 @@ def talk_path
     end
 
     def remove_duplicate_dates
-      update_columns(dates: dates.uniq)
+      update_columns(dates: dates.uniq) # rubocop:todo Rails/SkipsModelValidations
     end
```
Splines, line 116: How do we want to address this? Does `update_columns` refer to the currently dealt-with Talk or to all Talks in the database since it's not prefixes with something like `my_talk.`

fosterfarrell9: It refers to the currently dealt-with talk (it is equivalent to` self.update_columns(dates: date.uniq)` for which Rubocop would tell you to omit  the `self`). Note that it is not possible to replace `update_column` by `update` here: this would trigger an infinite loop since `update` would again call the `after_save` action which would then call `update` and so on.

However, what we could do here is the following: Replace the line

```ruby
 after_save :remove_duplicate_dates
```

by 

```ruby
before_save :remove_duplicate_dates
```

and replace the `remove_duplicate_dates` method by 

```ruby
def remove_duplicate_dates
  dates.uniq!
end
```

You should try it out e.g. in `/talks/1/edit`.

Splines: Fixed in eb92c86.

Thanks, I've tried it out and it works fine. Note that a more elegant solution would probably be to have a **set instead of an array** as underlying data structure in order to avoid duplicates. I've added a todo note as comment; this will presumably involve a migration to change the data type on the existing data.

fosterfarrell9: I am not sure that set is supported by the adapter for ActiveRecord, see [here](https://guides.rubyonrails.org/active_record_postgresql.html) for the ist of supported data types.

app/helpers/application_helper.rb
fosterfarrell9: I do not really think that is really necessary (in contrast to the lectures_helper there has not been a related change here), but feel free to do so if you think it is better.

fosterfarrell9 (approved):

app/models/talk.rb
Splines: Ah, it's a pity PostgreSQL doesn't support sets natively. Then, we'll have to live with it ;)

app/helpers/application_helper.rb
Splines: You're right, I've briefly checked the mentioned files and there seems to be no conflict (I searched for "announce" and these files do not even include that word).

PR #578, opened 2024-01-04
Upgrade Node.js to v20 (LTS)

Splines:
Upgrade Node.js to v20.10.0 (LTS until April 2026).

~~I tested this on the `experimental` branch and the run was successful (see Dozzle). Also check out [mampf-experimental](https://mampf-experimental.mathi.uni-heidelberg.de/).~~ Nope, that's what I hoped, but of course the old error keeps hunting us.

And we are not alone, the question for this error was viewed 2.5million times on Stack overflow [here](https://stackoverflow.com/q/69692842/9655481)! I will fix this! [maybe useful](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported#comment131441027_73465262)

### Finally fixed
In 8f67ce9, I finally fixed this issue after several failed attempts. See the comment in the code for further details. The `experimental` branch currently also points to this commit and you can see in the logs that it's working. [mampf-experimental](https://mampf-experimental.mathi.uni-heidelberg.de/) also shows the monotiles on the landing page.

### DIFF
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -2,8 +2,47 @@ const { environment } = require("@rails/webpacker");
 const coffee = require("./loaders/coffee");
 const css = require("./loaders/css");
 
-environment.loaders.prepend("coffee", coffee);
+// Fix OpenSSL error starting with Node.js 17.
+//
+// This error occurred since a new version of OpenSSL shipping with Node.js 17+
+// removed support for MD4. This is actually very good as MD4 is insecure.
+// However, webpack 5 is using MD4 as default hashing algorithm, that's why we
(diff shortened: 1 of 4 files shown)


fosterfarrell9 (approved):
Works like a charm. Is `/docker/run_cypress_tests/Dockerfile` excluded from the change because it is currently excluded from the pipeline anyway?

Splines:
Oh no, thanks I simply forgot about it 🙈 It's not used right now, that's true. I've still updated it to keep the symmetry. I need another approval then.

fosterfarrell9 (approved):

PR #579, opened 2024-01-06
[rebase] Continuous Release 1.9.0

Splines:
See [this](https://stackoverflow.com/q/16306012/9655481) and [this](https://stackoverflow.com/q/39048673/9655481) SO posts. Currently I don't have any plans why this is the case :(

[Rebasing](https://stackoverflow.com/a/45286781/9655481) works (see commits below) but assigns new commit ids and updates the timestamps.


Splines:
Closed in favor of #561

PR #584, opened 2024-01-08
Fix missing `to_a` to convert to Rails array

### LINKED ISSUE #583: Undefined method 'intersect?'
Since rolling out #579, we are seeing the follwing error in production:

```ruby
undefined method `intersect?' for #<ActiveRecord::Associations::CollectionProxy [#<Tag id: 636, created_at: "2019-01-05 17:21:40.169479000 +0100", updated_at: "2022-02-25 14:17:38.015382000 +0100", realizations: []>, #<Tag id: 656, created_at: "2019-01-28 18:53:41.680154000 +0100", updated_at: "2021-11-18 09:26:39.281654000 +0100", realizations: []>]> Did you mean? insert 
```

The reason is that in `media_controller.rb`, we have in line 639

```ruby
m.teachable_type == "Course" && !m.tags.intersect?(lecture_tags)
```

which replaced

```ruby
m.teachable_type == 'Course' && (m.tags & lecture_tags).blank?
```

but `m.tags` is an ActiveRecord collection, not an an array, so there is no `intersect?` method here (where as in the code before,
an automatic conversion seems to have happened).

I think an obvious solution would be to write `m.tags.to_a.intersect?`.

Splines:
Fixes #583.

This was due to the unsafe RuboCop rule in #566: [`Style/ArrayIntersect`](https://docs.rubocop.org/rubocop/cops_style.html#stylearrayintersect).

> Safety: This cop cannot guarantee that array1 and array2 are actually arrays while method intersect? is for arrays only.

That was the case here.

- [x] Checked locally that repro given in #583 is not applicable anymore in localhost.
- [x] I also searched for other occurrences of `intersect?` but this was the only one in our codebase.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -636,7 +636,7 @@ class MediaController < ApplicationController
         unless current_user.admin || @lecture.edited_by?(current_user)
           lecture_tags = @lecture.tags_including_media_tags
           search_results.reject! do |m|
-            m.teachable_type == "Course" && !m.tags.intersect?(lecture_tags)
+            m.teachable_type == "Course" && !m.tags.to_a.intersect?(lecture_tags)
           end
         end
       end


Splines:
I'd say we deploy this to `main` directly as a hotfix, ok?

fosterfarrell9 (approved):
Hotfix is fine with me.

PR #576, opened 2024-01-04
Fix unread comments status (new trial)

Splines:
~~Please comment at #574 and not here.~~

PR #585, opened 2024-01-08
Fix comment status

### LINKED ISSUE #573: Comments icon colored with no new comments
It appears that #515 contains a bug: In production I just posted an answer to a user comment on one of my media, and now I have a yellow comment icon in the navbar indicating that there are new comments, while clicking on the icon shows there are none:

new_comments

There is no way to decolor the yellow icon (only waiting until someone posts a new commmment somewhere and the clearing all commments). This is actually something we wanted to avoid in #515 but for some reason, it does not work.

Splines:
Fixes #573.

<sub>Previous PR for this was #576, closed in favor of this one as this directly branches off the new "dev" branch.</sub>

### Other things to do
- [x] Delete old branch of #574
- [x] Create migration to fix inconsistencies in db introduced with #515. ⚠ Should be done in conjunction with this PR. -> see #587 
- [x] Model tests ran through locally ✅

### DIFF
--- a/app/controllers/readers_controller.rb
+++ b/app/controllers/readers_controller.rb
@@ -9,7 +9,7 @@ class ReadersController < ApplicationController
     @reader = Reader.find_or_create_by(user: current_user,
                                        thread: @thread)
     @reader.touch
-    @anything_left = current_user.media_latest_comments.any? do |m|
+    @anything_left = current_user.subscribed_media_latest_comments_not_by_user.any? do |m|
       (Reader.find_by(user: current_user, thread: m[:thread])
             &.updated_at || 1000.years.ago) < m[:latest_comment].created_at
     end
--- a/app/controllers/readers_controller.rb
+++ b/app/controllers/readers_controller.rb
@@ -9,7 +9,7 @@ class ReadersController < ApplicationController
     @reader = Reader.find_or_create_by(user: current_user,
                                        thread: @thread)
     @reader.touch
-    @anything_left = current_user.subscribed_media_with_latest_comments_not_by_user.any? do |m|
+    @anything_left = current_user.subscribed_media_with_latest_comments_not_by_creator.any? do |m|
       (Reader.find_by(user: current_user, thread: m[:thread])
             &.updated_at || 1000.years.ago) < m[:latest_comment].created_at
     end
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -549,14 +549,36 @@ def subscribed_commentable_media_with_comments
...
+  # For each medium, the following information is stored:
+  # - the medium itself
+  # - the thread of the medium
+  # - the latest comment by somebody else than the current user
+  # - the latest comment by any user (which might include the current user)
+  def subscribed_media_with_latest_comments_not_by_creator
+    media = subscribed_commentable_media_with_comments.map do |m|
+      thread = m.commontator_thread
+      comments = thread.comments
+      next unless comments
```
fosterfarrell9, line 568: `comments` is an array, and empty arrays are truthy in Ruby. You might want to replace this by 

```ruby
next if comments.blank?
```

app/models/user.rb
```diff
@@ -549,14 +549,36 @@ def subscribed_commentable_media_with_comments
...
+
+      comments_not_by_creator = comments.reject { |c| c.creator == self }
+      next unless comments_not_by_creator
```
fosterfarrell9, line 571: Same here. You might want to replace this by 

```ruby
next if comments_not_by_creator.blank?
```

app/controllers/readers_controller.rb
```diff
@@ -9,7 +9,7 @@ def update
     @reader = Reader.find_or_create_by(user: current_user,
                                        thread: @thread)
     @reader.touch
-    @anything_left = current_user.media_latest_comments.any? do |m|
+    @anything_left = current_user.subscribed_media_with_latest_comments_not_by_creator.any? do |m|
       (Reader.find_by(user: current_user, thread: m[:thread])
             &.updated_at || 1000.years.ago) < m[:latest_comment].created_at
     end
```
fosterfarrell9, line 15: In some later PR, the code duplication in `main_controller.rb` and `readers_controller.rb` should be removed, e.g. by writing a method for the last two lines.

fosterfarrell9 (changes requested):

fosterfarrell9 (approved):

Splines:
thanks for the approval. I will merge once the respective migration is ready

PR #587, opened 2024-01-11
Migrate `unread_comments` flag (fix inconsistencies)

Splines:
Updates the `unread_comments` for all users such that it is consistent with the `Reader` model again. This is due to #515, which introduced unwanted behavior. The behavior itself is fixed in #585.

### Useful for reviewers
```
# Revert last migration to reapply it
# (e.g. by accessing the frontend localhost:3000
# and clicking on "Run pending migrations")

docker exec -it $(docker ps -qf "name=development-mampf") bash
bundle exec rake db:rollback:primary STEP=1

# Note you will probably have to also undo changes in the schema.rb file
# after each rollback to be able to reapply it.
```

Useful SQL query for [pgadmin](https://github.com/MaMpf-HD/mampf/wiki/Database#pgadmin):
```sql
SELECT id, name, unread_comments FROM public.users ORDER BY id
```

before

In your mampf docker container, search for "Ran through" after the migration. Please try out many use cases, e.g. when the flag shouldn't be changed for any user (log should show "Fixed 0 users"), when the flag for two users should be updated etc. Please test this migration thoroughly to avoid nasty things in the production db.


db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
```diff
@@ -0,0 +1,55 @@
...
+# different unread comments flags as the ones at the time of the migration.
+#
+# This migration is not reversible as we don't store the previous state of
+# the unread_comments flag.
+class FixUnreadCommentsInconsistencies < ActiveRecord::Migration[7.0]
+  def change
+    num_total_users = 0
+    num_fixed_users = 0
+
+    User.find_each do |user|
```
fosterfarrell9, line 16: I think you can just write
```ruby
User.find_each.with_index do |user, num_total_users| 
```
and then you can omit the `num_total_users = 0` and `num_total_users +=1` (did not check it though, just from memory). Detailed report to follow tomorrow or the day after.

Splines: I don't know why I counted manually in the first place as I could just use `array.length` to get the total number ;) Fixed this.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
```diff
@@ -10,16 +10,15 @@
 # the unread_comments flag.
 class FixUnreadCommentsInconsistencies < ActiveRecord::Migration[7.0]
   def change
-    num_total_users = 0
+    users = User.all
     num_fixed_users = 0
 
-    User.find_each do |user|
+    users.find_each do |user|
       was_user_fixed = fix_unread_comments_flag(user)
```
fosterfarrell9, line 17: I think that using `User.all` before `users.find_each` makes the whole point of using `find_each` (working in batches) moot since `User.all` loads all records into memory, see e.g. [here](https://www.mintbit.com/blog/why-to-use-find-each-instead-of-all-in-ruby-on-rails). If you do not want to use an index variable in the loop, you can use `User.count` to get the number of users with one SQL query.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
```diff
@@ -0,0 +1,54 @@
+# Fixes the unread_comments flag for all users. Unintended behavior was
+# introduced in pull request #515. Migration introduced in #587.
+# Behavior fixed in #585.
+#
+# This migration is generally *not* idempotent since users might have interacted
+# with the website since the migration was run and thus they will probably have
+# different unread comments flags as the ones at the time of the migration.
+#
```
fosterfarrell9, line 8: When I did the migration in my development setting that had no faulty unread comments flags, I obtained what I expected:

```
Ran through 11 users (unread comments flag)
Fixed 0 users (unread comments flag)
```

Directly after that (without any user interactions in the meantime) I rolled back and obtained:

```
Ran through 11 users (unread comments flag)
Fixed 2 users (unread comments flag)
```

That makes me a little suspicious in several ways:

- Since we only have a `change` method, what does Rails do when rolling back? I would assume it executes the `change` method again, since there is no way for it to deduce how to revert the migration (in contrast to the usual `add_column` migrations). The SQL statements in the log seem to confirm this. Would it not be better to put the content of the current `change` method into the `up` method and define an empty `down` method?
- If in the above scenario the change method in the migration did not touch the database and no user interactions happened, how can an application of the same method suddenly result in 2 changed user records? Also, I see nothing in the SQL logs in the down migration that indicates any actual updates. Maybe there is something wrong with the return value in `fix_unread_comments_flag`?

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    Rails.logger.debug { "Ran through #{users.length} users (unread comments flag)" }
+    Rails.logger.debug { "Fixed #{num_fixed_users} users (unread comments flag)" }
+  end
+
+  # Fixes the unread_comments flag for a given user.
+  # Returns true if the flag needed a change (and was changed), false otherwise.
+  def fix_unread_comments_flag(user)
+    readers = Reader.where(user_id: user.id)
```
fosterfarrell9, line 28: You can write 
```ruby
readers = Reader.where(user: user)
```
here.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
```diff
@@ -0,0 +1,54 @@
...
+    return false if readers.blank?
+
+    had_user_unread_comments = user.unread_comments
+    has_user_unread_comments = false
+
+    readers.each do |reader|
+      thread = Commontator::Thread.find_by(id: reader.thread_id)
+      next if thread.blank? # thread_id should never be nil, just to be sure
```
fosterfarrell9, line 36: If `thread_id` is `nil` or some other value that is not attained and no thread is found, then` thread` will bei `nil`. So, the statement 

```ruby
next unless thread
```
should actually work for all possible cases, since nil is falsy and an actual thread is truthy. If you want to make it more verbatim, I would probably rather use 

```ruby
next if thread.nil?
```
instead of `thread.blank?` but that is maybe a personal preference of mine. See also [here](https://stackoverflow.com/questions/885414/how-to-understand-nil-vs-empty-vs-blank-in-ruby).
If you want to leave it as it is this is also fine with me.

fosterfarrell9 (changes requested):

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
Splines: Mmm, I cannot reproduce this scenario in my local setup. Have you checked that entries in the `unread_comments` column actually change when it says "Fixed 2 users"? What command did you use to roll back?

With regards to the `change` method. You have a point here as I already marked the migration as *not* reversible in the comment. We should put the code in the `up` method and [throw an error to prevent reverts](https://guides.rubyonrails.org/active_record_migrations.html#throwing-an-error-to-prevent-reverts).

Splines: I throw an error in 144141245e now.

Splines: Oh, I think I found the flaw in my implementation and will fix it soon.

Splines: Thanks for noting, there were some serious flaws in my implementation that I addressed in 97f5ac3. Foremost, I only checked the `Reader` for unread comments but forgot about the case where there is no Reader object, but where there is actually unread media present.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
Splines: Fixed in b66348df0.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
Splines: Oh yeah, true. I changed it to `User.count`.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
Splines: Changed to `next if thread.nil?` in 3eb6157ecd2.

db/migrate/20240109180000_fix_unread_comments_inconsistencies.rb
fosterfarrell9: Raising an `ActiveRecord::IrreversibleMigration` in the down migration is fine with me.

db/migrate/20240116180000_fix_unread_comments_inconsistencies.rb
```diff
@@ -0,0 +1,59 @@
...
+      has_user_unread_comments = reader.updated_at < latest_thread_comment_time
+
+      return true if has_user_unread_comments
+    end
+
+    # User might still have unread comments but no related Reader objects
+    # -> Check for unread comments -- via Media
+    # TODO: rename media_latest_comments to subscribed_media_with_latest_comments_not_by_creator
+    # see change in PR #585
+    unseen_media = user.media_latest_comments.select! do |m|
```
fosterfarrell9, line 50: Isn't #587 going to be merged into dev after #585? Because #585 replaces the `media_latest_comments`
method of `User` by a method called `subscribed_media_with_latest_comments_not_by_creator`. The migration would then throw an error because this method no longer exists:

```
Migrating to FixUnreadCommentsInconsistencies (20240116180000)
== 20240116180000 FixUnreadCommentsInconsistencies: migrating =================
  TRANSACTION (0.2ms)  BEGIN
  User Load (0.5ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1000]]
  Reader Load (0.3ms)  SELECT "readers".* FROM "readers" WHERE "readers"."user_id" = $1  [["user_id", 1]]
  TRANSACTION (0.2ms)  ROLLBACK
   (0.3ms)  SELECT pg_advisory_unlock(8267686635516738675)
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined method `media_latest_comments' for #<User id: 1, email: "admin@mampf.edu", created_at: "2020-07-24 15:07:48.807973000 +0200", updat...
```

Before I noticed that, I had checked the branch out from `dev`, and then the migration ran through, but created  new inconsistencies:
Screenshot 2024-01-16 165217

My hope was then that if I merged `fix/unread-comments-db-inconsistencies` into `fix/unread-comments` and replaced `media_latest_comments` in the migration by `subscribed_media_with_latest_comments_not_by_creator` then the newly created inconsistencies would go away after migrating, but they did not.

fosterfarrell9 (changes requested):

db/migrate/20240116180000_fix_unread_comments_inconsistencies.rb
Splines: > Isn't https://github.com/MaMpf-HD/mampf/pull/587 going to be merged into dev after https://github.com/MaMpf-HD/mampf/pull/585?

Yes, that's true and which is why I wrote the respective "TODO" in the code as this has to be done after #585 is in `dev` (I'm going to merge #585 soon, but wanted to wait until this PR here is ready to go except for this method renaming).

---

> My hope was ...

The way you merged branches, it should have actually got rid of any inconsistencies. I'm keen to see the exact case in your local setup in the meeting.

Splines: > The problem seems to be that you use select! instead of select

Thanks a lot, that was the problem. This made it to the hall of fame of dumb bugs I introduced, see [here](https://github.com/MaMpf-HD/mampf/wiki/Code#hall-of-fame-of-dumb-ruby-bugs-%EF%B8%8F).

Splines:
@fosterfarrell9 After your approval, I will wait with merging until @chrisflav has updated the script on the server such that the containers rebuild from `dev` instead of `mampf-next`. This is especially important here as I want to see the migration running through correctly in the dev stage.

chrisflav:
The script and the webhook configuration is updated: `mampf-next` is replaced by `mampf-dev`.

Splines:
Hey @chrisflav thanks, however the respective branch is called `dev`, not `mampf-dev`.

The respective url can still be "https://mampf-dev.mathi.uni-heidelberg.de/" though, just like we have the branch `experimental` and the respective url "https://mampf-experimental.mathi.uni-heidelberg.de/".

chrisflav:
Okay, changed from `mampf-dev` to `dev`.

Splines:
@chrisflav Thanks a lot. However, I've just merged this PR and it did not trigger the redeploy of dev. Any ideas?

chrisflav:
Did you merge here in the github interface? The triggered webhooks all have this branch as head reference, that's why `dev` was not matched.

chrisflav:
The webhooks are only triggered on `push` and I don't know what github counts into that.

Splines:
I did indeed merge via the GitHub interface by pressing on the green "Squash and merge" button. Note that this is the exact same workflow we had before: we squash-merged onto `mampf-next` previously and this triggered the respective webhook. It's just that the branch has a different name now.

Maybe there's something missing in the webhook settings for the `dev` branch that was previously set for the `mampf-next` branch?

chrisflav:
Its redeploying now, I forgot to reload the webhook configuration and was tricked by Githubs webhook interface: I did not notice that the webhook (actually: all three) was triggered twice. Once with this branch as head ref and once with the `dev` branch, I only saw the former and concluded there was something wrong on the Github side.

Splines:
@chrisflav Working now, thanks 🙌

Issue #542, opened 2023-10-15
Env variables default to `nil`


Splines:
In [this commit](https://github.com/MaMpf-HD/mampf/pull/527/commits/696a395cce1bb24d1018e84479a85cc0a66d77ba) for the Feedback button (#527), I've replaced lines like

```rb
ENV.fetch('ERDBEERE_SERVER', nil)
```

by
```rb
ENV.fetch('ERDBEERE_SERVER')
```

in the `default_setting.rb`, **which is a good suggestion by rubocop since now we will get a key error if we forget to set an env variable**. This is good as we don't want to silently get a `nil` value assigned if we somehow forgot to specify the env variable.

---

And we directly got a key error in the redeploy (on experimental): The reason is that the line
```rb
SECRET_KEY_BASE="$(bundle exec rails secret)" DB_ADAPTER=nulldb bundle exec rails assets:precompile
```
in the Dockerfile now fails since no env variables are set at all (it just shows the first missing env variable in the logs).

## What we should do
We should find a workaround, so that `rails assets:precompile` already gets the env variables populated. Or does it even need to have any env variables set? I'd expect it to only run webpack and bundle static assets (js and css). In some html files, we use something like `DefaultSetting::MUESLI_LINK`, however this is not a static assets and instead rendered directly, right? This is probably, why it was fine to have `nil` values beforehand in this case for the env variables. But we should avoid this to distinguish the case when we really forget to set an env variable (which would defualt to `nil` too in this case).

TL;DR: find a way so that `rails assets:precompile` gets env variables from `docker.env` passed.

Splines:
Maybe helpful:

- [How to skip initializers while running rails assets:precompile?](https://www.reddit.com/r/rails/comments/tjyjlw/how_to_skip_initializers_while_running_rails/)
- [Assets precompile without a database in Rails 4 and 5](https://www.carlos-roque.com/2020/12/09/assets-precompile-without-a-database-in-rails-4-and-5/)

Splines:
[This comment](https://github.com/docker/compose/issues/5600#issuecomment-727905877) could be a valid fix for this issue. Will try it out.

PR #590, opened 2024-02-12
Authenticate mails sent by MaMpf

Splines:
Mails have been sent unauthenticated internally via the Mathi mail server. This PR introduces user/password authentication.

~~The change I had to introduce in f40080e is unfortunate but necessary due to #542. Hopefully, we can fix this in the near future.~~ Fixed in 4d066c3 (merge commit), now we don't have `, nil` anymore. See also issue #542.

**TODO**
- [x] Respective env variables have to be added to the server env files.


Splines:
Converted back to draft as I'd like to have #542 fixed first.

Splines:
Will merge without review as Denis is on vacation.

PR #593, opened 2024-02-15
Fix erroneous media search

fosterfarrell9:
This PR fixes some strange behaviour of the media search. The user input of tags was in many cases not actually added to the SOLR search request. Also, there was some strange second search taking place if the tag operator was 'OR' and some text was given to search. The new behavior is way more consistent: The different search options (type, associated to, tags,...) are now
combined by 'AND' and the tag operator behaves according to the chosen operator OR/AND.


Splines:
In localhost, what I found before your change:

image


After the change I don't find any results anymore in this setting which is probably what we want as the results don't have anything to do with the Weierstrass normal form as far as I know ;)

app/models/medium.rb
```diff
@@ -336,15 +336,12 @@ def self.search_by(search_params, _page)
         with(:release_state, search_params[:access])
       end
     end
-    if !search_params[:all_tags] == "1" &&
-       !search_params[:tag_operator] == "or" && (search_params[:tag_ids])
-      if search_params[:tag_operator] == "or" || search_params[:all_tags] == "1"
```
Splines, line 341: We had `search_params[:all_tags] == "1"` beforehand to force when we should use a logical OR. I think this makes sense: when a user wants to include "all tags" in their search, why would they want an AND to chain them since its very improbable that every search result will have all tags included. So, I think only OR makes sense here and the user should not even be able to select "AND" in this case. Of course, this was poorly or not at all communicated to the user before this change: even when the user selected "AND" it was still an "OR" in the code here. You can test this at `git checkout b3b7f6ac96`.

If you agree, I could go ahead and disable the OR/AND once the user clicked on "all tags". But this should still be reflected in the backend then, that is, if all tags are selected always use a logical "OR".

Splines (changes requested):

app/controllers/media_controller.rb
Splines, line 1: In localhost, what I found before your change:

image


After the change I don't find any results anymore in this setting which is probably what we want as the results don't have anything to do with the Weierstrass normal form as far as I know ;)

Chaining tags with `OR` als works:
image

with `AND` I don't get any results in this case, which is expected.

app/controllers/media_controller.rb
```diff
@@ -228,21 +228,6 @@ def search
...
-      && (search_params[:all_tags] == "0") \
-      && (search_params[:fulltext].size >= 2)
-      params["search"]["all_tags"] = "1"
-      search_no_tags = Medium.search_by(search_params, params[:page])
-      search_no_tags.execute
-      results_no_tags = search_no_tags.results
-      results = (results + results_no_tags).uniq
-      @total = results.size
-      params["search"]["all_tags"] = "0"
-    end
```
Splines, line 244: This was introduced in https://github.com/MaMpf-HD/mampf/commit/a67a2cf98b15e1b7b2acdb00ec0a182325e96c0b. Fabian wrote:

> updated media search by tags.
using the "or" operator now executes two searches,
one where the tags are included and one where
they are ignored. they are then combined so
that those with the selected tags are shown first
in the results.
also fixed the search with the "and" operator.

So that's the reason behind this.

Splines: But I agree that this can be very misleading if the search *with* the tags did not yield any results, but the search *without* tags did. Then, the user only sees the results that were gathered *without* the specified tags and is really confused. That's why I agree that this part here can be removed.

app/models/medium.rb
fosterfarrell9: Yes, I agree. The OR/AND should be removed in the fronend if all tags are selected. In the frontend I think it should be sufficient to remove lines 304 to 306 in `medium.rb`. Then only the code inside

```ruby
if search_params[:all_tags] == "0" && search_params[:tag_ids].any?
  search.build do
    if search_params[:tag_operator] == "and"
      with(:tag_ids).all_of(search_params[:tag_ids])
    else
      with(:tag_ids).any_of(search_params[:tag_ids])
    end
  end
end
```

sets any restriction on the SOLR search regarding tags. If `search_params[:all_tags] == "1"` then no such restrictions meaning that the result does not take into account which tags are involved.
If we agree on this please go ahead and make the frontend (and minor backend) change. If not, we can further discuss.

app/controllers/media_controller.rb
fosterfarrell9: We could leave it in, but I really see no use case for that now (one and  half years later).

Splines: Yeah, as this behavior might only lead to confusion I think we should get rid of it as you did 👍

app/models/medium.rb
Splines: You mean these lines?

https://github.com/MaMpf-HD/mampf/blob/469a5836c67ca01873805daf0e6fd6e870539958/app/models/medium.rb#L304-L306

What is `search_params[:tag_ids] = Tag.pluck(:id)` supposed to do in the first place?

Splines: Unfortunately, the search form has been copied in the code in some occurences, see for example the following string:

```rb
<%= f.label :tag_ids,
                  t('basics.related_tags'),
                  class: 'form-label' %>
```

in `app/views/main/start` as well as `app/views/media/catalog` and the respective complete formula that is copied.

What I struggle with is to find out the purpose of the `_search_results.html.erb` file, especially since it renders the `tags_form` where I would have to also reflect this change:

https://github.com/MaMpf-HD/mampf/blob/469a5836c67ca01873805daf0e6fd6e870539958/app/views/media/catalog/_search_results.html.erb#L105

This form is only shown if the purpose is either "media" or "clicker" as seen here:

https://github.com/MaMpf-HD/mampf/blob/469a5836c67ca01873805daf0e6fd6e870539958/app/views/media/catalog/_search_results.html.erb#L95

Search for the string `search_results"` and the only occurrence I get is in the `import_preview.coffee` file where the purpose is set. Do you know where in the UI I can find the place where `purpose` will be set to "media" such that I can see the `tags_form.html.erb` rendered?

fosterfarrell9: > What is search_params[:tag_ids] = Tag.pluck(:id) supposed to do in the first place? Apart from that I agree: if tag_ids is empty, we won't construct any restrictions, i.e. "all tags" are included.

It replaces the the tag_ids array (which is empty in this case) by the array of all ids of tags that exists. As you already wrote, it is highly unlikely that a medium exists whose associated tags are all existing tags. So, using this `tag_ids` array in combination with  `with(tag_ids).all_of(search_params[:tag_ids])` will result in no search results and using it with `with(tag_ids).any_of(search_params[:tag_ids])` results in media that have at least one (arbitrary) tag. I do not see why wee would need any of this.

fosterfarrell9: > Search for the string search_results" and the only occurrence I get is in the import_preview.coffee file where the purpose is set. Do you know where in the UI I can find the place where purpose will be set to "media" such that I can see the tags_form.html.erb rendered?

You can see `import_preview.coffee` in action in two places: For displaying search results for importing media in a lecture in a lecture edit page, see e.g. `http://localhost:3000/lectures/1/edit` in the accordion in the development setting, or for quizzes after you import questions for a quiz: `http://localhost:3000/quizzes/16/edit` (click on 'new vertex', then import question). In these cases the purpose is `quiz`/`import` so the `search_results` partial will not render the `tags_form` partial.

Splines: > It replaces the the tag_ids array (which is empty in this case) by the array of all ids of tags that exists.

Thanks for the clarification, I know get what the purpose of this line was.

>  I do not see why wee would need any of this.

Fair enough, so I removed the logic in 7dd388313f

> the `search_results` partial could be hugely simplified.

Ok yeah, we should really fix this code duplication in some future PR to remove this dreadful sin from the code ;) Especially since these search forms are a perfect fit for reusability across multiple places.
At least, I did not have to mirror the UI change as the AND/OR fields both have the same ID in the two files they occur in, so one JS file was sufficient to handle both cases. The only real thing I had to mirror is to add the `disabled: true` statement on the respective AND/OR tag operators to disable them initially since "all" is selected initially.

app/controllers/media_controller.rb
Splines: This was just intended as an observation that it is working as expected, so I resolve this conversation.

app/models/medium.rb
Splines: From my side, this can be resolved.

Splines (approved):
<sub>Note I temporarily switched the base of this branch and then went ahead marking this PR as draft to remark it as ready for review to trigger our linting checks again. Don't know why the tests weren't executed, as they should be on every commit. Maybe a small hickup on GitHub's side who knows. Now the tests ran through ;)</sub>

Please use "Squash and merge"

PR #600, opened 2024-03-20
Fix discard in quiz editor

### LINKED ISSUE #598: Discard is same as save in quiz editor
Discarding an item in the quiz editor just saves the item and does not discard it.

Splines:
Fixes #598


Splines:

PR #601, opened 2024-03-20
Continuous Release 1.10.0

Splines:
Feedback button ⭐ and other things ;)


Splines:
Will merge without approval as Denis is on vacation.

PR #602, opened 2024-03-22
Redesign comments page

Splines:
**Before:**
before

**After:**
after

**Even more comments:**
even more comments

**Alerts work as usual:**
comments page with alert


Splines:

PR #609, opened 2024-04-06
Upgrade Rails to v7.1 & update gems (mostly only minor versions)

### LINKED ISSUE #554: Migrate to Rails 7.1
Rails 7.1 has been released, so we should migrate at some point. See [here](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-0-to-rails-7-1) for the migration guide.

Splines:
This fixes #554. See the [upgrade guide](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html) and [all config keys](https://guides.rubyonrails.org/v7.1/configuring.html).

**Commands** used to perform the update:
```
# Edit Gemfile manually and update Rails version there. Then run
bundle update

# And then (development env variable was only set later in the process)
RAILS_ENV="development"  bin/rails app:update --trace

# I've also run `bin/rails app:update --trace` inside a local dev mampf container to finish the process
# (active record needed to connect to the database)
```

### Notable changes
Only changes I found noteworthy in particular.

- [x] ⚠ [`Memcache is using connection pooling by default`](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#memcachestore-and-rediscachestore-now-use-connection-pooling-by-default). Let's see if that will cause any troubles in production.
- [x] [`raise_on_missing_translation`](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#config-i18n-raise-on-missing-translations-true-now-raises-on-any-missing-translation) is great, I guess. So i've activated it.
- [x] Set [default log level](https://guides.rubyonrails.org/v7.1/configuring.html#config-log-level) in production to `info` instead of `warn`. This will log a bit more information that may be helpful in production. `info` is also the default by Rails.
- [x] Set [`config.require_master_key = true`](https://guides.rubyonrails.org/v7.1/configuring.html#config-require-master-key). With this config option activated, If the Rails master key is not available in production, the app won't boot, which is good.
- [x] ❓ I'm unsure whether this hash change affects us: [`allow_deprecated_parameters_hash_equality`](https://guides.rubyonrails.org/v7.1/configuring.html#config-action-controller-allow-deprecated-parameters-hash-equality)
- [x] Transactions callbacks are handled in a better way to avoid stale data. See [here](https://guides.rubyonrails.org/v7.1/configuring.html#config-active-record-run-commit-callbacks-on-first-saved-instances-in-transaction).
- [x] ❓❗ Active Record Encryption now uses [SHA-256 as its hash digest algorithm](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-encryption-algorithm-changes).
--> Not affected as we don't use Active Record Encryption.
- [x] Some migrations were introduced by the Rails updater. I don't really know what they do, will have to review them.
--> Deleted them as we don't use ActiveStorage, but `shrine` instead.
- [x] `globalize_attribute_names`, see [this](https://github.com/MaMpf-HD/mampf/pull/609#issuecomment-2074356660) comment

### For the future
- [ ] [`Content-Security-Policy`](https://guides.rubyonrails.org/security.html#content-security-policy-header) might be useful to set up but we have to watch out if this is working in conjunction with nginx.

### TODO
- [x] Go through notable changes list and update `new_framework_defaults_7_1.rb` accordingly
- [x] Check if everything is working locally
- [ ] Check if everything is working in mampf-experimental
- [x] ~~The Rails updater introduced some migrations and I'm not sure yet what their purpose is. Will have to review them. (see above)~~


fosterfarrell9:
> Some migrations were introduced by the Rails updater. I don't really know what they do, will have to review them.

I think we can delete them since we do not use ActiveStorage (but the shrine gem instead for uploads). They would not do anything anyway since they all contain the line

`return unless table_exists?(:active_storage_blobs)`

and this table does not exist in our setting.

> Active Record Encryption now uses [SHA-256 as its hash digest algorithm](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-encryption-algorithm-changes).

At least as of now, this does not affect us since we do not use [ActiveRecord Encryptio](https://guides.rubyonrails.org/active_record_encryption.html)n (which was introduced in Rails 7.0).

> I'm unsure whether this hash change affects us: [allow_deprecated_parameters_hash_equality](https://guides.rubyonrails.org/v7.1/configuring.html#config-action-controller-allow-deprecated-parameters-hash-equality)

Me as well. This seems to refer to this:

```
Comparing equality between `ActionController::Parameters` and a
`Hash` is deprecated and will be removed in Rails 7.2. Please only do
comparisons between instances of `ActionController::Parameters`. If
you need to compare to a hash, first convert it using
`ActionController::Parameters#new`.
To disable the deprecated behaviour set
`Rails.application.config.action_controller.allow_deprecated_parameters_hash_equality = false`
```
I would expect we do not use the deparacted comparisons but to be sure one would have to go through all the controllers.

Splines:
> I think we can delete them since we do not use ActiveStorage (but the shrine gem instead for uploads).

Thanks, I just deleted these migrations.


> At least as of now, this does not affect us since we do not use [ActiveRecord Encryptio](https://guides.rubyonrails.org/active_record_encryption.html)n (which was introduced in Rails 7.0).

Ah, great, one less thing to worry about.

Splines:
> This seems to refer to this:

Where do you have that info from?

fosterfarrell9:
In the Rails source code: https://github.com/rails/rails/blob/v7.1.3.2/actionpack/lib/action_controller/metal/strong_parameters.rb

Splines:
> So I think we just have to look if for == with params involved in our controllers.

I've checked that for our controllers by searching for "==". I could only find comparison to string or integers but not to any hashes. Maybe you can take a look as well?

In the `comments_controller` we have `@comment.thread.config.comment_voting.to_sym == :ld`, but that's not an issue, I guess, since the first argument is not an action controller parameter.

**In the future, before upgrading to Rails 7.2, we should check our production log and see if the warning you mentioned pops up there.** If it does, we know where to fix this in the code.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -0,0 +1,87 @@
...
+# Rails.application.config.action_controller.allow_deprecated_parameters_hash_equality = false
+
+# ❓ Invalid cache expiration time, what behavior do we want?
+###
+# Specify if an `ArgumentError` should be raised if `Rails.cache` `fetch` or
+# `write` are given an invalid `expires_at` or `expires_in` time.
+# Options are `true`, and `false`. If `false`, the exception will be reported
+# as `handled` and logged instead.
+#++
+# Rails.application.config.active_support.raise_on_invalid_cache_expiration_time = true
```
Splines, line 32: I think setting this to the new default `true` will be fine. What do you think?

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -0,0 +1,87 @@
...
+#
+# For more information, see
+# https://guides.rubyonrails.org/v7.1/configuring.html#config-active-support-message-serializer
+#
+# If you are performing a rolling deploy of a Rails 7.1 upgrade, wherein servers
+# that have not yet been upgraded must be able to read messages from upgraded
+# servers, first deploy without changing the serializer, then set the serializer
+# in a subsequent deploy.
+#++
+# Rails.application.config.active_support.message_serializer = :json_allow_marshal
```
Splines, line 63: I'm not even sure if [this](https://guides.rubyonrails.org/v7.1/configuring.html#config-active-support-message-serializer) affects us. Do we make use of the `ActiveSupport:MessageEncryptor` or the `ActiveSupport::MessageVerifier`?

If no, maybe we should use the default for Rails 7.2, which will be `:json`? Otherwise, we could also try `:message_pack`, but that requries the `msgpack` gem, which we don't have in our Gemfile right now. If we're not using this anyways, `:json` is a viable option I guess.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -0,0 +1,87 @@
...
+
+# ❓ Should discuss what we want here.
+# See: https://guides.rubyonrails.org/v7.1/configuring.html#config-active-record-commit-transaction-on-non-local-return
+###
+# Whether a `transaction` block is committed or rolled back when exited via
+# `return`, `break` or `throw`.
+#++
+# Rails.application.config.active_record.commit_transaction_on_non_local_return = true
```
Splines, line 71: We don't have any occurrences of `.transaction` in our code, so this is not a problem for us, right? That is, we can just accept the new default here.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -0,0 +1,87 @@
...
+#
+# Changing this default means that all new cache entries added to the cache
+# will have a different format that is not supported by Rails 7.0
+# applications.
+#
+# Only change this value after your application is fully deployed to Rails 7.1
+# and you have no plans to rollback.
+# When you're ready to change format, add this to `config/application.rb` (NOT
+# this file):
+#   config.active_support.cache_format_version = 7.1
```
Splines, line 87: > Only change this value after your application is fully deployed to Rails 7.1
> and you have no plans to rollback.

So, the plan is to tackle this in a subsequent PR in a few weeks once Rails 7.1 is running in production for at least a few weeks. Sounds good?

Splines:
@fosterfarrell9 Could you go through the checklists in my initial PR comment and tick off everything that's ok in your opinion? Note that I've only included changes I found noteworthy in particular.

fosterfarrell9:
> I've checked that for our controllers by searching for "==". I could only find comparison to string or integers but not to any hashes. Maybe you can take a look as well?

I did not find anything as well.

config/initializers/new_framework_defaults_7_1.rb
fosterfarrell9: I agree.

config/initializers/new_framework_defaults_7_1.rb
fosterfarrell9: We do not use this (at least not actively), so `:json` seems fine to me.

config/initializers/new_framework_defaults_7_1.rb
fosterfarrell9: Yes, I think so.

config/initializers/new_framework_defaults_7_1.rb
fosterfarrell9: Sounds good to me.

Splines: Comment added here: https://github.com/MaMpf-HD/mampf/issues/554#issuecomment-2045979728

Splines:
Due to the Rails upgrade, we also somehow upgraded rubocop to a newer version, that's why it's failing here. Will merge #617 into `dev` beforehand, then bring the changes back into this branch.

config/environments/test.rb
```diff
@@ -6,12 +6,14 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.cache_classes = false
+  # While tests run files are not watched, reloading is not necessary.
+  config.enable_reloading = false
```
Splines, line 10: Note that I changed the behavior here.

config/application.rb
```diff
@@ -9,25 +9,22 @@
 module Mampf
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults(7.0)
+    config.load_defaults(7.1)
```
Splines, line 12: Here we load all Rails `7.1` defaults.

Splines:
In f486935, I made the changes regarding `globalize_attribute_names` according to [this `mobility` issue](https://github.com/shioyama/mobility/issues/356).

### For reviewers
- [ ] I've verified that it works for `Division`, could you verify it for `Program` and `Subject`? (since I don't know where to find the latter two in the UI and I'd have to look through the code where we use it, maybe you can show it to me today in persona).

fosterfarrell9 (approved):

Splines:
@fosterfarrell9 Could you resolve the one missing conversation?

PR #617, opened 2024-04-15
Update Rubocop version and enforce all cops

Splines:
Commands used to run Rubocop

```
bundle exec rubocop -A  # in a second run
```

## For reviewers
- [x] You have to run `bundle install` locally to install the new versions.
- [x] Please review the unsafe cops carefully as they might introduce breaking behavior. E.g. the autocorrection of [`Style/MapIntoArray`](https://docs.rubocop.org/rubocop/cops_style.html#stylemapintoarray) failed for us because we return an object and not an array. I fixed that, also see the commit body of afccc70.
- [x] Note that I disabled one potential fix in a migration file as we don't want to alter db migration files once they've already been deployed to production.


fosterfarrell9 (dismissed):

app/models/submission.rb
```diff
@@ -63,7 +63,7 @@ def correction_size
   end
 
   def preceding_tutorial(user)
-    assignment.previous&.map { |a| a.tutorial(user) }&.compact&.first
+    assignment.previous&.filter_map { |a| a.tutorial(user) }&.first
```
Splines, line 66: Reviewing the code, I'm not sure why the autocorrection of this cop is not marked as safe. The justification is not too clear in my opinion:

> This cop’s autocorrection is unsafe because `map { …​ }.compact` that is not compatible with `filter_map`. ~ from [here](https://docs.rubocop.org/rubocop-performance/cops_performance.html#safety-11)

Having looked at the API for [`compact`](https://apidock.com/ruby/v2_5_5/Array/compact) (remove `nil` elements from array) and [`filter_map`](https://ruby-doc.org/3.1.4/Enumerable.html#method-i-filter_map):

> Returns an array containing truthy elements returned by the block. With a block given, calls the block with successive elements; returns an array containing each truthy value returned by the block:

it seems like this cop would be safe. But it's not, so I'm just wondering what edges cases there might be and if it's really safe to accept this change.

fosterfarrell9: I do not see why it shouldn't be safe in our situation (when I reviewed the code I studied the same API that you linked, looked at sveral blogs where examples where given and played around a little in the Rails console).

Splines: Ah, I found [this issue](https://github.com/rubocop/rubocop-performance/issues/280), which is a case where it gives different results, but this should not affect us here.

Splines: I've opened https://github.com/rubocop/rubocop-performance/pull/450 to improve the wording in the docs.

Splines: You may resolve this conversation.

fosterfarrell9 (approved):

PR #621, opened 2024-04-16
Change config files -- dummy PR for CI/CD testing

Splines:


Splines:
working as expected (pipeline fails with trailing whitespaces detected, among others)

PR #625, opened 2024-04-22
Remove emergency link

Splines:
As discussed in persona, the emergency is no longer needed as we don't think it will provide any benefit. The argument is that even motivated lecturer won't come up with a dedicated website tailored to help with specific content-related things in a video. Lecturers could link to the helpdesk service, but if this is offered in a semester, students will know this for sure through other communication channels. For students that have a specific question at a specific timestamp of a video, such an emergency link will probably never provide the tailored help they need. That's why we remove it here from the annotation tool.

## For reviwers

The schema is updated in this PR via a new migration.

- [x] Please make sure that there are is no leftover code regarding the emergency link.
- [x] Students should still be able to create/edit annotations
- [x] Lecturers won't see the "edit emergency link" field in the settings of a lecture anymore


fosterfarrell9 (approved):
LGTM.

PR #628, opened 2024-04-26
Redesign lecture edit page for lecturers

Splines:
Current preview:
image

## TODO
- [x] Add deep linking according to [this guide](https://webdesign.tutsplus.com/how-to-add-deep-linking-to-the-bootstrap-4-tabs-component--cms-31180t) and [this SO answer](https://stackoverflow.com/a/52456272/9655481)
- [x] Avoid redirect with `Cancel` button
- [x] Fix layout issues
- [x] Fix sample DB `Save` & `Cancel` buttons
- [x] Fix `Comments` save button not saving any changes (it even triggers an exception)

## For reviewers

_Sorry for the over 1,000 line insertions/deletions in this PR. This is mainly due to having migrated the `lectures.coffee` file to `lectures.js` and due to copying/pasting some old content to new locations while also modifying it, such that Git does not recognize it as replacements and instead deletions/insertions._

- [x] Headers should be internationalized
- [x] Every functionality that was in the accordion beforehand, should still work.
- [x] Clicking on the "cancel" button should not redirect to the "content" tab but stay in the current tab
- [x] Jumping directly to a tab via a url should work, e.g. [this one](http://localhost:3000/lectures/1/edit#assignments)
- [x] Accessibility: navbar header can be navigated by using Tab, then Arrow keys (up down as well as left right)
- [x] URL hashes (e.g. "#settings") should be a good naming choice as they shouldn't be changed in the future very often (to allow easy bookmarking for example)
- [x] Announcements should be scrollable (reproduce by creating at least 10 new announcements)
- [x] Importing media should still work.


app/views/lectures/edit/_form.html.erb
```diff
@@ -1,63 +1,191 @@
...
+              locals: { lecture: lecture } %>
+          <%= render partial: 'lectures/edit/erdbeere',
+              locals: { lecture: lecture } %>
+        </div>
       </div>
     </div>
   </div>
+
+
+  <!-- TODO (Lectures): Find out when these partials are used -->
```
Splines, line 176: we may have to talk about these partials here

Splines: I went through all modals in the UI and they still work fine. Exception to this are the special cases for seminars, I did not go through them in the UI. However, everything should work as beforehand since this PR just changes the UI representation. All elements have been on the same page beforehand (in the accordions) and now still are on one page (with a tab navigation bar), so data loading still stays the same.

I wondered where the `tags/modal` is used and have not found a button to activate on the lectures edit page.

fosterfarrell9 (changes requested):
I think the the behaviour concerning where you land after having saved some changes should be more uniform. Also, I think that since the tabs are now way more easily accesible, the default should be that after having saved something, one should stay in the tab instead of being placed in the content tab.

- If I change something in the "Preferences" tab, there appears a "Save and Exit"-Button and I land back in the content tab ( I would probably prefer a "Save" button and would want to remain in the "preferences" tab).
- In the "People/Tutorials" tab, if I add an editor, there appears a "Save" button, which then takes me back to the "Content" tab (again, I would prefer to remain in the tab). If I add a tutorial, I remain in the tab.
- In the "Orga" tab, a "Save and Exit" button appears and takes me back to the "Content" tab. 
- In the "Communication" tab, most of the actions you can do will let you stay in the tab, only the checkbox and radiobutton will produce "Save and Exit" buttons that take you back to the "Content" tab.
- In the "Assignments" tab, creating an assignment lets you stay in the tab, making other changes take you out of it.
- In the "Info" tab, there is a "Save" button even if nothing was changed (in contrast to the behaviour in the other tabs).

Another thing that I noticed is that the use of the "back" and "forward"-buttons of the browser is somehow broken (sometimes you land at unexpected pages, sometimes they just dont do anything). Probably a turbolinks issue.

app/views/lectures/edit/_form.html.erb
Splines: We came to the conclusion that this modal is not used here anymore, so I deleted the respective line in 6ea1e758

Splines:
Thanks for the review. I've addressed all your points in the last commits.
- Any action on any subpage (e.g. `#people`) should let the user stay on the current subpage. Note that for some actions, we reload the page, which is why some small visual flickering might appear as the "Content" tab is loaded first, then we jump to the correct section. This could be addressed in a future PR, but is maybe not worth the effort (one would have to implement some kind of js-data-storage mechanism or wrap the forms in partials, create a submit/cancel route and send custom JS from the backend such that the partial is reloaded with the initial data. This way, we wouldn't have to reload the entire page.
- The back/forward buttons should now work properly again.

fosterfarrell9 (changes requested):
LGTM, I played around with it for some time. I only stumbled upon one very strange bug involving the navigation of the navbar headers by Tab and arrow keys. If you make use of that and return to the content tab, the layout for the content tab is suddenly broken.
Screenshot 2024-05-28 192644

Splines:
> I only stumbled upon one very strange bug involving the navigation of the navbar headers by Tab and arrow keys. If you make use of that and return to the content tab, the layout for the content tab is suddenly broken.

I've listened to the wrong event listener, see the correct ones [here](https://getbootstrap.com/docs/5.3/components/navs-tabs/#events). I'm now listening to `shown.bs.tab`, i.e. only initialize the grid once the content of the tab is already loaded. Otherwise, the `focus` event triggered too early and the grid system was initialized when the content was not visible yet, therefore you saw the broken layout. It should now work flawlessly, even when resizing the browser.

fosterfarrell9 (approved):

PR #629, opened 2024-04-27
Fix broken interactions controller

fosterfarrell9:
When clicking the `Export statistics` button in the admin view (corresponding to the `/interactions` route), we got an error. In this PR we fix the broken interactions controller by adding the missing `current_ability` method.

### DIFF
--- a/app/controllers/interactions_controller.rb
+++ b/app/controllers/interactions_controller.rb
@@ -3,6 +3,10 @@ class InteractionsController < ApplicationController
   authorize_resource
   layout "administration"
 
+  def current_ability
+    @current_ability ||= InteractionAbility.new(current_user)
+  end
+
   def index
   end
 


Splines (approved):
For posterity: `current_ability` docsstrings are available [here](https://github.com/CanCanCommunity/cancancan/blob/178d2c5bdf97225751a1a089e9815a4060b03265/lib/cancan/controller_additions.rb#L342-L355). The change in this PR makes sense according to [the CanCanCan docs](https://github.com/CanCanCommunity/cancancan/blob/178d2c5bdf97225751a1a089e9815a4060b03265/docs/changing_defaults.md?plain=1#L20-L27).

I confirm that the `/interactions` route now works again in the frontend 🎉

Splines:
Use `Squash and merge` to merge into `dev`.

PR #632, opened 2024-04-29
Downgrade Rack to 2.2.9

fosterfarrell9:
Due to some change in policy regarding how rack parses requests from version 3 on, some of our nested forms are parsed in way that produces an incorrect parameters hash in the controller. In this hotfix we downgrade to Rack 2.2.9. We have to go through each form, update them and migrate to Rack 3 at a later time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem "jbuilder"
 # Reduces boot times through caching; required in config/boot.rb
 gem "active_model_serializers"
 gem "bootsnap", ">= 1.4.2", require: false
-gem "rack"
+gem "rack", "<3"
 # Use CoffeeScript for .coffee assets and views
 gem "coffee-rails", "~> 5.0.0"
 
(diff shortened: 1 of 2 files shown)


Splines:
related: https://github.com/rack/rack/issues/2128

and: https://github.com/rails/rails/issues/49601
and: https://github.com/rails/rails/issues/49783

Splines:
To reproduce: upgrade to Rack 3, then go to the profile page. There should be an error thrown due to nested forms and how parameters are passed as arrays.

fosterfarrell9:
There is an update of the UPGRADE guide on the Rack gem site whivh [explains the new behavious vr old behaviour](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#invalid-nested-query-parsing-syntax). See also the changes I made in [this erdbeere PR](https://github.com/MaMpf-HD/erdbeere/pull/6) in which I fixed the same issue. We should move this up to higher priority since

> Rack 2.2
> This version of Rack is receiving security patches only, and effort should be made to move to Rack 3.

PR #647, opened 2024-05-31
Delete users that haven't logged in for too long

### LINKED ISSUE #410: User cleaner does not detect all bounces correctly
The UserCleaner  worker which is run periodically to autodetect users who have mail adresses that are no longer valid does not detect many of the bounces.

Splines:
Fixes #410.

### Description

- The previous user cleaner logged in to the MaMpf IMAP server and looked for hashes in mails it had sent before to users to make sure they didn't bounce. For bounces, the respective users were deleted immediately.
- This behavior was deactivated as "unwanted" in #599. It's too rigorous. And maybe a mail server happens to be down that evening and whoop, that user is deleted. We don't want that.
- Since #553, we track the last sign in date of users. This puts us in a new position: we can now use the last sign in date to delete users that haven't signed in for more than 6 months. We send deletion warning mails 40days, 14days, 7days and 2days before the actual deletion happens. The deletion of the user itself is irreversible.
- For the complete flow, please see https://github.com/MaMpf-HD/mampf/issues/410#issuecomment-2136875776 or rather the docstrings for the `UserCleaner` class. Please complain if they don't explain some concepts that are actually implemented!

### Changes

- Implement the flow in the `UserCleaner` class (with unit tests).
- Previous code in the `UserCleaner` is entirely deleted.
- Schema changes: Remove old `ghost_hash` for users. Add new `deletion_date` field for users.
- Don't miss out on the new initializer `config/initializers/after_sign_in.rb`

#### No mail bounce detection anymore

What about mail bounce detection? Sadly enough, detecting mail bounces is a non-trivial problem. Parsing the body of mails is not robust enough as texts like "Mail could not be sent" are not standardized. There exists a good strategy called [Variable envelope return path (VERP)](https://en.wikipedia.org/wiki/Variable_envelope_return_path) which sounds very promising. However, we might ask ourselves if it's worth the effort to implement this in light of the benefits we get.

Yes, a mail server of a user might be down for some time. But how likely is it that a user hasn't logged in to MaMpf for at least 6 months, then their mail server is down when we send the warning deletion mail 40 days, 14 days, 7 days AND 2 days before actual deletion? Considering how unlikely this event is, I think we can invest our time better instead of implementing a bounce detection algorithm that has to log in to our own IMAP server to retrieve all emails. It's also a lot less pain to maintain the UserCleaner without such a brittle component.

### TODO

- [x] Check what happens to comments that a generic user posted. We should have a "Deleted User" anonymous user that is assigned to those comments (or at least have that written in the comment title) -> comments are currently deleted, I will tackle this in a separate PR
- [x] Check what happens to media that a generic user posted in a seminar talk. We probably want to completely delete it. -> while the user is deleted, the seminar still exists (with an empty user) and the media is preserved. We may want to change this  behavior in a separate PR.
- [x] Add maximum limit of deletions per run.

### Reviewers

The best place to start the review is probably the specs file for the UserCleaner.

> [!CAUTION]
> The code presented here might have very **destructive consequences** if something goes wrong. In the worst case scenario, it will delete active users that we didn't intend to delete. Unit tests should cover these cases but please be extremely vigilant with this PR and take extra steps and time to make sure the unit tests cover all edge cases and actually reflect semantically-wise what we want to happen.

> [!TIP]
> If tests don't work locally, you might have to run the migrations in the test docker container first.
That is: `cd ./docker/test`, `docker compose up -d`, `docker compose exec mampf bash`, `rake db:migrate`.

> [!TIP]
> In the `Files changed` tab on GitHub, you can mark files as done to get a better overview of what you've already reviewed. Every single line of code here should be looked through.

### Before merge
- [x] Plan when this code will come to the main branch, so that everyone (including sysadmins) are on standby in case something goes wrong.
- [x] Check that our "import backups" strategy still works flawlessly and can be fully executed within 1 hour.
- [ ] Set `MAX_DELETIONS_PER_RUN` env variable in production/dev/experimental
- [ ] Set `PRODUCTION_NAME=${PROJECTNAME}` for `docker compose up -d` in redeploy scripts for production/dev/experimental


Splines:
Unit tests related to this PR all run through as of commit 18bd4f4f2. The whole suite of unit tests just fails due to #650 (watchlist tests).

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,121 @@
...
+  #
+  # Note that this method just serves as a safety measure. The deletion date
+  # should be unset after every successful user sign-in, see the Warden callback
+  # in `config/initializers/after_sign_in.rb`. If for some reason, the callback
+  # does not work, this method will prevent active users from being deleted
+  # as a last resort.
+  def unset_deletion_date_for_recently_active_users
+    inactive_users_cached = inactive_users
+
+    User.where.not(deletion_date: nil).find_each do |user|
```
fosterfarrell9, line 68: Iterating over all users where the deletion date is not `nil` could become rather slow at some point. If you wanted to make this method faster, you could define an `active_users` method similar to the `inactive_users` method and then use something like `active_users.where.not(deletion_date: nil).update_all(deletion_date: nil)` to have it done with very few SQL operations. But since we loop over all these users in several other places as well, we might also just leave it here as it is, so feel free to resolve this without any code change.

spec/models/user_cleaner_spec.rb
```diff
@@ -1,19 +1,187 @@
 require "rails_helper"
 
 RSpec.describe(UserCleaner, type: :model) do
-  it "has a factory" do
-    expect(FactoryBot.build(:user_cleaner))
-      .to be_kind_of(UserCleaner)
+  describe("#set/unset_deletion_date") do
```
fosterfarrell9, line 4: Technically, `UserCleaner` has now become a worker instead of a model because `user_cleaner.rb` is placed in the workers folder. Would it not be more logical to place the `user_cleaner.rb` file in the models folder and rename the worker to `user_cleaner.rb` (the worker folder contains jobs by design)?

db/migrate/20240530200000_add_deletion_date_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeletionDateToUser < ActiveRecord::Migration[7.0]
```
fosterfarrell9, line 1: I am slightly confused here because we are on Rails 7.1 so the migration file should state `ActiveRecord::Migration[7.1]`.
Autogenerated migrations contain the correct version, I checked that.

db/migrate/20240605200000_remove_ghost_hash_from_user.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveGhostHashFromUser < ActiveRecord::Migration[7.0]
```
fosterfarrell9, line 1: Same as above.

fosterfarrell9 (changes requested):
LGTM, only some minor comments. I tested it intensively in the development setting using the rails console to manipulate user data. I worked as it should.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,121 @@
...
+  # warning mails sent on a given day might be higher than this number:
+  # - If on a given day the cronjob is not run (for whatever reason),
+  #   we have more users with a deletion date lying in the past than
+  #   MAX_DELETIONS_PER_RUN. However, we don't send an additional mail once
+  #   the user is deleted, so this shouldn't be a problem.
+  # - Despite that there cannot be more than MAX_DELETIONS_PER_RUN users with the
+  #   same deletion date, warning mails might be sent on the same date to users
+  #   with varying deletion dates, since the 40-, 14-, 7- and 2-day warning mails
+  #   can overlap temporally.
+  MAX_DELETIONS_PER_RUN = 50
```
fosterfarrell9, line 31: Do you think it would make sense to put the actual number in an environment variable? The we would not have to open new PR if we wanted to change it.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,121 @@
...
+  # The deletion date for the users must have been set beforehand by calling
+  # `set_deletion_date_for_inactive_users`.
+  def delete_users_according_to_deletion_date!
+    num_deleted_users = 0
+
+    User.where("deletion_date <= ?", Date.current).find_each do |user|
+      next unless user.generic?
+
+      user.destroy
+      num_deleted_users += 1
```
fosterfarrell9, line 89: After having submitted the review, I realized there is one case where we possibly have to do more. That is the case when the user is a generic user but has created media for a talk that they have given. Destroying that user will not destroy the corresponding media. In the preseeded development database, the user with id 6 falls into this category. In their case, the medium is owned by another user as well, it is co-owned by the user with id 5 who is not generic. If you strip the user with id 5 of its elevated status by revoking their editing rights for a lecture, you can delete user 5 as well and then you will have produced a medium with id 46 that has no editor. You can check in the console that it is invalid because of that. However, it seems that at least the views are still working. But having invalid media around with no editors is not something that we can have. I think that if the user has seminars talk with associated media, then the media should be deleted if the user is the only speaker of the talk or the medium has explicitly only the user as editor. That means some before destroy-callback would be needed on the user model.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,121 @@
...
+  # and updates their deletion date to be 40 days from the current date.
+  #
+  # The maximum number of deletion dates set in one run is limited by
+  # MAX_DELETIONS_PER_RUN.
+  def set_deletion_date_for_inactive_users
+    inactive_users.where(deletion_date: nil)
+                  .limit(MAX_DELETIONS_PER_RUN)
+                  .find_each do |user|
+      user.update(deletion_date: Date.current + 40.days)
+      UserCleanerMailer.with(user: user).pending_deletion_email(40).deliver_later
```
fosterfarrell9, line 51: We send out pending deletion mails also to non-generic users even if their accounts will not be deleted in the end. This might be confusing to the recipients of the mails as they might think that their account was deleted when it actually was not.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,121 @@
...
+  #
+  # In addition to the initial warning mail 40 days before deletion, this method
+  # sends warning mails 14, 7 and 2 days before the account is deleted.
+  def send_additional_warning_mails
+    User.where.not(deletion_date: nil).find_each do |user|
+      num_days_until_deletion = (user.deletion_date - Date.current).to_i
+
+      if [14, 7, 2].include?(num_days_until_deletion)
+        UserCleanerMailer.with(user: user)
+                         .pending_deletion_email(num_days_until_deletion)
```
fosterfarrell9, line 105: Same as above: These mails go out to all users, whereas deletion is not performed for all of them.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,121 @@
...
+  # `set_deletion_date_for_inactive_users`.
+  def delete_users_according_to_deletion_date!
+    num_deleted_users = 0
+
+    User.where("deletion_date <= ?", Date.current).find_each do |user|
+      next unless user.generic?
+
+      user.destroy
+      num_deleted_users += 1
+    end
```
fosterfarrell9, line 90: Also, shouldn't we possibly send a mail that confirms the deletion of the account if the user was destroyed?

app/workers/user_cleaner.rb
Splines: fixed in 08cae2aa62

spec/models/user_cleaner_spec.rb
Splines: When I started with this PR, the UserCleaner was indeed in the `models` folder. I changed its location since I didn't see why we would consider such a class as a model in the usual ORM kind of view where database fields are modeled in object-oriented code.

That's why I've placed it in the `workers` folder as that seemed to be the best place for it as far as I could see. Maybe you know a better place that would fit better? But what's wrong with having the actual job implementation next to file that defines the job and calls the implementation?

Splines: (Note that you can also attach PR comments to files themselves instead of specific code lines.)

app/workers/user_cleaner.rb
Splines: I will tackle this in a separate PR soon that we can merge before this one.

app/workers/user_cleaner.rb
Splines: Ups, that's true. Fixed in 28788595

app/workers/user_cleaner.rb
Splines: Added in d970c4fbcf

spec/models/user_cleaner_spec.rb
fosterfarrell9: As we discussed last week, it should be moved to the `models` for uniformity with the other workers. Changing the overall placing of PORO files would be a different PR for later.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,136 @@
...
+  MAX_DELETIONS_PER_RUN = ENV.fetch("MAX_DELETIONS_PER_RUN").to_i
+
+  # The threshold for inactive users. Users who have not logged in for this time
+  # are considered inactive.
+  INACTIVE_USER_THRESHOLD = 6.months
+
+  # Returns all users who have been inactive for INACTIVE_USER_THRESHOLD months,
+  # i.e. their last sign-in date is more than INACTIVE_USER_THRESHOLD months ago.
+  def inactive_users
+    User.where("last_sign_in_at < ?", INACTIVE_USER_THRESHOLD.ago)
```
fosterfarrell9, line 40: If you have users that never logged in since #553 then their `last_sign_in_at` column will bei `nil` which means that they will not be counted as inactive here.

fosterfarrell9 (changes requested):

config/locales/en.yml
```diff
@@ -2853,6 +2853,21 @@ en:
...
+      We've noticed that you haven't logged in to your MaMpf account in the last
+      6 months. Your account will be deleted in %{num_days_until_deletion} days
+      <strong>unless you log in to your MaMpf account before then</strong>.
+      Please note that once your account is deleted, all of your data stored
+      at MaMpf will be permanently lost.
+    deletion_body_html: >
+      <strong>Your MaMpf account has been deleted entirely as you have not
+      logged in for more than 6 months.</strong> Note that we've sent several
+      reminders beforehand. All of your data stored at MaMpf has been
+      permanently lost. Thanks for using MaMpf.
```
fosterfarrell9, line 2870: Shouldn't it be "have been lost"?

spec/models/user_cleaner_spec.rb
Splines: Done in d65228ae3 and f1184867d

config/locales/en.yml
Splines: In my opinion,  `data` should be treated as mass noun since "The data show us that.." doesn't just sound right in my ears, whereas "The data show**s** us that..." sounds better. I'm backed by a [random translation on Deepl](https://www.deepl.com/de/translator#de/en/Alle%20deine%20Daten%20sind%20verloren%20gegangen.) and many answers on StackOverflow, e.g. [this one](https://english.stackexchange.com/a/73963) as well as [Grammer Monster](https://www.grammar-monster.com/lessons/data_singular_or_plural.htm)  (don't know how trustworthy the latter one is haha). At least, they have a summary I can live with:

> "Data" is the Latin plural of "datum," but "data" is transitioning to a singular word that means "information." Many academics and statisticians retain "data" as a plural word, but if you naturally treat it as singular, that is now fully acceptable.

Also [thesaurus](https://www.thesaurus.com/e/grammar/data-is-or-data-are/) (which is hopefully more trustworthy) speaks for my point of view:

> The word data can be either singular or plural depending on meaning and context. In general usage, data is treated as singular when used as a mass noun to mean “information” and as plural when used to mean “individual facts.” In scientific and academic writing, data is almost always used as a plural noun. In digital technology, data is usually treated as a singular mass noun to mean “digitally stored information.” 

Especially:

> In the context of computers and digital technology, data is usually treated as a singular mass noun to refer to digitally stored information. 
> For example:
> User data is gathered by many different social media companies. 
All of the data on the hard drive was corrupted by a computer virus.

app/workers/user_cleaner.rb
Splines: Wow, thanks for noting that, that's really important! Fixed in 25041698ef

config/locales/en.yml
fosterfarrell9: > In scientific and academic writing, data is almost always used as a plural noun

That's probably the reason why I view it as a plural noun by default. But if digital natives think of it as a singular noun, that's fine by me.

fosterfarrell9 (approved):

Splines:
@fosterfarrell9 Do we want to disable the UserCleaner for our staging environments? Otherwise, me might lose users over there as they haven't logged in since a while...

fosterfarrell9:
> @fosterfarrell9 Do we want to disable the UserCleaner for our staging environments? Otherwise, me might lose users over there as they haven't logged in since a while...

I guess that would make sense.

app/workers/user_cleaner.rb
```diff
@@ -0,0 +1,13 @@
+class UserCleanerJob
+  include Sidekiq::Worker
+
+  def perform
+    # Only run this job in production, not for mampf-experimental or mampf-dev.
+    # Note that Rails.env.production? is not sufficient in this context
+    # as both mampf-experimental and mampf-dev also run in production mode.
+    production_name = ENV.fetch("PRODUCTION_NAME", nil)
+    return if ["mampf-experimental", "mampf-dev"].include?(production_name)
+
```
fosterfarrell9, line 10: An alternative here would be to write
`return unless production_name=="mampf"`
(or whatever name our production instance is given). This would be more stable if we should ever decide to establish another environment.

fosterfarrell9 (dismissed):
LGTM. Just one minor comment which you may choose to ignore.

app/workers/user_cleaner.rb
Splines: Yeah, you're right, I think your way is more robust. Modified in afe4ff3.

fosterfarrell9 (dismissed):

fosterfarrell9 (approved):

PR #651, opened 2024-06-17
Destroy talk media upon user deletion

Splines:
Related to #647.

We noticed that destroying users by calling `destroy` directly (instead of calling the `archive_and_destroy` method) might result in invalid media as it has no editors anymore. From now on, we want to delete talk media of users when the respective user is destroyed under the following conditions:

- If called from `archive_and_destroy` -> STOP and don't do anything
- The user must not have any other media than talk media, e.g. the user must not have edited any lecture media.
Otherwise -> STOP and don't do anything (*)
- A talk medium is only deleted if the user is the sole editor of that medium. That is, if there are multiple editors, we will just ignore that specific talk medium and try to delete the next one.

(*) "Don't do anything" is just a safety measure. In the UserCleaner in #647, those users are considered to be non-generic, so we shouldn't call their `destroy` method in the first place.


fosterfarrell9 (approved):
LGTM!

PR #652, opened 2024-06-18
Get Cypress up and running again (with docker and interactive testing)

Splines:
This PR does _not_ rework the cypress tests. Instead, it sets all the necessary infrastructure to run cypress tests locally and in our GitHub Actions pipeline without a big fuss.

### Start cypress interactive GUI
This may be subject to frequent changes as this PR is only a draft right now.

- ~~Install [VcXsrv Windows X Server](https://sourceforge.net/projects/vcxsrv/). Run XLaunch. In the "Extra settings", make sure to choose "Disable access control".~~<br>This is not needed anymore, new versions of WSL come with an X11 server, yeah 🙌 See the [WSLg repo](https://github.com/microsoft/wslg).<br>**The only thing you have to do is run `wsl --update`** (in Powershell).
- To start Cypress with the interactive test runner, execute the following command inside the `docker/test/` folder:
```
docker compose -f docker-compose.yml -f cypress.yml -f cypress-interactive.yml up --exit-code-from cypress
```

Wait a few seconds and the GUI should automatically pop up. Choose E2E testing and try to run the submission dummy test I've added. You should see the *edit profile* page at the end.

Note that cypress shares the test database with the Rails unit tests. This shouldn't be a problem as they both clean the whole database before they execute anything. But this means, we shouldn't run cypress and unit tests at the same time. I don't feel like this is a big issue locally, if you feel it is, it's easy to spin up another docker environment.

### cypress-on-rails & more links
We don't use `cypress-on-rails` anymore.

- [`on-rails.js`](https://github.com/shakacode/cypress-on-rails/blob/master/lib/generators/cypress_on_rails/templates/spec/cypress/support/on-rails.js)
- [Example for Authenticating a User](https://github.com/shakacode/cypress-on-rails/blob/master/docs/authentication.md#example-for-authenticating-a-user) (did not work)
- [A more succinct authenticating example](https://github.com/shakacode/cypress-on-rails/issues/16#issuecomment-669819936) -> that one works ;)
- [Official cypress guides for authenticating efficiently](https://docs.cypress.io/guides/end-to-end-testing/testing-your-app#Improving-performance)
- [A long list of Cypress Recipes](https://docs.cypress.io/examples/recipes/)

### TODOs
- [x] Add Cypress back to CI/CD
- [x] Use GitHub Actions formatter -> I've opened a feature request [here](https://github.com/cypress-io/cypress/issues/29786) -> ▶ probably not available right now, so will do in future PR
- [x] Long-term: write more robust cypress tests following [best practices](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements). The [starting guide](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress) is also a good resource. -> ▶ will do in future PRs
- [x] Fix ESLint errors in cypress files (these are not visible here in the pipeline as we didn't change these files)
- [x] Implement clean login mechanism in combination with app scenarios, unify interfaces
- [x] Provide `just` commands for easy startup of cypress GUI -> ▶ in future PR
- [x] Maybe find a better way than X11 server, see [here](https://github.com/cypress-io/cypress-docker-images/issues/1122) and my [StackOverflow](https://stackoverflow.com/q/78639075) question.
- [x] Add Cypress section to [wiki](https://github.com/MaMpf-HD/mampf/wiki/Testing-%E2%9C%94#-cypress-frontend-tests)


spec/cypress/e2e/submissions_spec.cy.archive.js
Splines, line 1: Note the GitHub diff doesn't work as expected here: I didn't add these lines, I just renamed `submissions_spec.cy.js` to `submissions_spec.cy.archive.js` and added a simple `submissions_spec.cy.js´ as dummy file (see below).

initialize.sh
```diff
@@ -20,6 +20,12 @@ check_for_preseeds() {
   fi
 }
 
+if [ "$RAILS_ENV" = "production" ]
+then
+    echo "This script is not intended for usage with RAILs_ENV=production. Aborting."
+    exit 1
+fi
+
```
Splines, line 28: The file is actually not used by the production environment as far as I can see. Please verify this.

fosterfarrell9: Checked it and can confirm.

fosterfarrell9 (approved):
LGTM. Great work!

PR #662, opened 2024-07-07
Don't use VSCode ESLint experimental flag anymore

Splines:
## **Changes**
- Don't use ESLint experimental flag anymore (it's now in the official plugin version)
- Add the `.docker` folder to hidden folders
- Remove custom Ruby LSP command. This was an attempt to mitigate https://github.com/Shopify/ruby-lsp-rails/issues/382, however it did not work (or rather was too much effort), so we just don't use the few Ruby LSP server-dependent features (for that we would have to run it inside docker containers) and stick with the static code analysis with is sufficient for us.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,7 +13,6 @@
         "source.fixAll.eslint": "explicit"
     },
     "eslint.format.enable": true, // use ESLint as formatter
-    "eslint.experimental.useFlatConfig": true,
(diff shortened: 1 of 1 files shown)


Splines:
Missing anything in terms of settings in VSCode? This might be a good time to add respective settings here.

fosterfarrell9 (approved):

PR #665, opened 2024-08-01
Fix DB preseeding

Splines:
This fixes DB preseeding that was introduced wrongly in #652.

For development, instead of seeding, we restore a sample database and then still use migrations.
For testing, we continue to use db seeding.

The respective Just command was adjusted to allow for reseeding without having to manually stop the mampf service beforehand. This will now be done automatically.

We also moved the SOLR indexing step to the end to avoid errors when SOLR is accessing the database while it is reseeded in the background.

Previously, the `db:create` command as issued twice, I've reduced it such that it only occurs once at the very beginning.


fosterfarrell9 (approved):
Tried it out, works nicely.

PR #668, opened 2024-08-02
Add annotations overview page & improve Cypress commands

### LINKED ISSUE #658: Index page for annotations
Currently, if you are a teacher and want to see all annotations that have been made for your videos, you would have to click through each video separately and look for annotations. The same problem arises if you are a student and want to look at all annotations you made. It would be great to have an index page for annotations.

Splines:
This resolves #658 

This adds a new page where users can view all of their annotations across all lectures. The list items can be clicked to jump to the timestamp at the annotation of the respective video.

image

image

### TODO
- [x] Color according to annotation
- [x] Button to open annotation in video -> no, instead just clickable list item
- [x] Pagination -> as annotations are grouped by lectures in this view, I don't think we need pagination. Even if the list becomes big, it's still manageable by just scrolling on the page a bit.
- [x] Categorize annotations in subsections for lectures
- [x] No annotations at all -> add link to blog post to see how to create them
- [x] Allow rendering of math content inside annotations -> had nothing to do here, worked automatically
- [x] Localization
- [x] Deal with annotations that were already posted as comments -> have no comment field anymore
- [x] Caching issue when accessing page with timestamp
- [x] Open annotation at timestamp (user shouldn't have to make another click to open the annotation)
- [x] Annotation overview for profs that want to view the annotations shared by students with them
- [x] Deal with special cases of media inheritance, e.g. media imported into another lecture
     -> who can see annotations... make consistent with what can be seen in the feedback player
- [x] Fix coloring: shared annotations shouldn't have the color from the annotation itself, but the category color
- [x] Use same icon in Thyme player for annotations than in the navbar
- [x] `annotation.medium.teachable.lecture.title` might trigger an exception since `teachable` is marked optional in `Medium`
- [x] Cypress tests

### Won't fix
- [ ] Default sorting -> I think the default sorting that Rails gives us is fine
- [ ] Fix math-content not displayed properly if truncated before end of dollar sign -> do we really want to fix that? I don't think there's an easy solution to that... Use the following string in an annotation to see the problem:
```
and yet another one here with $m = a t + h$ content and a $\sum_{i=0}^n i = \pi^2$

$$a+b$$

$c+d+e+f+g$ whenever it gets very long
```
- [ ] Notifications for new annotations? ->maybe something for the future

Don't be afraid by 700 lines of code changed, most of them are for Cypress tests.

- [x] Locally in dev, we only have one lecture with videos. Please add one manually to another one to test the functionality that we have multiple entries in the accordion. Note that in the Cypress tests, we have multiple lectures, so it's not a "problem" there.
- [x] Log in as student and as teacher to test different views (also done in Cypress tests)
- [x] Note that since the re-initialization of Cypress was performed very recently, I had to add/customize our current Cypress setup to accomodate for new needs that I didn't have in mind earlier on. I hope it's ok to have this in this PR as it would be a bit cumbersome to split it up into another PR first.


Splines:
@fosterfarrell9 I have implemented everything related to the annotation overview page. Feel free to check it out locally and play around with it. My plan is to only add Cypress tests now.

fosterfarrell9:
> annotation.medium.teachable.lecture.title might trigger an exception since teachable is marked optional in Medium

I do not think that this is aproblem here because we have

`  validates :teachable, presence: true, if: :proper?`

in `medium.rb`, and only RandomQuizzes are non-proper (and they should not get annotations).

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      student_annotations =
+        Annotation.where(medium_id: medium_ids_for_teacher_or_editor,
+                         visible_for_teacher: true)
+                  .map { |a| extract_relevant_information(a, is_shared: true) }
+      @student_annotations_by_lecture = student_annotations.group_by do |annotation|
+        annotation[:lecture]
+      end
+    end
+
+    render "annotations/annotations_overview"
```
fosterfarrell9, line 22: If the action is renamed to `index` and moved into the `AnnotationsController`, and the view is renamed to `index.html.erb` then this line becomes obsolete.

app/helpers/media_helper.rb
```diff
@@ -149,4 +149,14 @@ def external_link_description_not_empty(medium)
     # link url itself.
     medium.external_link_description.presence || medium.external_reference_link
   end
+
+  def video_link_timed(medium_id, timestamp)
+    play_path = Rails.application.routes.url_helpers.play_medium_path(medium_id)
```
fosterfarrell9, line 154: It is not necessary invoke `Rails.application.routes.url_helpers` if you are already inside a helper, `play_medium_path` should be sufficient. Also, it would be more in the Rails way to give `medium` as a parameter and use `play_medium_path(medium)`.

app/helpers/media_helper.rb
```diff
@@ -149,4 +149,14 @@ def external_link_description_not_empty(medium)
...
+    "#{play_path}?time=#{timestamp.total_seconds}"
+  end
+
+  def feedback_video_link_timed(medium_id, timestampe)
+    feedback_play_path = Rails.application.routes.url_helpers.feedback_medium_path(medium_id)
```
fosterfarrell9, line 159: Same as above.

app/views/annotations/_annotations_overview_accordion.html.erb
```diff
@@ -0,0 +1,72 @@
...
+  <% annotations_by_lecture.each_with_index do |(lecture, annotations), index| %>
+    <div class="accordion-item">
+      <% collapseLabel="annotation-overview-collapse-#{index}-#{is_students_annotations}" %>
+
+        <!-- Accordion header -->
+        <h2 class="accordion-header">
+          <button class="accordion-button collapsed" type="button" data-bs-toggle="collapse"
+              data-bs-target="#<%= collapseLabel %>"
+              aria-expanded="true" aria-controls="#<%= collapseLabel %>">
+            <%= lecture %>
```
fosterfarrell9, line 14: I think it would be nice to display how many annotations there are for this lecture (this might prevent one from needlessly opening the accordion).

app/views/commontator/comments/_body.html.erb
```diff
@@ -8,10 +8,11 @@
 <!--- Show timestamps in comments that have a linked annotation --->
 <% annotation = comment.annotation %>
 <% unless annotation.nil? %>
-  <% medium = annotation.medium %>
+  <% medium_id = annotation.medium.id %>
```
fosterfarrell9, line 11: Remove the `id` stuff, see above.

app/views/commontator/comments/_body.html.erb
```diff
@@ -8,10 +8,11 @@
...
   <% timestamp = annotation.timestamp %>
   <br>
   Timestamp:
-  <a href= '<%= Rails.application.routes.url_helpers.play_medium_path(medium) %>?time=<%= timestamp.total_seconds %>'>
-  <%= timestamp.hms_colon_string %> </a>
+  <a href="<%= video_link_timed(medium_id, timestamp) %>">
```
fosterfarrell9, line 15: Why not use the Rails helper for links?
`<%= link_to video_link_timed(medium, timestamp) do ... %>`
reads nicer (at least to me).

config/routes.rb
```diff
@@ -58,6 +58,10 @@
 
   resources :annotations, only: [:new, :create, :edit, :update, :destroy]
```
fosterfarrell9, line 60: If you add the index action to the AnnotationsController, then the annotation_controller should have all seven RESTful actions and you can just write 
`resources :annotations`
and delete the next lines.
What I am wondering is: Why is there a `show` action in the `AnnotationController` but no route for it? Is it unused?

spec/factories/annotations.rb
```diff
@@ -0,0 +1,33 @@
+FactoryBot.define do
+  factory :annotation do
```
fosterfarrell9, line 2: Now that we have a factory for annotations, I think it would be nice to have some unit tests for the `Annotation` model (possibly in a different PR).

spec/factories/lectures.rb
```diff
@@ -54,6 +54,21 @@
       end
     end
 
+    trait :with_title do
+      transient do
+        title { nil }
```
fosterfarrell9, line 59: Form my understanding (which may be wrong), this sets the default value for `title` as `nil`. I would expect that
`course { association :course, title: title }`
will fail if the `title` is `nil` (because of 
`  validates :title, presence: true, uniqueness: true`
in `course.rb`). Is this intended?

spec/factories/lectures.rb
```diff
@@ -54,6 +54,21 @@
...
+      end
+      # also see https://github.com/thoughtbot/factory_bot/issues/1391
+      course { association :course, title: title }
+    end
+
+    trait :with_teacher_by_id do
+      transient do
+        teacher_id { nil }
+      end
+      teacher { User.find(teacher_id) }
```
fosterfarrell9, line 69: Same as above: I would expect an exception for the default value. Is this intended?

spec/factories/media.rb
```diff
@@ -72,6 +72,13 @@
       sort { "Kaviar" }
       teachable_sort { :valid_lesson }
       after(:build) { |m| m.editors << m.teachable.lecture.teacher }
+
+      trait :with_lesson_by_id do
+        transient do
+          lesson_id { nil }
+        end
+        teachable { Lesson.find(lesson_id) }
+      end
```
fosterfarrell9, line 81: Same as above: I would expect an exception for the default value. Is this intended?

app/views/shared/_navbar.html.erb
```diff
@@ -43,6 +43,14 @@
           </li>
         <% end %>
 
+        <li class="nav-item">
+          <a href="<%= annotations_overview_path %>"
```
fosterfarrell9, line 47: If the change to an index action is made this would be annotations_path.

fosterfarrell9 (changes requested):

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
+class AnnotationsOverviewController < ApplicationController
+  layout "application_no_sidebar_with_background"
+
+  def show
```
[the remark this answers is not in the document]
Splines: Thanks for the detailed review. My original idea was to put it in the existing AnnotationsController, but that one already had a `show` method and I didn't know there was also an `index` method available. Will implement all your suggestions.

app/views/commontator/comments/_body.html.erb
Splines: But this will send the full medium information to the frontend, not just the id, won't it? If that's the case, I'd rather go for `.id` to reduce bandwidth (although that might be neglegible to be honest)

app/views/commontator/comments/_body.html.erb
Splines: ups, forgot to replace that occurrence thanks

spec/factories/lectures.rb
Splines: Yes, this indeed fails when the `title` is `nil` and that was my intention as I wanted to ensure the title is passed in from outside. But I see the point in having a default value for other tests, so will make sure there is one using `Faker`.

spec/cypress/e2e/annotations_overview_spec.cy.js
```diff
@@ -0,0 +1,240 @@
...
+    FactoryBot.create("annotation", "with_text",
+      { medium_id: context.medium1.id, user_id: context.user.id }).as("annotation1");
+    FactoryBot.create("annotation", "with_text",
+      { medium_id: context.medium2.id, user_id: context.user.id }).as("annotation2");
+    FactoryBot.create("annotation", "with_text",
+      { medium_id: context.medium3.id, user_id: context.user.id }).as("annotation3");
+    FactoryBot.create("annotation", "with_text",
+      { medium_id: context.medium3.id, user_id: context.user.id }).as("annotation4");
+  });
+}
```
Splines, line 63: With regards to the Cypress tests: I admit that it took way to long for me to come up with this "scenario creation" where we need a lecture, lessons and annotations. While trying out to find the right way to construct those, I thought if it wouldn't be better to create a kind of "scenario mechanism" (similar to what was available in the cypress plugin we used before).

But then the problem is that we need to verify that UI elements hold the data that it should. And not all such data is returned via FactoryBot, for example something is constructed by means of multiple fields in a model method like `def title`. If we really wanted to create "scenario Ruby files", one would have to find a way to return multiple objects at the end to the frontend and that would also be a bit of a mess with many params, e.g. in the current file: two lectures, some lessons, media and many annotations.

In the current way, whenever something is not returned by FactoryBot, we have to adjust the factory, e.g. include a `with_text` factory where we can pass in our own title. This way, we can check if the UI element holds that title, because we are in control of how it is constructed. Another possible approach could be to let FactoryBot construct the element and then later call the backend to call the `title` method on the respective elements. But that would require very specific endpoints and I'm also not too keen to create a "one fits all" cypress endpoint by letting us run arbitrary Rails commands from JS files.

It's a bit of a balancing act and with missing cypress experience I don't have the foresight to see which way will serve us "best" in the long run. With the current approach taken here, I think we are quite flexible. But there might be cases in the future, where we need more Rails functionality during the construction. Then, we could adjust the Rails helpers accordingly and if really necessary add more endpoints to cypress and hopefully those won't become rampant.

Splines: Another thing to mention: in FactoryBot, we sometimes need to pass the full object. E.g. somewhere in the MaMpf codebase, we do something along those lines

```ruby
@course = FactoryBot...
FactoryBot.create(:lecture, course: @course)
```

or maybe
```ruby
our_teacher = FactoryBot....
FactoryBot.create(:lecture, teacher: our_teacher)
```

However, in Cypress, when we use FactoryBot to construct `our_teacher`, we get back its JSON representation. We can't use that to pass it to the `teacher` field during the lecture construction (it has to be a Ruby object). Therefore, I had to add a `with_teacher_by_id` trait to the respective factory which seems to be a bit of a "by hook or by crook" way to get things done.

app/views/annotations/_annotations_overview_accordion.html.erb
```diff
@@ -0,0 +1,72 @@
...
+        <div class="accordion-collapse collapse" id="<%= collapseLabel %>"
+            data-bs-parent="#<%= accordionLabel %>">
+          <div class="accordion-body">
+
+            <% annotations.each do |annotation| %>
+              <a href="<%= annotation[:link] %>" style="text-decoration: none;">
+                <!-- Card shared by students will obtain their border color -->
+                <!-- according to their category via JS. -->
+                <% border_color_style = is_students_annotations ? \
+                  "" : "border-color: #{annotation[:color]}" %>
```
fosterfarrell9, line 28: This could be made into a helper method in `annotations_helper.rb` in order to get the business logic out of the view.

spec/cypress/e2e/annotations_overview_spec.cy.js
```diff
@@ -0,0 +1,240 @@
...
+    ].forEach((test, i) => {
+      cy.visit("/annotations/overview");
+      cy.getBySelector(CARD_SELECTOR).eq(i).as("card");
+      cy.get("@card").parents(".accordion-collapse").siblings(".accordion-header").click();
+      cy.get("@card").click();
+
+      cy.url().should("contain", `/media/${test.medium.id}`);
+      let timestamp = `0:00:${test.annotation.timestamp.seconds}`;
+      cy.getBySelector("current-time").should("contain", timestamp);
+      cy.getBySelector("annotation-comment").should("contain", test.annotation.comment);
```
fosterfarrell9, line 188: This fails in my setup.
Screenshot 2024-08-12 113157

spec/cypress/e2e/annotations_overview_spec.cy.js
```diff
@@ -0,0 +1,240 @@
...
+    ].forEach((test, i) => {
+      cy.visit("/annotations/overview");
+      cy.getBySelector(CARD_SELECTOR).eq(i).as("card");
+      cy.get("@card").parents(".accordion-collapse").siblings(".accordion-header").click();
+      cy.get("@card").click();
+
+      cy.url().should("contain", `/media/${test.medium.id}/feedback`);
+      let timestamp = `0:00:${test.annotation.timestamp.seconds}`;
+      cy.getBySelector("current-time").should("contain", timestamp);
+      cy.getBySelector("annotation-comment").should("contain", test.annotation.comment);
```
fosterfarrell9, line 237: This line fails in my setup as well.
Screenshot 2024-08-12 113909

app/views/annotations/_annotations_overview_accordion.html.erb
```diff
@@ -0,0 +1,72 @@
...
+          </button>
+        </h2>
+
+        <!-- Accordion body -->
+        <div class="accordion-collapse collapse" id="<%= collapseLabel %>"
+            data-bs-parent="#<%= accordionLabel %>">
+          <div class="accordion-body">
+
+            <% annotations.each do |annotation| %>
+              <a href="<%= annotation[:link] %>" style="text-decoration: none;">
```
fosterfarrell9, line 24: It might be worth thinking about whether we want the link to the video open in a new tab: This would be uniform with the behaviour of the "Play" button on media cards.

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      # TODO: teachable is marked optional in medium, so we should handle the case
+      # where it is nil
+      lecture: annotation.medium.teachable.lecture.title,
+      link: annotation_open_link(annotation, is_shared),
+      medium_title: annotation.medium.caption,
+      medium_date: annotation.medium.lesson&.date_localized
+    }
+  end
+
+  def annotation_open_link(annotation, is_shared)
```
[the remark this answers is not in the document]
Splines: moved to a helper

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+                  .map { |a| extract_relevant_information(a, is_shared: true) }
+      @student_annotations_by_lecture = student_annotations.group_by do |annotation|
+        annotation[:lecture]
+      end
+    end
+
+    render "annotations/annotations_overview"
+  end
+
+  def medium_ids_for_teacher_or_editor
```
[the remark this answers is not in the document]
Splines: moved to user model

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    @show_students_annotations = current_user.teachable_editor_or_teacher?
+    if @show_students_annotations
+      student_annotations =
+        Annotation.where(medium_id: medium_ids_for_teacher_or_editor,
+                         visible_for_teacher: true)
+                  .map { |a| extract_relevant_information(a, is_shared: true) }
+      @student_annotations_by_lecture = student_annotations.group_by do |annotation|
+        annotation[:lecture]
+      end
+    end
```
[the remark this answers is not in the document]
Splines: Moved some business logic to the user model (and also created a helper).

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    render "annotations/annotations_overview"
+  end
+
+  def medium_ids_for_teacher_or_editor
+    lectures = current_user.given_lectures + current_user.edited_lectures
+    lectures.map(&:media_with_inheritance).flatten.pluck(:id)
+  end
+
+  def extract_relevant_information(annotation, is_shared: false)
```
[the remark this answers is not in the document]
Splines: I disagree that this extraction is "not a controller method". I've now scoped it privately in the annotations controller and renamed it to `extract_annotation_overview_information`, but I think that what this method does is exactly what a controller is for. Here, we create the interface for the view by mapping model-specific data to a Ruby hash that will be sent as JSON objects to the frontend. Instead of `category`, we could call this field `superfield` and acceses it as `superfield` in the frontend and the modal wouldn't care. This shows that this code should reside in a more outer layer and is not core business logic (in my opinion). In clean architecture, this code would be part of the outer onion rings that can rot without affecting the business logic in deeper-located rings.

app/controllers/annotations_overview_controller.rb
Splines: modified this line to read: `render layout: "application_no_sidebar_with_background"` since we want to use the custom layout here

app/views/commontator/comments/_body.html.erb
Splines: What I said is probably premature optimization and Rails can probably handle that case just fine and still quite performant, so I just applied your suggestion.

app/views/commontator/comments/_body.html.erb
Splines: Ah, now I see what you meant, use `link_to` instead of the custom anchor link tag `<a>`. That's just my personal preference which probably stems from me not having available proper syntax highlighting for ERB syntax yet, while for `<a>` I get the well-established HTML syntax highlighting. But still changed it to `link_to` and hopefully we can find a good HTML ERB linter in the future.

config/routes.rb
Splines: To be honest, I don't know why we have a `show` method there. Things work just fine without it. Will investigate in future PR when I write cypress tests for the annotations.

app/controllers/annotations_overview_controller.rb
```diff
@@ -0,0 +1,54 @@
+class AnnotationsOverviewController < ApplicationController
+  layout "application_no_sidebar_with_background"
+
+  def show
+    user_annotations = Annotation.where(user_id: current_user.id)
+                                 .map { |a| extract_relevant_information(a) }
+    @annotations_by_lecture = user_annotations.group_by { |annotation| annotation[:lecture] }
```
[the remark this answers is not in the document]
Splines: done

app/views/annotations/_annotations_overview_accordion.html.erb
Splines: I wouldn't consider this "business logic", but still outsourced this to a helper method now.

spec/factories/lectures.rb
Splines: Ok, so I've added a default title via `Faker::Book` here.

spec/factories/lectures.rb
Splines: Here, I think I'd also expect an exception because the trait is called "..._by_id", so we expect the caller to pass in an id from the outside.

spec/factories/media.rb
Splines: See above

fosterfarrell9 (changes requested):

app/helpers/media_helper.rb
```diff
@@ -149,4 +149,14 @@ def external_link_description_not_empty(medium)
     # link url itself.
     medium.external_link_description.presence || medium.external_reference_link
   end
+
+  def video_link_timed(medium, timestamp)
+    play_path = play_medium_path(medium)
+    "#{play_path}?time=#{timestamp.total_seconds}"
+  end
```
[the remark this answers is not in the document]
Splines: nice, that's so much better!

app/controllers/annotations_overview_controller.rb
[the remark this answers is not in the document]
fosterfarrell9: What I meant to say was actually  "is not a controller *action*". Do we agree on this?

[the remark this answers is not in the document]
fosterfarrell9: Yes.

fosterfarrell9 (approved):

PR #671, opened 2024-08-13
Vouchers for user promotion - Part 2: Redemption

fosterfarrell9:
This PR continues #670. Vouchers can be redeemed in order to take over tutorials, become a lecture editor, become a teacher or become a speaker (of a talk in a seminar).

### To Dos

- [x] Redemption of tutor vouchers
- [x] Redemption of editor vouchers
- [x] Redemption of teacher vouchers 
- [x] Redemption of speaker vouchers 
- [x] Notifications for the redemption of tutor vouchers,  editor vouchers, teacher vouchers and speaker vouchers
- [x] Write unit tests
- [x] Write cypress tests

### Further To Dos (for reviewer)
- [x] Moving people around in lecture tutorials, should still be visible in dropdowns -> this is already covered by the existing tests
- [x] Test user deletion with redemptions -> should be deleted as well
- [x] Make sure secure hash also works with empty spaces at the beginning / at the end (when copied since oftentimes whitespaces are also copied)
- [x] Tell browser that secure hash text box shouldn't be auto-completable as to avoid storing these hashes as user suggestions.
- [x] Test sending of mails
- [X] Refactor VoucherProcessor
- [x] Refactor Lecture model
- [x] "Such a voucher [teacher voucher] will be invalidated directly after its redemption." -> make sure this is tested

### Out of scope
- Restrict the "My courses" invisible form to be rendered only when user is administrator. See `views/administration/index/_my_courses.html.erb`.
- ~~Allow instance methods to be passed to spec `FactoriesController` for Cypress specs~~<br>-> moved to #696
- ~~Refactor `UserCreatorController` for Cypress specs~~<br>-> moved to #696
- Fix some CSpell configuration (increase max number of problems shown in one file and enable spell checking in ruby files)

### Notes for the reviewer

The redemption of Vouchers is basically done via two actions of the `VouchersController`, namely `#verify` and `#redeem`.
In order to keep the controller lean, I introduced a service model `VoucherProcessor` in `app/services` (see e.g. [here](https://www.toptal.com/ruby-on-rails/rails-service-objects-tutorial)). Also, I extracted some stuff into the `Notifier` concern in `apps/controllers/concerns` since this contains some code also used by the `LecturesController`. On the model side, two new `ActiveRecord` models are used to store redemptions of vouchers: `Redemption` and `Claim`. The first one is basically a join table between `User` and `Voucher` (it stores that a certain user redeemed a certain voucher at a given time). The second one is a (polymorphic) join table between `Redemption` and so called `Claimables`, where the latter ones can be tutorials or talks. A claim describes what was taken over ("claimed") in the act of redeeming a voucher. In order to understand things it is probably best to play around a little in the GUI (with two users, one who is creating vouchers and one who is redeeming them) and observe what happens in the database. There are four different kinds of vouchers, hence there should be a lot to play around with. Depending on what kind of voucher that is redeemed, notifications in MaMpf and by email are issued:
For each kind of voucher, notifications in MaMpf are created for the teacher and editors of the lecture. Additionally,
- for the redemption of a teacher voucher, one email is sent to the new teacher and one to the previous teacher informing them of the change 
- for the redemption of an editor voucher, an email is sent to the new editor
- for the redemption of a speaker voucher, an email is sent to all cospeakers of the same talk informing them that another user has joined them

Note that redemptions only work additively: E.g. you can only take over tutorials, not resign from ones you are already a tutor of; similarly for the other kinds of vouchers. This is deliberate and by design: If you want to resign from a tutorial as a tutor you should contact a lecture editor. The same goes for the other kinds of status upgrades provided by vouchers.
I wrote unit tests for the vouchers model, the voucher processor and some security relevant routes in the vouchers controller. I have not yet written unit tests for some newly defined methods in the lecture model.


app/abilities/user_ability.rb
```diff
@@ -14,11 +14,7 @@ def initialize(user)
       user.admin? || (!user.generic? && user == given_user)
     end
 
-    can :fill_user_select, User do
-      user.active_teachable_editor?
-    end
-
-    can :list_generic_users, User do
+    can [:fill_user_select, :list_generic_users], User do
```
Splines, line 17: test

app/controllers/cypress/user_creator_controller.rb
```diff
@@ -12,11 +14,13 @@ def create
...
-                         locale: I18n.default_locale)
+                         name_in_tutorials: "#{role} Real Cypress Name",
+                         password: CYPRESS_PASSWORD, consents: true,
+                         admin: is_admin, locale: I18n.default_locale)
       user.confirm
 
-      render json: user.to_json, status: :created
+      additional_data = { password: CYPRESS_PASSWORD }
+      render json: user.as_json.merge(additional_data), status: :created
     end
```
Splines, line 24: This is a good idea to return the cypress password such that we only have to define it once. However, in the `commands.js` for Cypress, you never actually use this returned parameter as far as I can tell.

app/models/redemption.rb
```diff
@@ -0,0 +1,19 @@
...
+                               source_type: "Tutorial"
+  has_many :claimed_talks, through: :claims, source: :claimable,
+                           source_type: "Talk"
+
+  delegate :lecture, to: :voucher
+  delegate :role, to: :voucher
+  delegate :tutor?, to: :voucher
+  delegate :editor?, to: :voucher
+  delegate :teacher?, to: :voucher
+  delegate :speaker?, to: :voucher
```
Splines, line 18: This delegates to methods that stem from the definition of an enum in the `Voucher` class. It complicates the code path and adds one additional layer that is unnecessary IMHO. What is the advantage of having `redemption.editor?` over `redemption.voucher.editor?` (or something similar)

app/views/vouchers/_redeem_tutor_voucher.html.erb
```diff
@@ -0,0 +1,36 @@
...
+
+  <% if current_user.in?(voucher.lecture.tutors_by_redemption) %>
+
+      <%= t('profile.already_tutor_by_redemption') %>
+      <%= cancel_voucher_button %>
+
+  <% else %>
+
+    <p class="mt-4">
+      <%= redeem_voucher_button(voucher) %>
```
Splines, line 31: When you redeem a tutor voucher but haven't selected any tutorials, you still get the message `Your tutor status has been updated`. This might lead to confusion if users actually forgot to mark a specific tutor slot and think that everything worked fine. In my opinion, we should issue a warning instead and not allow the successfull termination of the user flow in this case.

app/models/lecture.rb
Splines, line 1: We might want to consider some better abstractions here as to not "dump everything" into the Lecture class which combines so many different behaviors. This might be a good place to gain some insights into how a refactoring of a fat model could look like.

I think this PR is the right place to refactor the voucher-related methods instead of postponing this to a future refactoring (that might be harder to do by then as some things are already established. We have the freedom now.)

Especially methods like` voucher_redeemers` or `tutors_by_redemption` feel like they don't belong in the file `lecture.rb`. But I haven't looked yet at where those methods are actually used, so this is just a first impression when glancing over this file.

Splines (changes requested):
Some first commets while I work through the code. So far, everything works flawlessly in my manual frontend tests.

app/views/vouchers/_redeem_tutor_voucher.html.erb
Splines: Should we show a warning, this should also be added to tests.

config/locales/en.yml
```diff
@@ -2062,7 +2111,56 @@ en:
       Request data
     data_request_sent: >
       We have received your request and will send you your data as soon as possible.
-    redeem_voucher: Redeeem Voucher
+    redeem_voucher: Redeem Voucher
```
Splines, line 2114: This might be a very contrived example that might not be worth to consider, just wanted to point it out:

- User verifies valid voucher code
- User selects valid tutorial
- Teacher removes tutorial
- User clicky on "redeem voucher". They get the message that their tutor status was updated successfully even though it wasn't as the tutorial didn't exist anymore at that time. This is not a security issue, it's just that the user might be confused after this message that they can't access any tutorials. But to be honest: what's the probability of this happening? I guess too low to bother...

app/controllers/cypress/user_creator_controller.rb
fosterfarrell9: I think it is used, namely in the following lines:

```js
Cypress.Commands.add("login", (user) => {
  return cy.request({
    method: "POST",
    url: "/users/sign_in",
    form: true,
    failOnStatusCode: true,
    body: {
      "user[email]": user.email,
      "user[password]": user.password,
    },
  }).then((response) => {
    expect(response.status).to.eq(200);
  });
});
```

Without returning the additional password  in `user_creator_controller.rb`,  `user.password` is `undefined` and the login fails (see our basecamp ping from August 24). That's why I introduced this.

app/models/redemption.rb
fosterfarrell9: I do not cling to the delegate methods. It is a Rails Active Support Core extension, see [here](https://guides.rubyonrails.org/active_support_core_extensions.html#delegate). As usual, there a are blog posts pro and contra delegation. If you think we should not use it, I have no problem when we delete these delegations.

app/views/vouchers/_redeem_tutor_voucher.html.erb
fosterfarrell9: I think being able to not select any tutorials (or talks) will be an important use case. Often, exact details for tutorials (or talks) are only known rather late, and it would be good that people can still acquire the status as tutor (or speaker), and then the teacher or an editor can assign tutorials (or talks) later when more information is known. So, in my mind, this is a feature. In a sense, the tutor status has been upgraded - from "not a tutor" to "tutor without tutorial (yet)". We might think about a more precise message to avoid confusion.

app/models/lecture.rb
fosterfarrell9: I agree, we should make some decisions in #694 before. These also affect e.g. whether we want to have service objects or not  (like the `VoucherProcessor`).

app/views/administration/index/_my_courses.html.erb
```diff
@@ -18,12 +18,14 @@
...
-                 locals: { course: Course.new } %>
-    </div>
+    <% if current_user.admin? %>
+      <div id="new-course-area"
+           class="bg-green-lighten-4 p-3 mb-3 rounded"
+           style="display: none;">
+        <%= render partial: 'courses/new',
+                   locals: { course: Course.new } %>
+      </div>
+    <% end %>
```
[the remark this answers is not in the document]
fosterfarrell9: Yeah, I agree, it is out of scope, but since the greeter topic here is "improvement of role management" I also feel we could leave it in here.

app/views/vouchers/verify.js.erb
```diff
@@ -0,0 +1,10 @@
+$("#redeem-voucher").empty()
+  .append("<%= j render partial: 'vouchers/redeem_' + @voucher.role + '_voucher', locals: { voucher: @voucher } %>");
+
+// activate selectize
+// @reviewer: This is used several times in our code base (in the coffeescript
+// variant). Should we extract this into a function in order to avoid code
+// duplication?)
```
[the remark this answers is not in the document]
fosterfarrell9: Okay. But maybe change the comment a little then?

app/helpers/notifications_helper.rb
```diff
@@ -26,6 +28,7 @@ def notification_color(notification)
     return "bg-post-it-blue" if notification.generic_announcement?
     return "bg-post-it-red" if notification.announcement?
     return "bg-post-it-orange" if notification.course? || notification.lecture?
+    return "bg-post-it-green" if notification.redemption?
```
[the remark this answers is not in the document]
fosterfarrell9: Years ago,  I invested a lot of time to mimick the colors of Post It notes which I was told were very popular among students (the notes, maybe not the colors) ;-)

config/locales/en.yml
fosterfarrell9: As I wrote above: The tutor status was still upgraded in the sense that the user can now be selected by the teacher for tutorials, so it is a "tutor without tutorial".

app/views/vouchers/verify.js.erb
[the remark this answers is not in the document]
fosterfarrell9: Just saw that this was already done. Thanks!

app/controllers/cypress/user_creator_controller.rb
Splines: moved to #696

app/models/redemption.rb
Splines: Removed in e2786e3f49

config/routes.rb
```diff
@@ -290,6 +290,14 @@
        to: "lectures#import_toc",
        as: "import_lecture_toc"
 
+  post "lectures/:id/become_tutor",
+       to: "lectures#become_tutor",
+       as: "become_tutor"
+
+  post "lectures/:id/become_editor",
+       to: "lectures#become_editor",
+       as: "become_editor"
```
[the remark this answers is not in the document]
fosterfarrell9: You're right. They are unused and should be deleted. They are artefacts of an earlier approach I employed in this PR, but abandoned at some point and apparently forgot these lines when I removed it.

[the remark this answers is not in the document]
Splines: removed in 71fc49aec8

app/views/vouchers/_redeem_editor_voucher.html.erb
```diff
@@ -0,0 +1,24 @@
+<p data-cy="redeem-voucher-text">
+  <%= t('profile.become_editor_html', lecture: voucher.lecture.title) %>
+</p>
+
+<% if current_user == voucher.lecture.teacher %>
+
+  <p>
+    <%= t('profile.teacher_cant_become_editor') %>
+    <%= cancel_voucher_button %>
```
Splines, line 9: If you put a redeem_voucher_button for this case, the teacher will remain the teacher but the editor will be removed from the lecture. That means, a teacher could possibly construct a POST request to do exactly this. But then, they could also do this directly in the UI by revoking the editor for the lecture, so we shouldn't bother. Just to let you know.

app/views/vouchers/_redeem_tutor_voucher.html.erb
Splines: Ah ok, yeah that makes sense. I've reworked the messages a bit in the latest commits.

config/locales/en.yml
Splines: Ah ok, yeah that makes sense. I've reworked the messages a bit in the latest commits.

config/application.rb
```diff
@@ -12,6 +12,10 @@ class Application < Rails::Application
     config.load_defaults(7.1)
     config.autoloader = :zeitwerk
 
+    # Autoload subfolders of modules (recursively)
+    # https://stackoverflow.com/a/4794775/
+    config.autoload_paths += Rails.root.glob("app/models/**/")
+
```
Splines, line 18: Unfortunately, as of d33072d, the ERD creation does not work anymore as it cannot find some classes. This is probably due to this modified `autload_paths`. Will investigate.

Splines: Fixed in 4ed73bccabf9

app/controllers/lectures_controller.rb
```diff
@@ -66,6 +67,7 @@ def edit
 
   def create
     @lecture = Lecture.new(lecture_params)
+    @lecture.teacher = current_user unless current_user.admin?
```
Splines, line 70: To understand this line, am I right that:
- `@lecture.teacher = current_user` was done beforehand implicitly by Devise
- `unless current_user.admin?` is needed such that upon creation of a new lecture, the teacher field is empty.

spec/controllers/lectures_controller_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    describe "POST #update" do
+      it "does not update the teacher" do
+        new_teacher = FactoryBot.create(:confirmed_user)
+        post(:update, params: { id: lecture.id,
+                                lecture: { teacher_id: new_teacher.id } })
+        lecture.reload
+        expect(lecture.teacher).to eq(teacher)
+      end
+    end
```
Splines, line 22: Oh nice, thanks, this is indeed a very important test ;)

app/views/lectures/_new.html.erb
```diff
@@ -52,23 +52,34 @@
...
+                       {},
+                       { class: 'selectize',
+                         data: { ajax: true,
+                                 model: 'user',
+                                 filled: false,
+                                 placeholder: t('basics.enter_two_letters'),
+                                 no_results: t('basics.no_results'),
+                                 current: current_user.id,
+                                 modal: modal } } %>
+        <% else %>
```
Splines, line 73: If this is from now on only accessible by admins: what is the `current_user` information used for in the form and why is it sent as data to the backend? Is it relevant?

Splines (changes requested):

app/views/tutorials/_form.html.erb
```diff
@@ -13,17 +13,12 @@
...
                      { class: 'selectize',
                        multiple: true,
-                       style: 'display: none;',
                        id: "tutorial_tutor_ids_#{tutorial.id}",
-                       data: { ajax: true,
-                               filled: false,
-                               model: 'user',
-                               placeholder: t('basics.enter_two_letters'),
-                               no_results: t('basics.no_results') } } %>
+                       data: { cy: "tutor-select"} } %>
```
Splines, line 21: For four dropdowns (teacher, editor, tutor, speaker), users can no longer search for arbitrary users. Did you test this somewhere?

Splines (changes requested):

app/views/vouchers/_redeem_editor_voucher.html.erb
fosterfarrell9: Thanks for noticing!

app/controllers/lectures_controller.rb
fosterfarrell9: Beforehand, the line

```ruby
allowed_params.push(:course_id, :teacher_id, { editor_ids: [] }) if action_name == "create"
```

in the `lecture_params` method made sure that `teacher_id` is whitelisted, and then

```
@lecture = Lecture.new(lecture_params)
```
would make sure that `@lecture.teacher_id` is set as in the parameter hash. Of course, the create action's authorization is checked in the corresponding ability file, so anyone who passes authorization (i.e. course editors or admins) could make an arbitrary user the teacher of the new lecture. Now the line

```ruby
allowed_params.push(:teacher_id) if current_user.admin?
```
makes sure that `teacher_id` is whitelisted only for admins. But in order to get a valid lecture, the `teacher_id`
needs to be set also in case the lecture is ccreated by a course editor that is not an admin. But the new default here is that users can only set themselves as teachers. This should also be reflected in the corresponding forms for creating a lecture.

app/views/tutorials/_form.html.erb
fosterfarrell9: No, I guess I missed this. I think in this direction I only wrote this one unit test that you commented on.

app/views/lectures/_new.html.erb
fosterfarrell9: Hm, it seems that it is used in `app/views/lectures/new.coffee` where we have these lines

```coffee
# make sure that there will always be a teacher selected
teacherSelector = document.getElementById('lecture_teacher_id')
sel = teacherSelector.tomselect
sel.on 'blur', ->
  value = sel.getValue()
  if value == ''
    sel.setValue(teacherSelector.dataset.current)
  return
```

Do we still need this?

app/views/tutorials/_form.html.erb
Splines: Ok, so I will add more tests to test this as well.

Splines (approved):
Finally, after quite some time, I approve this PR. With this, we have a really nice voucher system at MaMpf, thank you 🎉

Please make sure to review the changes I've made before **squash-merging**. Follow-ups to this PR will be a redesign of the profile page where we can make the voucher redemption more prominent. I will also add a warning box to dropdowns that don't allow non-admin users to search for arbitrary users anymore. Just to not confuse them when no results are shown in the dropdown search box.

PR #680, opened 2024-08-14
Clean up Gemfile & pin all gems

Splines:
- Pin gem versions instead of using the newest commit on the `main` branch
- Don't use `>=`, but `~>` instead, which will not automatically install major version updates
- Clean up comments in the Gemfile and remove old packages that were commented out. Note that when the comments are not on a separate line, VSCode can do the alphabetical ordering for us automatically.
- ⚠ Remove SQLite dependency: as far as I can tell, we don't use SQLite anywhere, is that correct? We have been using PostgreSQL instead for a long time.

In this PR, we don't _update_ any dependencies, i.e. we didn't do `bundle update`, just `bundle install` with the pinned versions. An exception is the SQLite dependency, which is removed as seen above.

### TODO
- [x] Pin version for every single gem! (Is there a bundler command for this that we could use? -> didn't find one, but GitHub Copilot generated a neat little script that saved me a lot of time searching the version numbers. Of course, you have to manually fix some mistakes (e.g. version number after the line comment), but this is easy to fix manually afterwards.

```ruby
# ./pin_gemfile_versions.rb

require "bundler"

# Parse the Gemfile.lock
lockfile_parser = Bundler::LockfileParser.new(Bundler.read_file("Gemfile.lock"))

# Read the original Gemfile
gemfile_content = File.read("Gemfile")

# Create a hash of gem names and their versions
gem_versions = lockfile_parser.specs.each_with_object({}) do |spec, hash|
  hash[spec.name] = spec.version.to_s
end

# Update the Gemfile content with version specifiers
updated_gemfile_content = gemfile_content.lines.map do |line|
  if line =~ /gem\s+["']([^"']+)["']/
    gem_name = Regexp.last_match(1)
    version = gem_versions[gem_name]
    if version
      line.sub(/$/, ", '~> #{version}'")
    else
      line
    end
  else
    line
  end
end.join

# Write the updated content back to the Gemfile
File.write("Gemfile", updated_gemfile_content)

```


fosterfarrell9 (approved):
LGTM! Indeed, SQlite has not been used by us for a long time.

PR #683, opened 2024-08-20
Fix lecture edit page if active term is not set

### LINKED ISSUE #682: Lecture edit page not working when no active term is not set
If `Term.active` is false, then calling the `lectures#edit` action results in an exception.

fosterfarrell9:
This fixes #682 by adding the missing case in the `older_than?` method in the `Lecture` model.


Splines (changes requested):
LGTM.

- Maybe you can add an end-to-end test to ensure that the lecture#edit page looks as expected? (instead of showing an error?)
- And maybe even add a unit test for the `lecture.stale?` method to ensure it gives proper results even in the case where we just have a lecture and not an active term set.

fosterfarrell9:
I added unit tests for `lecture#stale?`, made `lecture#older_than?` private (it is not used anywhere except by the lecture#stale? method) and added a cypress test for the lecture edit page that should fail when the page throws an error.

spec/cypress/e2e/lecture_spec.cy.js
```diff
@@ -0,0 +1,17 @@
+import FactoryBot from "../support/factorybot";
+
+describe("Lecture edit page", () => {
+  it("shows content tab button", function () {
+    cy.createUserAndLogin("teacher").as("teacher");
+
+    cy.then(() => {
+      FactoryBot.create("lecture", "with_teacher_by_id",
+        { teacher_id: this.teacher.id }).as("lecture");
+    });
```
Splines, line 10:
```suggestion
    cy.createUserAndLogin("teacher").then((teacher) => {
      FactoryBot.create("lecture", "with_teacher_by_id",
        { teacher_id: teacher.id }).as("lecture");
    });
```

spec/cypress/e2e/lecture_spec.cy.js
```diff
@@ -0,0 +1,17 @@
...
+
+    cy.then(() => {
+      cy.visit(`/lectures/${this.lecture.id}/edit`);
+      cy.getBySelector("content-tab-btn").should("be.visible");
```
Splines, line 14: LGTM, this will definitely fail if the page throws an error ;)

app/models/lecture.rb
```diff
@@ -830,12 +830,6 @@ def speakers
     User.where(id: SpeakerTalkJoin.where(talk: talks).select(:speaker_id))
   end
 
-  def older_than?(timespan)
-    return true unless term
-
-    term.begin_date <= Term.active.begin_date - timespan
-  end
-
   def stale?
```
Splines, line 833: Maybe one last thing: could you add a docstring including an explanation for what "stale" means? (doesn't have to include every detail, just a high-level overview maybe)

spec/models/lecture_spec.rb
```diff
@@ -86,6 +86,52 @@
...
+
+    context "when there is no active term" do
+      it "returns false" do
+        lecture = FactoryBot.build(:lecture)
+        expect(lecture.stale?).to be(false)
+      end
+    end
+
+    context "when there is an active term" do
+      let(:year) { 2024 }
```
Splines, line 100:
```suggestion
```

Splines (changes requested):

spec/models/lecture_spec.rb
```diff
@@ -86,6 +86,52 @@
     end
   end
 
+  describe "#stale?" do
+    let(:year) { 2024 }
```
Splines, line 90:
```suggestion
```

Splines (changes requested):

spec/cypress/e2e/lecture_spec.cy.js
Splines, line 1: rename to `lectures_spec.cy.js` (plural) to be consistent with other cypress spec files

Splines (changes requested):

spec/cypress/e2e/lecture_spec.cy.js
fosterfarrell9: Thanks for catching this!

Splines (approved):
LGTM. Note that you'll have to approve the tests such that they can run through before merging via `Squash and merge`.

fosterfarrell9:
run tests

PR #686, opened 2024-08-22
Fix annotation feedback not shown when share feature is disabled

Splines:
In #668, we added an overview page for annotations (`/annotations`). In the following scenario, the behavior was incoherent:

- Teacher enables "allow students to share annotations with the teacher" in the lecture settings
- Student creates an annotation and shares it with the teacher
- Teacher disables "allow students to share annotations with the teacher" in the lecture settings

In the annotation overview page, the annotation was then listed under the section for students annotations. However, if the teacher clicked on it, they were redirected to the feedback player but didn't see the annotation.

We came to the conclusion that **we want the annotation to still be visible to the teacher in this case**. This is justified since the student has deliberately clicked on "Make the annotation visible for the teacher" when creating the annotation, so the annotation should still be visible to the teacher regardless of the discussed lecture setting "allow students to share annotations with the teacher". In our interpretation, this option only refers to _newly_ created annotations.

In this PR, we fix the incoherent behavior by still displaying shared annotations to the teacher (in the feedback player) even if the option in the lecture settings (see above) is disabled.

### DIFF
--- a/spec/cypress/e2e/annotations_spec.cy.js
+++ b/spec/cypress/e2e/annotations_spec.cy.js
@@ -32,6 +32,7 @@ describe("Annotations visibility", () => {
         // Disable annotation sharing in lecture settings
         cy.visit(`/lectures/${this.lecture.id}/edit#communication`);
         cy.getBySelector("annotation-lecture-settings")
+          .should("be.visible")
           .find("input[value=0]").should("have.length", 1).click();
 
         // Click on submit button to save changes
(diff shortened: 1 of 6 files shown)


spec/cypress/e2e/annotations_spec.cy.js
```diff
@@ -0,0 +1,69 @@
...
+
+      cy.then(() => {
+        // Create new annotation
+        FactoryBot.create("annotation", "with_text", "shared_with_teacher",
+          { medium_id: this.medium.id, user_id: this.user.id }).as("annotation");
+
+        // Disable annotation sharing in lecture settings
+        cy.visit(`/lectures/${this.lecture.id}/edit#communication`);
+        cy.getBySelector("annotation-lecture-settings")
+          .find("input[value=0]").should("have.length", 1).click();
```
fosterfarrell9, line 35: Would it be possible to wait until the tab is actually displayed? If you watch the test in the GUI, at this point it looks like this:
Screenshot 2024-08-23 120819
I think it would be nicer to actually see what is clicked in the GUI (in order to get closer to the look that the users have) and to faster identify what is wrong if the test should fail in the future at some point.

fosterfarrell9 (changes requested):

spec/cypress/e2e/annotations_spec.cy.js
Splines: good point, I now wait for respective elements to be visible on the page as of 36ff03b8

fosterfarrell9 (approved):
LGTM.

PR #688, opened 2024-08-22
Update Gemfile dependencies (`bundle update`)

Splines:
I ran `bundle update` to update our dependencies according to our Gemfile (which I cleaned up recently in #680 and pinned Gems there). Might be worth checking out if there are new major versions for some gems (in a future PR).

I also updated my local bundler version by executing `bundle update --bundler`. You can see this in the last line in the `Gemfile.lock`. The newest version is `2.5.17`, released on August 1, 2024.

### TODO for myself & reviewers
- [x] Check that everything is still working with the new dependencies.
We should probably not encounter any problems as we pinned the gems in #680, such that they only undergo automatic _minor_ version upgrades. There could still be a bad apple somewhere...


fosterfarrell9 (dismissed):

Splines:
@fosterfarrell9 Good that we have tests: the update for the `js-routes` gem here broke MaMpf and you couldn't access some pages in the frontend anymore since `Routes. ...` was not defined anymore for JavaScript files. Some cypress tests detected this, e.g. your test that the lecture page shows a button.

Instead of fixing the underlying problem during the update, I've pinned `js-routes` to the older version. This is because there is already a new major version and I don't see the point in investing time here in upgrading a minor version if we could just as well upgrade to the major version soon, see #690.

fosterfarrell9 (approved):
LGTM.

PR #693, opened 2024-09-14
Deal with registration edge cases in UserCleaner

Splines:
In #647, we introduced the "UserCleaner" that deletes users that haven't logged in for too long. We now noticed that we don't deal with specific **edge cases related to user registration**. This means that our initial `where` clause

```ruby
User.where(last_sign_in_at: ...INACTIVE_USER_THRESHOLD.ago)
        .or(User.where(last_sign_in_at: nil))
```

(to determine which users are inactive) is not sufficient to handle all cases. The new special cases are described as docstrings in the `app/models/user_cleaner.rb`.

Furthermore, we also replace `last_sign_in_at` by `current_sign_in_at` in this PR. See [this Devise code](https://github.com/heartcombo/devise/blob/72884642f5700439cc96ac560ee19a44af5a2d45/lib/devise/models/trackable.rb#L20-L23) to see the difference. `current_sign_in_at` always holds the newest value of the last user sign in, so we want to use that instead of `last_sign_in_at`, which refers to the last sign in before the `current_sign_in_at` date. Yes, the names could have been chosen better by Devise 😅

### Reviewers

I admit that it was way more difficult than I originally thought to arrive at the new statement, which is a lot more complicated than the original one. Nevertheless, I think its worth it to also handle the edge cases, especially in order not to irritate users with deletion mails when they have just registered (and confirmed their account -> note that also in this case `last_sign_in_at` is still `nil`).

- [x] As a challenge: please try to write a test that breaks my new `inactive_users` method. I've done this multiple times to refine the statement and in the current state, I wasn't able to break it anymore. But maybe you can?
- [x] I've replaced multiple occurrences of `:user` by `:confirmed_user` in the tests such that they still work. Please make sure that this does not mean I forgot to handle cases where we should actually also test for non-confirmed users.
- [x] I'm sure there are ways to write the `where` statement in a more performant way, e.g. by resorting to direct SQL or by using some other tricks. I was happy that I arrived at any workable solution for all the edge cases in the first place and didn't measure anything related to performance. This is just to inform you about this.

### DIFF
--- a/app/models/user_cleaner.rb
+++ b/app/models/user_cleaner.rb
@@ -50,16 +50,11 @@ class UserCleaner
   #   but the user should only be considered inactive if the confirmation_sent_at
   #   date is older than the threshold.
   def inactive_users
-    User.where.not(confirmed_at: nil).and(
-      User.where(current_sign_in_at: ...INACTIVE_USER_THRESHOLD.ago)
-      .or(User.where(current_sign_in_at: nil,
-                     confirmation_sent_at: ...INACTIVE_USER_THRESHOLD.ago))
-    ).or(
-      User.where(
-        confirmed_at: nil,
-        confirmation_sent_at: ...INACTIVE_USER_THRESHOLD.ago
-      )
-    )
+    threshold = INACTIVE_USER_THRESHOLD
+    User.confirmed.and(
+      User.inactive_for(threshold)
+      .or(User.no_sign_in_data.confirmation_sent_before(threshold))
+    ).or(User.unconfirmed.confirmation_sent_before(threshold))
   end
 
   # Returns all users who have been active in the last INACTIVE_USER_THRESHOLD months,
--- a/app/models/user_cleaner.rb
+++ b/app/models/user_cleaner.rb
@@ -39,9 +39,27 @@ class UserCleaner
   #
   # Users without a last_sign_in_at date are also considered inactive. This is
   # the case for users who have never logged in since PR #553 was merged.
+  #
+  # Edge cases for registration (that refine the above statements):
+  # - A user might have registered but never actually logged in (confirmed their
+  #   email address). In this case, we don't look at the last_sign_in_at date
+  #   (as this one is still nil), but at the confirmation_sent_at date to
+  #   determine if the user is considered inactive.
+  # - Another edge case is when users have registered and confirmed their mail,
+  #   but never logged in after that. In this case, last_sign_in_at is indeed nil,
+  #   but the user should only be considered inactive if the confirmation_sent_at
+  #   date is older than the threshold.
   def inactive_users
-    User.where(last_sign_in_at: ...INACTIVE_USER_THRESHOLD.ago)
-        .or(User.where(last_sign_in_at: nil))
+    User.where.not(confirmed_at: nil).and(
+      User.where(last_sign_in_at: ...INACTIVE_USER_THRESHOLD.ago)
+      .or(User.where(last_sign_in_at: nil,
+                     confirmation_sent_at: ...INACTIVE_USER_THRESHOLD.ago))
+    ).or(
+      User.where(
+        confirmed_at: nil,
+        confirmation_sent_at: ...INACTIVE_USER_THRESHOLD.ago
+      )
+    )
   end
 
   # Returns all users who have been active in the last INACTIVE_USER_THRESHOLD months,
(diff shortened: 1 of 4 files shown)


app/models/user_cleaner.rb
```diff
@@ -37,17 +37,35 @@ class UserCleaner
...
+      User.where(current_sign_in_at: ...INACTIVE_USER_THRESHOLD.ago)
+      .or(User.where(current_sign_in_at: nil,
+                     confirmation_sent_at: ...INACTIVE_USER_THRESHOLD.ago))
+    ).or(
+      User.where(
+        confirmed_at: nil,
+        confirmation_sent_at: ...INACTIVE_USER_THRESHOLD.ago
+      )
+    )
   end
```
fosterfarrell9, line 63: I totally agree that readability and maintaianability is more important that performance here.
If you wanted to make it even more readableyou could do the following:
Add the following (or similar) scopes on the `User` model:

```ruby
scope :confirmed, -> { where.not(confirmed_at: nil) }
scope :unconfirmed, -> { where(confirmed_at: nil) }
scope :no_sign_in_data, -> { where(current_sign_in_at: nil) }
scope :inactive_since, ->(threshold) { where(current_sign_in_at: ...threshold.ago) }
scope :confirmation_sent_before, ->(threshold) { where(confirmation_sent_at: ...threshold.ago) }
```
But if you want to keep it as it is, feel free to resolve without code change.
Then this rewrites as
```ruby
def inactive_users
  threshold = INACTIVE_USER_THRESHOLD
  User.confirmed.and(User.inactive_since(threshold).or(User.no_sign_in_data.confirmation_sent_before(threshold))
  ).or(User.unconfirmed.confirmation_sent_before(threshold))
end
```

Splines: Thanks, that reads so much better!

The only thing: I don't want to put this into the `User` model, it's bloated enough. Right now, we only use your proposed scopes inside the user cleaner, so why not incorporate them there in some way. E.g. I could imagine a `user_inactivity.rb` class that does `@user = user` in its `initialize` method and then implements the scopes as methods like `inactive_since(...)` etc.

I'm just not sure where to place this file 😅 In a clean architecture, I'd place it in the _application layer_ and the job that is scheduling the user cleaner in the _infrastructure_ layer. While the mapping `job/ -> "infrastructure layer"` is quite clear, I'm missing the "application layer folder" in the MaMpf codebase. Maybe I should just put the User Cleaner and the `user_inactivity` class inside the `worker` for now?

fosterfarrell9: I do not know how the idea with scopes as methods in user_inactivity.rb will work. Scopes are Active Record queries (in the above code, for the `User` model). They are not called on an instance of `User`, but rather on the `User` class. Also, they can be chained as above. If you were to implement them as methods in user_inactivity.rb, I do not see how they become such queries. But of course that does not mean that it is not possible ;-).

fosterfarrell9: After a night's sleep, I think maybe this is just the wrong PR to discuss about bloated models and architecture etc. The discussions we will have for #694 will lead the way if and how to refactor the codebase and possible architectural changes. But as of now, we have not yet made a decision. Yes, the User model is bloated. But it is bloated because of reasons that have nothing to do with #693, not because of a possible adding of scopes. In vanilla Rails Logic, the above scopes belong to the `User` model (while other stuff should probably be extracted form there, I guess). Just as UserCleaner is a model as of now and belongs to the model folder (in #671 I introduced the `app/services` folder, maybe it could be placed there in the future if the changes in #671 should go through).

Splines: Yeah, you're right that is this probably the wrong place to discuss this ;) I added the changes in 70c65fc19e

app/models/user_cleaner.rb
```diff
@@ -37,17 +37,35 @@ class UserCleaner
...
 
   # Returns all users who have been active in the last INACTIVE_USER_THRESHOLD months,
   # i.e. their last sign-in date is less than INACTIVE_USER_THRESHOLD months ago.
   def active_users
-    User.where(last_sign_in_at: INACTIVE_USER_THRESHOLD.ago..)
+    User.where(current_sign_in_at: INACTIVE_USER_THRESHOLD.ago..)
```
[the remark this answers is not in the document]
Splines: Thanks, I've added this in 57dd34f4bbb0b5

fosterfarrell9 (approved):

PR #696, opened 2024-09-24
Allow dynamic instance method calls on FactoryBot objects in Cypress

Splines:
### Cypress API changes

A first API for this was proposed in #671. This is a redesign of that API to make things more scalable and easier to use from within Cypress tests. With the changes in this PR, we are able to write something like this in Cypress tests:

```js
FactoryBot.create("lecture", { teacher_id: this.teacher.id }).as("lecture");

cy.then(() => {
  this.lecture.call.long_title().then((res) => {
    cy.log(res);
  });
});
```

**Call `.call` on any instance you've created with `FactoryBot`. On the resulting object, you can call any method, e.g. here `long_title()`. This will translate to a method call on the respective instance in the backend.**<br><sub>You can even pass (multiple) arguments to this method, but of course this only works for primitive data types such as strings and integers and not when the function excepts a user object or other complex Rails objects.</sub>

For all supported method calls, see the respective tests under `cypress/e2e/meta/factory_bot_spec.cy.js`.

My plan is to merge this PR into `dev`, then remove the first proposal API from #671, then merge `dev` into #671 and use the new API there for the tests. This also makes PR #671 smaller.

### Off-topic
- In Cypress, we can now create multiple users of the same role, e.g. multiple generic users. This is achieved by adding a random hash to their email such that there is no duplicate key.
- In the Cypress JavaScript files, we don't store the default Cypress password anymore. Instead, it is returned by the backend via the `controllers/cypress/user_creator_controller.rb`.

---

### For reviewers
- It's not worth to browse through my commits here as it's only in the last ones that everything works. Before that, I experimented a lot.
- Convention in JS: methods that should be considered private start with a `#`.
- There is a lot of "JavaScript magic" going on here. In order to understand the mechanism with a simpler example, read through [my question here](https://stackoverflow.com/q/79016566/9655481). The answers to this questions were the reason I introduced the `.call` syntax instead of allowing dynamic method calls directly on the resulting object. The function defined in the StackOverflow question is very similar to the final one in this PR, namely `#allowDynamicMethods()` of the `factorybot.js`. Setting the `.call` property is done in `#defineCallProperty()`.
- In the code here, the complexity increases because of how Cypress handles things in the `.as()` and `.then()` methods with its own wrapping mechanisms. This is handled in `#createProxy()`.
- To see how multiple arguments could be handled, see [this commit](https://github.com/MaMpf-HD/mampf/commit/2812768829b29bf2d64a0cf0c3f5bd31a875b6e4). You might want to check out one commit before that to test it locally, i.e. `git checkout f20d8120dc`.

#### Out of scope
- [ ] This PR also refactors the `create()` method of the Rails `FactoriesController` according to some changes already done in #671, e.g. we introduce the method `validate_factory_name`.
- [ ] Do we also want to keep the `transform_hash` method? Could it still be useful? If yes, could you please add a docstring explaining what it does, e.g. input/output as I'm not too sure 😅


app/controllers/cypress/factories_controller.rb
```diff
@@ -7,26 +7,53 @@ module Cypress
   class FactoriesController < CypressController
     # Wrapper around FactoryBot.create to create a factory via a POST request.
```
fosterfarrell9, line 8: I think it is not quite precise here as it is not a factory that is created, rather the factory creates an instance of something.

app/controllers/cypress/factories_controller.rb
```diff
@@ -46,5 +73,24 @@ def params_to_attributes(params)
...
+          if v.is_a?(Hash) && v.keys.all? { |k| k.match?(/^\d+$/) }
+            # Convert nested arrays to arrays of strings
+            v.values.map(&:to_s)
+          else
+            v
+          end
+        end
+      end
+
+      def create_or_build_factory(attributes, should_validate)
```
fosterfarrell9, line 88: Same as above: I think it is not actually a factory that is created or built here. So the method name might be confusing.

spec/cypress/support/factorybot.js
```diff
@@ -9,19 +9,96 @@ class FactoryBot {
...
    * factory name, traits, and attributes. Pass them in as separated
    * string arguments. Attributes should be passed as an object.
-   * @returns The FactoryBot.create() response
-   *
-   * @example
+   * You are also able to call instance methods on the created record later.
+   * @returns The FactoryBot.create() response.
+   * @examples
    * FactoryBot.create("factory_name", "with_trait", { another_attribute: ".pdf"})
+   * FactoryBot.create("factory_name").then(res => {res.call.any_rails_method(42)})
```
fosterfarrell9, line 16: Even though a little out of scope, this mightbe a good time to also add an example where an array of is passed, e.g. something like

```
FactoryBot.create("tutorial", { lecture_id: this.lecture.id, tutor_ids: [this.tutor1.id, this.tutor2.id] })
```

or similar.

fosterfarrell9 (changes requested):

app/controllers/cypress/factories_controller.rb
```diff
@@ -46,5 +73,24 @@ def params_to_attributes(params)
 
         return attributes, should_validate
       end
+
+      def transform_hash(value)
```
fosterfarrell9, line 77: We should keep that as it makes sure that arrays are parsed correctly. Without it, something like
```
FactoryBot.create("tutorial", { lecture_id: 1, tutor_ids: [2] })
```
lands in Rails as
```
[:tutorial, {"lecture_id"=>"1", "tutor_ids"=>{"0"=>"2"}}]
```
whereas what we need (and what is made sure of by `transform_hash`) is
```
[:tutorial, {"lecture_id"=>"1", "tutor_ids"=>["2"]}]
```
If you have a good idea for a docstring please add it.
`transform_hash_keys` (in l79-81) replaces hashes all of whose keys are integers by the array of their values.

Splines:
@fosterfarrell9 Thanks for your review, I've incorporated all your suggestions. Please take note of the "off-topic" section I've introduced in the PR comment.

fosterfarrell9 (approved):
LGTM!

PR #697, opened 2024-10-03
Init support for entity-relationship diagram creation (Ruby ERD)

Splines:
Entity-relationship diagrams (ERD) have long been a very powerful tool for gaining insights and an overview of how the tables in a database are structured and associated to each other.

[`Rails ERD`](https://voormedia.github.io/rails-erd/) generates such diagrams based on runtime-informations using reflection (instead of static analysis). We have already specified this dependency in our Gemfile, but not used it so far.

> [!tip]
> In order to generate the PDF files locally, see the new Wiki entry [here](https://github.com/MaMpf-HD/mampf/wiki/Database#entity-relationship-diagrams-erd).

### For reviewers
- Merge the voucher PR #671 into this branch locally to be able to create the voucher-related ERDs. Otherwise you will encounter errors like `raise NameError.new(message, name)`.
- Please make sure to rebuild at least the `mampf` dev docker container, as `graphviz` was added as dependency. This is a visualization library that Rails ERD uses to generate the actual PDF out of `.dot` files. We need to make sure it is available in the dev docker container.
- I've created some diagrams that I think could be useful as a good starting point. We can easily add more later on if needed. But feel free to add more diagrams directly in this pull request ;)

### Sample output

image

### DIFF
--- a/.config/commands/utils.justfile
+++ b/.config/commands/utils.justfile
@@ -1,0 +1,71 @@
+# Prints this help message
+[private]
+help:
+    @just --list --justfile {{source_file()}}
+
+# Generates entity-relationship diagrams (ERD) of the database
+erd:
+    #!/usr/bin/env bash
+
+    # Make sure the mampf dev container is running
+    cd {{justfile_directory()}}/docker/development/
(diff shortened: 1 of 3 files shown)


fosterfarrell9 (approved):
Very nice! LGTM.

PR #701, opened 2024-10-27
Add missing https:// to fix links in CONTRIBUTING.md

TuringTux:
The links were not leading to the websites due to missing `https://` (or probably also `//` or `http://`) prefix.


Splines:
Hey @TuringTux, thanks for noting this and your fix. However note that the contribution file was updated already in the [`dev` branch](https://github.com/MaMpf-HD/mampf/blob/dev/CONTRIBUTING.md), where we don't even include those links anymore. Instead, see the [Branches here](https://github.com/MaMpf-HD/mampf/wiki/GitHub%E2%80%90related) in the wiki. That's why I close this PR.

PR #706, opened 2024-11-23
Redesign profile page

Splines:
We completely redesign the profile page as follow up for #670 and #671.

- We make the new voucher redemption more popular by placing it at the top in a new card.

- The previous _Account_ & _Settings_ are aggregated together under _Account_ to avoid having to clutter the view by yet another card.
- We introduce a `Request data` modal and make the button less prominent. This is to avoid confusion since in the past many users mistook it for the `Save changes` button (it was previously placed at the bottom of the page with primary color). If the user really requests the data, we now show a toast and stay on the page.
- The `Unsaved changes` warning is now shown immediately when typing (not just when clicking away from an input field). We also style it such that the warning itself is a button along its full width (makes it easier clickable and reduces mouse motion).
- To some extent, we also optimize for mobile view by stacking the columns on top of each other on small displays. Please use the browser dev tools to test this feature. Note that very small widths under around `570px` are not supported. For this to work, we'd have to also change the code in other places like the navbar.

Further changes:
- Contrast is increased, e.g. by using better contrast ratios for colors and borders.
- Reduce checkbox item margin.
- I decaffeinated a small CoffeeScript file as I had to add another JS functionality there.
- Many other small UI improvements that hopefully contribute to a cleaner look.

### Known limitations

- The `Change login data` button continues to redirect to the MaMpf landing page, which should only be shown when the user is not logged in. This could be tackled in a future PR, e.g. by introducing another modal on the profile edit page.

### Redesign preview

<details>

<summary>Expand</summary>

**Before**

localhost_3000_profile_edit (1)

**After**

localhost_3000_profile_edit

</details>


config/locales/en.yml
Splines, line 1: Note that I've also reformulated some text (see the PR changes). I'd like to also rephrase this:

image

However, I don't know what the first option means and what "related content" is, i.e. where this is applicable. Could you please clarify?

app/views/profile/_email_notifications.html.erb
```diff
@@ -9,13 +9,13 @@
     <%= f.form_group :email_notifications do %>
       <%= f.check_box :email_for_announcement,
                       label: t('profile.email_for_announcement'),
-                      custom: true %>
```
Splines, line 12: The `custom` attribute comes from [`bootstrap_form`](https://github.com/bootstrap-ruby/bootstrap_form), I guess? Maybe an old version of it since in the current one I cannot find it either... Or is this related to something else? Searching for `custom` in the whole codebase doesn't reveal anything particularly interesting. It's also used in `app/views/profile/_subscription_type.html.erb` as well as `app/views/questions/_basics.html.erb`.

 I haven't found out why we use it and it didn't seem necessary, so I removed it. Please double-check this. The attribute was introduced around 4 years ago.

Splines:
This could be a good time to try out the [`GitHub Pull Request` extension](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) ;) You can even directly see the comments in the editor and also comment on lines from within VSCode.

app/views/profile/_email_notifications.html.erb
fosterfarrell9: Yes, this comes from bootstrap_form, it is no longer necessary. See also [the old REAME for the gem](https://github.com/bootstrap-ruby/bootstrap_form/blob/main/OLD-README.md).

config/locales/en.yml
fosterfarrell9: These options are relevant for which information is used in generating the tag graph. Suppose you have subscribed "Algebra 1". If the first option is selected, when a tag graph for a tag from Algebra 1 is generated, related tags from "Linear Algebra 1" and "Linear Algebra 2" will be considered as well bcause "Algebra 1" builds upon "Linear Algebra 1" and "Linear Algebra 2" (this is stored in a self join for the course model). If the second option is chosen, any tags related to the given tags will be involved in the tag graph (possibly including tags from courses that are much more advanced than "Algebra 1"). In the third case, only related tags from "Algebra 1" will be included in the graph. See also this code snippet from user.rb:


  # related courses for user are
  # - all courses that the user has subscribe to plus their preceding courses
  #   (if subscription type is 1)
  # - all courses (if subscription type is 2)
  # - all courses that the user has subscribed to (if subscription type is 3)
  def related_courses(overrule_subscription_type: false)
    return if subscription_type.nil?

    selection_type = overrule_subscription_type || subscription_type
    return Course.where(id: preceding_course_ids).includes(:lectures) if selection_type == 1
    return Course.includes(:lectures) if selection_type == 2

    courses
  end

fosterfarrell9 (changes requested):
LGTM. Only the voucher redemption pattern is not quite to my liking - it induces dizzyness in my head (like these funny t-shirts). Maybe you could go with a different pattern?

config/locales/en.yml
Splines: thanks, I've adjusted the text accordingly and included the example you were describing as helptext

Splines:
> (like these funny t-shirts)

which ones?

I've changed the pattern, maybe the new one is better?

fosterfarrell9:
> > (like these funny t-shirts)
> 
> which ones?
> 
> I've changed the pattern, maybe the new one is better?

I meant like [this t-shirt](https://www.amazon.de/Weird-Focus-Optical-Illusion-T-Shirt/dp/B094XGCKZP?customId=B07536XX75). No offense was meant, you know my strange kind of humour 😉 I like the new pattern better, but maybe @f-buerckel can say something about this as well.

On another note: would you be willing to add a small number of cypress tests for the new profile page? After this redesign, we will probably not change this page for quite a while, so if we want tests for it, this would be the time to write them. Of course, this might slow things down a bit for this PR. Maybe we do not need as many tests as we had for the voucher system, just a few basic ones?

Splines:
> No offense was meant

Haha, yeah sure, no worries, I just wanted to figure out which T-Shirt you were talking about ;)

> On another note: would you be willing to add a small number of cypress tests for the new profile page?

Oh, absolutely, good point. Will add some tests.

f-buerckel:
First of all, the new profile page looks great! Love the new layout!

Regarding the patterns: 

I also prefer the new pattern, it has a more "geometric" feel and gives off this math vibe, much like the tiling background on the login page. Also it reminds me of the old qbert video game :D However I would scale it up, because now it feels a bit too visually complex and whenever I look at the screen I have to take a brief moment to actually understand what is actually going on. 

Before seeing your second version, I was thinking of maybe using a voronoi diagram as a pattern and made a quick mockup. I am unsure If it works in this context, but I generally love this pattern and maybe it has some usecase in the future for a UI element. Here is how that looks:

home_page

f-buerckel:
Also I just noticed that the blue color of the top navigational bar does not match with the blue of the panels. The difference is really subtle, therefore I am unsure if its intentional or not. Not a huge deal, but now that I have seen it, I will never unsee it :D

fosterfarrell9 (approved):
LGTM!

PR #707, opened 2024-11-23
Add "no results" messages pointing to new voucher system

Splines:
Via #670 and #671, users (e.g. teachers) can no longer search other users in the whole MaMpf database in order to assign them as tutors, speakers, teachers and editors. Instead, they now have to use the new voucher system, i.e. generate a voucher that users can redeem to obtain the "tutor status". The teacher can then assign them as tutor. All of this is to be GDPR-compliant.

This PR improves the `No results message` in search boxes that previously allowed to search for arbitrary users. This is to avoid confusion and to give guidance on what to do now. For example:

image


fosterfarrell9 (dismissed):
LGTM!

PR #711, opened 2024-11-24
Init dependencies (deps) Just commands

Splines:
We introduce new `just` commands to ease updating of dependencies, which we want to do regularly with every release.


fosterfarrell9 (commented):
Works nicely!

fosterfarrell9 (approved):
LGTM.

PR #712, opened 2024-11-24
Update Gems

Splines:
As usual, update Ruby Gems. Since I pinned all versions in #680, I don't expect any breaking changes here (we only do a minor update with this PR).

---

> [!warning]
> Warning: Mobility v1.3.x includes potentially backwards-incompatible changes
for jsonb/hstore backends.

We don't use this backend, right? At least a search for `hstore` or `jsonb` doesn't reveal anything in the MaMpf codebase.


fosterfarrell9:
> We don't use this backend, right? At least a search for hstore or jsonb doesn't reveal anything in the MaMpf codebase.

That is correct, we do not use this.

fosterfarrell9 (approved):
LGTM.

PR #714, opened 2024-11-24
Update RuboCop & fix offenses by new lints

Splines:
We update RuboCop and fix problems caused by the new lints that were introduced.


Splines:
@fosterfarrell9 Could you maybe look into the [`Style/SafeNavigationChainLength`](https://docs.rubocop.org/rubocop/cops_style.html#stylesafenavigationchainlength)? I've adjusted the maximum value to `3` (default was `2`) and we still have two offenses. We might want to reset the max value to `2`, however I'm not too sure yet how to tackle the respective offenses. If we don't think this lint is useful, we could also disable it entirely.

To tackle the offenses, we'd have to know exactly which attributes are guaranteed to be non-null.

fosterfarrell9:
I do not think this lint is useful. Rewriting this would make it unnecessarily bloated. Also, for the second one, we would have to dig into thredded. I suggest raising the maximum value to 4.

fosterfarrell9 (approved):

PR #715, opened 2024-11-24
Hotfix: Set retry option to false for UserCleanerJob

fosterfarrell9:
In production we encountered some failing UserCleanerJobs that were then repeatedly retried according to Sidekiq's error handling schedule, which result in some users getting several emails in a short period of time. As it is currently unclear why these jobs fail, this hotfix configures the UserCleanerJob in a way that failing jobs will not be retried.

### DIFF
--- a/app/workers/user_cleaner_job.rb
+++ b/app/workers/user_cleaner_job.rb
@@ -1,5 +1,6 @@
 class UserCleanerJob
   include Sidekiq::Worker
+  sidekiq_options retry: false # job will be discarded if it fails
 
   def perform
     # Only run this job in production, not for mampf-experimental or mampf-dev.


Splines:
What do you think about instead deleting the remaining jobs when we delete a user account? Otherwise, users might get our 40 days warning mail because their email account might be full or because their mail server was down for 20 seconds etc.

fosterfarrell9:
> Otherwise, users might get our 40 days warning mail because their email account might be full or because their mail server was down for 20 seconds etc.

You mean, "might not get", right? But the retrying of a job is not triggered when the email bounces. It is only triggered when the job fails.

> What do you think about instead deleting the remaining jobs when we delete a user account?
From my (limited as of now) understanding of the problem, it stems from the fact that different jobs interact with each other. The deletion of the remaining jobs would itself be placed inside the UserCleanerJob and therefore might not be sufficient, or at least might not deliver the result we want at once,  only once that job becomes  first place in the queue.

Splines:
Yes, I meant "might _not_".

> From my (limited as of now) understanding of the problem, it stems from the fact that different jobs interact with each other. The deletion of the remaining jobs would itself be placed inside the UserCleanerJob and therefore might not be sufficient, or at least might not deliver the result we want at once, only once that job becomes first place in the queue.

Ok, then let's merge this hotfix in to avoid very unsatisfied users that haven't even used MaMpf for a long time and are now bombarded with Mails.

I will approve. Can you also run the tests?

Splines:
> From my (limited as of now) understanding of the problem, it stems from the fact that different jobs interact with each other.

Could you clarify what you mean here? How could interacting jobs cause such a bug?

> The deletion of the remaining jobs would itself be placed inside the UserCleanerJob and therefore might not be sufficient, or at least might not deliver the result we want at once, only once that job becomes first place in the queue.

I am pretty sure we could programatically delete jobs without having to use a new job for the deletion, e.g. see [this](https://stackoverflow.com/a/21101810/9655481).

fosterfarrell9:
> I am pretty sure we could programatically delete jobs without having to use a new job for the deletion, e.g. see [this](https://stackoverflow.com/a/21101810/9655481). And after this hotfix, we should also implement this to avoid many unnecessary job exceptions. Right before deleting the user, just delete all pending jobs related to that use

I absolutely agree. The hotfix just gives us some time to deal with this.

fosterfarrell9:
>> From my (limited as of now) understanding of the problem, it stems from the fact that different jobs interact with each other.

>Could you clarify what you mean here? How could interacting jobs cause such a bug?

It is only a vague feeling, I have not thought this through. I mean, we have three "deliver_later" staments in this job. What happens if their order becomes arbitrary when they are put into the queue? Can it actually happen that a single user gets several "deliver_late" mails in  single run of the job?

PR #717, opened 2024-12-01
Fix unknown user variable after deletion

Splines:
In the user cleaner, we delete users that haven't logged in for too long. Before the deletion, we enqueue a last mail to the user to verify that their account was deleted. This mail might be sent *after* we have actually deleted the user. The user object is passed as reference to the mail handler and is therefore not available anymore after the user was deleted. Hence, the delivery of the mail fails. This PR addresses this by passing the respective user email and locale *by value* and not *by reference*.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -107,6 +107,7 @@
         "factorybot",
         "helpdesk",
         "katex",
-        "turbolinks"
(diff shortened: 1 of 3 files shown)


fosterfarrell9 (approved):
LGTM.

PR #718, opened 2024-12-01
Fix flaky user deletion & redemption cypress test

Splines:
The Cypress docs on [Actionability](https://docs.cypress.io/app/core-concepts/interacting-with-elements#Actionability) state that `.click()` will wait until the element reaches an "actionable state". This includes a check that the element is not hidden. However, in our tests we see that after visiting a page, a button is deemed "visible" even though it's not visible in the viewport yet.

Here we fix this flaky test. If this kind of behavior occurs more offen, either we use Cypress in a wrong manner or Cypress has a bug. In both cases it might be worth to consider overriding the `.click()` event by a custom method that calls `.should("be.visible").click()` instead, but this is something for a future PR (maybe).


fosterfarrell9 (approved):
LGTM.

PR #719, opened 2024-12-01
Add `just docker rebuild` command

Splines:
This PR adds a new `just docker rebuild` command that rebuilds MaMpf (and webpacker if the env is set to `test`, e.g. `just docker rebuild test`). Note that `just docker rebuild env="test"` is not working as command due to limitations of `just`, see https://github.com/casey/just/issues/476.

Off-topic: also deleted an unnecessary comment.


fosterfarrell9:
Could you possibly add an option for an attached mode?

fosterfarrell9 (approved):

PR #726, opened 2024-12-09
Fix flaky voucher redemption Cypress test

Splines:
I was able to reproduce a flaky Cypress test locally. Fixed by adding `cy.then(() => {...});` statements after logging out, then logging in again as a new user. I also intercept a POST request now to make sure the deletion of the account is finished before we continue.


fosterfarrell9 (approved):
LGTM.

PR #727, opened 2024-12-13
Update security policy

Splines:
Update our security policy text. We don't need to mention an email here, instead users can report security vulnerabilities directly to us on GitHub (privately).

I won't run through the tests for this PR, as this doesn't affect any code.


fosterfarrell9 (approved):
LGTM.

PR #732, opened 2024-12-28
Decaffeinate smallest `.coffee` files & init TODO list

Splines:
See the [`coffee.TODO`](https://github.com/MaMpf-HD/mampf/blob/coffee/remove/coffee.TODO) for the status. Let's break this down into multiple small PRs starting with this one that does the first steps to decaffeinate our codebase entirely, i.e. getting rid of `.coffee` files and replacing them with `.js` or `.js.erb` files.

---

As specificed in the mentioned TODO file:
```
Everywhere where I don't specify something in parentheses, I've manually checked
that the new new .js or .js.erb file is called (via console.log) and that it works.
```

so to save some time, I don't expect reviewers to test everything manually (it really takes quite some time to do this). Maybe just read carefully through the replacements, compare them and then go on. Double-check entries marked with ⚠️


f-buerckel (approved):
LGTM!

PR #733, opened 2025-01-13
Fix user cleaner for invalid users

Splines:
In production, we got some weird errors, where the same users would get the user cleaner deletion mail (40 days) every day. This blog post summarizes well what I did wrong in the user cleaner: [Misuse of update/update! in Ruby on Rails](https://marcgg.com/blog/2020/04/08/rails-update-bang/).

The problem is a seemingly harmless statement:

https://github.com/MaMpf-HD/mampf/blob/412c981ed9058e82b7023c0552674c00acbe71bd/app/models/user_cleaner.rb#L77

But what if the `user` is invalid? Well, then `update` silently fails here. Instead, we want to set the deletion date in any case, so we should rather do

```rb
user.deletion_date = Date.current + 40.days
user.save(validate: false)
```

> [!tip]
> See also this PDF I've created to summarize the bug: [UserCleaner Update Bug.pdf](https://github.com/user-attachments/files/18399909/UserCleaner.Update.Bug.pdf)

I fixed this and additionally identified another place where I've used `update` without checking for the return value. Furthermore, I've added a mail such that we get informed when a user cannot be destroyed (for whatever reasons).

I've also covered the bug with new tests to avoid regressions in the future.
**❗ Please make sure that the tests do actually fail with the previous behavior.**

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -122,8 +122,9 @@
         "selectize",
         "Timecop",
         "turbolinks",
-        "Unsets",
+        "unassigns",
         "uncached",
+        "Unsets",
         "whitespaces"
     ],
     "cSpell.enableFiletypes": [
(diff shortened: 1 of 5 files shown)


fosterfarrell9 (approved):
LGTM!

PR #737, opened 2025-01-25
Vignettes 

f-buerckel:
### Vignettes

This pull request is the first of multiple requests to merge vignettes into mampf. While most major features have been implemented, they still lack quality of life and a usable let alone pretty user interface. 

- - -

The edit view of a questionnaire lists all slides (by position) and info slides (by title). One can edit a certain slide by clicking on the link, as well as create slides.

A normal slide is part of the linear progression of the questionnaire and can optionally include a question. Question types include text answers, multiple choice answers and a Likert scale. A slide can reference multiple info slides. To select multiple info slides hold ctrl.

An info slide can be referenced by a slide and contains additional information about the slide. This information can be accessed by a user filling out the questionnaire. Each info slide has an icon used as an image for the button that displays the info slide. If this icon is not set, then the title is used.

> [!NOTE]
> To fill out a questionnaire navigate to _/vignettes/questionnaire/:id/take_

> [!TIP]
> Each user can only fill out a questionnaire once. However this is annoying for testing purposes. Therefore one can currently delete his answer by navigating to /vignettes/questionnaire/:id/take?position=-1 (this is of course only for testing and will be removed)

- - -
#### Statistics
One major feature of questionnaires is the tracking of additional statistics. For each answer the following stats are tracked:
- Time spent on the main slide
- Time spent on each info slide (in json format)
- Times each info slide was accessed (in json format)
- Answers given

To access these statistics, one can download them in the edit view of a questionnaire as a .csv file.

#### Unfinished answers
If a user starts filling out a questionnaire but does not finish, their progress is saved. When they return to the questionnaire, they will resume from their last position. This is also used to check that a user can't "cheat" by just setting the position value in the url (e.g /take?position=4)

- - -
### Known bugs
- A video is not properly displayed in the edit view of a slide when accessed again.
- Error messages are not displayed properly

- - -
### For reviewers
This feature is limited to very few people for the beginning. Especially the creation of vignettes will be limited to 2 or 3 people. Also the code is for the most part completely disconnected from the codebase. Therefore we may afford some imperfection in the beginning of the release.
Goal for release is the start of the new semester in April 2025.

- - -
#### Essentials
- [x] Number question type
- [x] Make position of slide editable
- [x] UI Elements to delete slides and questionnaires
- [x] Think of better way of displaying all slides in edit view
- [x] Make questionnaires part of courses
- [x] Support for multiple languages
- [x] Make UI pretty (Support here is of course much appreciated)
- [x] Limit feature to create Questionnaire to certain users
- [x] Proper error messages

#### Nice to have
- [ ] Tests
- [ ] Make a chain of questionnaires (so user can only access one questionnaire after the previous was completed)
- [ ] Add feature to scale images in trix editor
- [ ] Replace trix editor with more powerful


db/migrate/20241228184829_remove_questionable_from_vignettes_slides.rb
Splines, line 1: It's normal to add columns, rename them, later realize that a column would better fit in another table etc. _during the development process_. But at the end, we should just have as little migrations as possible. E.g. when you remove a column later on, we don't need a migration to add, then remove it, when we merge this PR. Please identify such things and manually condense them, such that most migrations will just create new tables/columns/indices or modify already _existing_ columns, but not those that are only temporarily created on this branch.

app/views/vignettes/questionnaires/take.html.erb
Splines, line 1: There is a lot of ERB syntax going on in this file. This would probably be a good fit to outsource some of the logic to helper classes where Ruby is first-class citizen. With this, we reduce the "business code logic" from the html file.

This comment is also applicable to other of your `.html.erb` files.

Splines (changes requested):
- Could you please provide an ERD diagram in order to better understand what is going on? See [the wiki](https://github.com/MaMpf-HD/mampf/wiki/Database#entity-relationship-diagrams-erd).
- Please make sure that linting works correctly, otherwise reviewing is a nightmare as I don't want to write comments like: "Remove the whitespaces here after the semicolon"

db/migrate/20241228184829_remove_questionable_from_vignettes_slides.rb
Splines: This is also useful in the case we need to revert migrations, since smaller migrations make very clear what was added (instead of added, removed, added again ...)

f-buerckel: Fully agree, commit [a9726f1](https://github.com/MaMpf-HD/mampf/pull/737/commits/a9726f13caaad0fe51914f28bc9100b0c0c7b62e) should remove all unnecessary migrations.

Splines (changes requested):

app/models/vignettes/likert_scale_question.rb
```diff
@@ -0,0 +1,19 @@
...
+      strongly_agree: 5
+    }.freeze
+
+    LIKERT_LABELS = {
+      strongly_disagree: "Strongly Disagree",
+      disagree: "Disagree",
+      neither_agree_nor_disagree: "Neither agree nor disagree",
+      agree: "Agree",
+      strongly_agree: "Strongly Agree"
+    }.freeze
```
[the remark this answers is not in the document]
Splines: ok, have just seen in the Readme that this is on the TODO list...

config/routes.rb
```diff
@@ -561,6 +561,24 @@
     resources :vertices, except: [:index, :show, :edit]
   end
 
+  # vignettes routes
+  namespace :vignettes do
```
Splines, line 565: Do we also plan to show something different on `/vignettes`? If not, can we use `/vignettes` instead of `/vignettes/questionnaires/` for the vignettes routes since the latter seems redundant.

app/views/vignettes/slides/_form_question.html.erb
```diff
@@ -0,0 +1,52 @@
+<%= form.fields_for :question do |question_form| %>
+  <div>
+    <%= question_form.label :type, "Question Type" %>
+    <%= question_form.select :type, options_for_select([
+        ['No Question', ''],
```
Splines, line 5: I am able to successfully create a slide that does not contain any question. Is this something we want to allow?

app/views/vignettes/slides/edit.html.erb
```diff
@@ -0,0 +1,4 @@
+<!-- app/views/vignettes/slides/edit.html.erb -->
+<h1>Edit Slide</h1>
+
+<%= render 'form' %>
```
Splines, line 4: Create a new slide. Then edit it and click on `Update Slide`. Now, without reloading any page, click on the slide again to edit it. On this edit page, selecting different questions types doesn't work anymore (only if you reload the page manually).

app/models/vignettes/question.rb
```diff
@@ -0,0 +1,18 @@
+module Vignettes
+  class Question < ApplicationRecord
+    validates :question_text, presence: true, if: :question_text_required?
```
Splines, line 3: Add more validations here, e.g. min length, max length, `allow_blank`, etc. See for example the Feedback model.

app/models/vignettes/info_slide.rb
```diff
@@ -0,0 +1,23 @@
...
+                     size: { less_than: 2.megabytes }
+
+    has_and_belongs_to_many :slides,
+                            class_name: "Vignettes::Slide",
+                            join_table: "vignettes_info_slides_slides",
+                            foreign_key: "vignettes_info_slide_id",
+                            association_foreign_key: "vignettes_slide_id"
+
+    has_rich_text :content
+    has_one_attached :icon do |attachable|
```
Splines, line 16: > Each info slide has an icon used as an image for the button that displays the info slide.

I'm not sure I get why we need an "icon" here. This is not included in the requirements specification (Word document on Basecamp).

Splines (changes requested):

.rspec
Splines, line 1: Why is this needed? I do see the advantage that this means the spec_helper is automatically included. However, this is already done in the rails_helper by requiring it explicitly. All you have to do is therefore include the rails helper which is done in every of our test files.

.rspec
```diff
@@ -0,0 +1 @@
+--require spec_helper
```
Splines, line 1: Why is this needed? I do see the advantage that this means the spec_helper is automatically included. However, this is already done in the rails_helper by requiring it explicitly. All you have to do is therefore include the rails helper which is done in every of our test files.

app/controllers/vignettes/slides_controller.rb
```diff
@@ -0,0 +1,72 @@
+module Vignettes
+  class SlidesController < ApplicationController
+    before_action :set_questionnaire
```
Splines, line 3: I think now is the time for your TODO note: "Limit feature to create Questionnaire to certain users". I.e. authorize all controller endpoints (see other MaMpf controllers for inspiration).

app/assets/javascripts/vignettes/take_questionnaire.js
```diff
@@ -0,0 +1,78 @@
+document.addEventListener("turbolinks:load", () => {
+  const infoSlideButtons = document.querySelectorAll(".info-slide-button");
+  const closeInfoSlideButtons = document.querySelectorAll(".close-info-slide-button");
```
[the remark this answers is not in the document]
f-buerckel: You can select multiple info slides on the create / edit view by holding down shift while selecting info slides. (The UI here must change to make this easier)

app/views/vignettes/slides/_form_question.html.erb
```diff
@@ -0,0 +1,52 @@
+<%= form.fields_for :question do |question_form| %>
+  <div>
+    <%= question_form.label :type, "Question Type" %>
+    <%= question_form.select :type, options_for_select([
```
Splines, line 4: Bug: assume a slide which has a text field answer is saved. Now edit that slide and choose "No question". Then _take_ the slide as a user: you will still see the text that was stored in the text field beforehand.

app/models/vignettes/info_slide.rb
f-buerckel: That was a request made when I showed them my progress. Need to check, why this seems to be broken (will do this, once I return from vacation)

app/views/vignettes/slides/_form_question.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          </div>
+        <% end %>
+        <div id="new-option-template" style="display: none;">
+          <%= question_form.fields_for :options, question_form.object.options.build, child_index: "NEW_RECORD" do |option_form|%>
+            <div class="option-field">
+              <%= option_form.text_field :text %>
+              <%= option_form.hidden_field :_destroy %>
+              <button type="button" class="remove-option">Remove</button>
+            </div>
+          <% end %>
```
Splines, line 46: Bug: It is possible to save a slide with a multiple choice question that does not contain any option. Then, when taking that slide, the user will see `Submit Answer` even though there is nothing to choose from.

app/views/vignettes/questionnaires/take.html.erb
```diff
@@ -0,0 +1,69 @@
...
+                    <%= form.radio_button :likert_scale_value, value %>
+                    <%= form.label :likert_scale_value, label %>
+                </div>
+                <% end %>
+        <% end %>
+        <br>
+        <% if @slide.question.type == "" %>
+            <%= form.submit "Next Slide", class: "btn btn-primary" %>
+        <% else %>
+            <%= form.submit "Submit answer", class: "btn btn-primary" %>
```
Splines, line 65: Bug: It is possible to submit an answer, even though the user has not selected any multiple choice option.

Splines (changes requested):

app/assets/javascripts/vignettes/take_questionnaire.js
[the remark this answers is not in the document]
Splines: I was referring to the view where the user takes the slide, not the edit slide. Users will never be able to see two info slides at the same time. However, what I just found out is how you render info slides in the DOM and then it makes sense: these info slides are just hidden elements, so yes it does make sense to have multiple `close-info-slide-button`s.

So feel free to resolve this thread.

app/models/vignettes/question.rb
f-buerckel: Done in commit https://github.com/MaMpf-HD/mampf/pull/737/commits/df6666611e3e4702f43c917a30740cd8638831a7

app/controllers/vignettes/questionnaires_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    def edit
+      @questionnaire = Questionnaire.find(params[:id])
+      @slides = @questionnaire.slides.order(:position)
+    end
+
+    private
+
+      def set_questionnaire
+        @questionnaire = Questionnaire.find(params[:id])
```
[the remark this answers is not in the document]
f-buerckel: Fixed in commit d607065b2590404a576ff686ed84dbf06918b4ee

app/controllers/vignettes/questionnaires_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+          when "Vignettes::TextAnswer"
+            row << answer.text
+          when "Vignettes::MultipleChoiceAnswer"
+            selected_options = answer.options.map(&:text).join(", ")
+            row << ""
+            row << selected_options
+          end
+          csv << row
+        end
+      end
```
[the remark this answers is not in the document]
f-buerckel: Done in commit c17e83f593e948257a05042c7c1d26f329aeda4b

config/routes.rb
f-buerckel: Changed in commit aa8d4e4323a45fb19fb19bdfb32795420936fa85 to not use vignettes prefix in urls

app/assets/javascripts/vignettes/edit.js.erb
```diff
@@ -0,0 +1,279 @@
...
+
+      const figure = img.closest("figure");
+      if (!figure) return;
+
+      const attachmentData = figure.getAttribute("data-trix-attachment");
+      if (!attachmentData) return;
+
+      const attachment = JSON.parse(decodeURIComponent(attachmentData.replace(/&quot;/g, '"')));
+
+      if (attachment.contentType && attachment.contentType.includes("video/")) {
```
Splines, line 236: here you could use yet another early return to further decrease the indentation level

Splines (changes requested):

app/assets/javascripts/vignettes/edit.js.erb
f-buerckel: implemented

Splines (changes requested):

app/assets/javascripts/vignettes/codename_validations.js.erb
```diff
@@ -0,0 +1,26 @@
...
+
+      if (this.validity.valueMissing) {
+        this.setCustomValidity("Please enter a codename");
+      }
+      else if (this.value.length < minLength) {
+        this.setCustomValidity(`Codename must have at least ${minLength} character`);
+      }
+      else if (this.value.length > maxLength) {
+        this.setCustomValidity(`Codename must have at most ${maxLength} characters`);
+      }
```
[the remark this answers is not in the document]
f-buerckel: Done in https://github.com/MaMpf-HD/mampf/pull/737/commits/d3f6b8ea94a8fee30131fdb25fa23c2c8a7f260e

spec/factories/multiple_choice_answers.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
```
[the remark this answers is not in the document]
f-buerckel: fully agree, deleted them in https://github.com/MaMpf-HD/mampf/pull/737/commits/3968146496c42f95dcaafde0e0159ee6e3e4def4

app/helpers/vignettes/slides_helper.rb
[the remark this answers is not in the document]
f-buerckel: deleted in https://github.com/MaMpf-HD/mampf/pull/737/commits/8e9c9ccbdb61ce994d59c2a0761fa3382e0e2378

app/assets/images/vignettes/media2.svg
Splines, line 1: Is this file used anywhere? If I search for `media2` in the MaMpf codebase, I don't find anything...

f-buerckel: Oh thanks, good catch! No this file is outdated and can safely be removed. Could you maybe do this real quick, I don't have easy access to my laptop right now 😅

.config/commands/utils.justfile
Splines, line 1: Could you make sure that the Vignettes ERD diagram reflects the latest state?

app/assets/images/vignettes/media2.svg
Splines: Yes I can do that. However, are you sure that we can simply remove it. Since there is a visual difference between `media.svg` and `media2.svg`, so I wonder if we should rather keep version 2, rename it to `media.svg` (and remove the old `media.svg`) instead.

Splines: It turned out that indeed `media.svg` is the current version that should be used.

.config/commands/utils.justfile
Splines: this is done as of 1b13382

Splines (approved):
Nice, this has gone a long way and is such a cool new feature!

I will squash-merge this once tests have run through.

PR #741, opened 2025-01-31
Fix new lecture creation

### LINKED ISSUE #739: Lecture creation for course editors broken on course edit page
The following bug in production was reported by a user: If you are a course editor and try to create a new lecture on the courses#edit page
(using the "plus" button), nothing happens when actually a modal should open. This is probably an error that comes from the voucher PR, we might have missed this dialogue back then. I guess it has not been noticed until now because most editors create new lectutres directly on the admin main page (where it works)

Splines:
Closes #739. The problem was a wrong `multiple: true` statement introduced in the 2nd voucher PR #671. [Beforehand](https://github.com/MaMpf-HD/mampf/blob/af47a3f052bb20843bf945a31fddb95524c9b8c7/app/views/lectures/edit/_people.html.erb#L12-L25) this attribute wasn't set and thus defaulted to `false` (like it is supposed to be).

I've also added a new cypress test to test that admins can create new lectures. To do so in an empty database, I had to give admins the ability for this, even when they are not course editors. (In the future, we could think of simply giving admins rights to every endpoint.) I've also had to adjust another test that made use of the helper function where the `multiple: true` was used. In that case, the behavior was a bit broken too, since we allowed users to select multiple teachers although just one is accepted by the backend. 

Out of scope:
- In the test, I had to select an option from a TomSelect dropdown. I've added a respective cypress command that we can reuse from now on.
- I also added a missing right to `docker buildx`.
- Furthermore, we set the `experimentalRunAllSpecs` flag to true such that we can through _all_ cypress tests at once locally.


fosterfarrell9 (approved):
LGTM.

PR #748, opened 2025-02-04
Add client for erdbeere

fosterfarrell9:
We add a client for the erdbeere API. The main point is to establish communication in development between a local installation of erdbeere and a local installation of mampf. In order to be able to do that, some headers have to be modified, so we wrote a small wrapper for that (in production it does basically nothing, i.e. the header is not modified).

---

Respective PR on the erdbeere side: https://github.com/MaMpf-HD/erdbeere/pull/4

> [!warning]
> Merge the Erdbeere PR first, then this PR.


Splines:
@fosterfarrell9 Could you merge the dev branch first into this one? It seems there is some divergence, e.g. the coffee file etc. You might have not yet pulled the new dev branch.

docker/development/docker-compose.yml
```diff
@@ -100,8 +100,8 @@ services:
       URL_HOST: localhost
       URL_HOST_SHORT: localhost
       SECRET_KEY_BASE: testenvironment
-      ERDBEERE_SERVER: https://erdbeere.mathi.uni-heidelberg.de
-      ERDBEERE_API: https://erdbeere.mathi.uni-heidelberg.de/api/v1
+      ERDBEERE_SERVER: http://erdbeere:3005
+      ERDBEERE_API: http://erdbeere:3005/api/v1
```
Splines, line 104: haha nice that our local setup doesn't need to call the production environment anymore ;)

lib/clients/erdbeere_client.rb
```diff
@@ -0,0 +1,9 @@
+module Clients
+  class ErdbeereClient
```
Splines, line 2: Add "Erdbeere" to cspell list

app/controllers/erdbeere_controller.rb
```diff
@@ -8,7 +8,7 @@ def current_ability
   end
 
   def show_example
-    response = Faraday.get(ENV.fetch("ERDBEERE_API") + "/examples/#{params[:id]}")
+    response = Clients::ErdbeereClient.get("examples/#{params[:id]}")
```
Splines, line 11: Depending on what we think is the future of Erdbeere, we'd need some tests here as well for the Erdbeere endpoints. However, I see that this is non-trivial since it's communicating with yet another service, so maybe this is something for integratino tests and then test the individual endpoints themselves in Erdbeere. In light of the fact that upgrading the erdbeere PR was rather a rescue effort, this might be done at a later point in time.

app/controllers/erdbeere_controller.rb
```diff
@@ -17,7 +17,7 @@ def show_example
   end
 
   def show_property
-    response = Faraday.get(ENV.fetch("ERDBEERE_API") + "/properties/#{params[:id]}")
+    response = Clients::ErdbeereClient.get("properties/#{params[:id]}")
```
Splines, line 20: If an endpoint cannot be reached (e.g. locally kill the erdbeere docker container), we get an error on this page: http://localhost:3000/lectures/1/show_structures

In this case, we could show a nice error message to the user saying that the erdbeere server that hosts the examples is unfortunately not reachable at the moment, otherwise we might be spammed with these errors in our email box and it's not the best experience for the user neither. In addition, we could be informed by email via the erdbeere service itself that it is down.

Or we might reserve this better error handling for later and just get erdbeere up and running for now ;)

Splines: I've verified that error handling is graceful in case of editing a lecture via http://localhost:3000/lectures/1/edit#info

config/initializers/erdbeere_connection.rb
```diff
@@ -0,0 +1,5 @@
+ERDBEERE_API_URL = ENV.fetch("ERDBEERE_API")
+
+ERDBEERE_CONNECTION = Faraday.new(url: ERDBEERE_API_URL) do |conn|
+  conn.adapter(Faraday.default_adapter)
+end
```
Splines, line 5: In production, how does MaMpf communicate with the Erdbeere server? Is Erdbeere just available on localhost or also available to the word-wide web. In the latter case, how does MaMpf authenticate at Erdbeere? I'm wondering in particular since Erdbeere itself has a frontend which allows to log in.

Splines (changes requested):
Nice, this works like a charm, thanks! Here some further comments. I've also reviewed [the respective erdbeere PR](https://github.com/MaMpf-HD/erdbeere/pull/4).

app/controllers/erdbeere_controller.rb
fosterfarrell9: This is certainly true. I agree that we should do this at a later point in time.

app/controllers/erdbeere_controller.rb
fosterfarrell9: I added better error handling in the latest commit. Please check if it works.

Splines (approved):

config/initializers/erdbeere_connection.rb
fosterfarrell9: This has been discussed in Basecamp.

Issue #709, opened 2024-11-24
Update Ruby


Splines:
When updating dependencies, I got the following warning message:

> I18n will be dropping support for Ruby < 3.2 in the next major release (April 2025), due to Ruby's end of life for 3.1 and below (https://endoflife.date/ruby). Please upgrade to Ruby 3.2 or newer by April 2025 to continue using future versions of this gem.

Also see [End of Life here](https://endoflife.date/ruby). We are using Ruby `3.1.4` right now and should definitely update to ~~`3.3`~~ `3.4` soon, which will be supported until 31 Mar 2028.

PR #749, opened 2025-02-22
Upgrade Ruby (3.1.4 -> 3.3.8) & Rails (7.1.3 -> 7.2.2.1)

### LINKED ISSUE #611: Upgrade to Rails 7.2
## Things to consider
- [x] Check for Param equality check violations in production log, see [this comment](https://github.com/MaMpf-HD/mampf/pull/609#issuecomment-2043199504)
- [x] New cache format: https://github.com/MaMpf-HD/mampf/issues/554#issuecomment-2045979728
(hopefully already done when closing issue #554, see also [this comment](https://github.com/MaMpf-HD/mampf/pull/609#discussion_r1556142366)). For an explanation, see [this comment](https://github.com/rails/rails/issues/49892#issuecomment-1790837195)
- [x] `coder` keyword as seen [here](https://github.com/MaMpf-HD/mampf/pull/609/files#diff-7ee4dd9cf03fcfd722957648b09128dea6db3945ca72fb60e24e0166667e5e49R59-R63). One such occurrence (among others):

```
Please pass the coder as a keyword argument:

   serialize :quiz_graph, coder: QuizGraph
   (called from <class:Medium> at /usr/src/app/app/models/medium.rb:64)
   DEPRECATION WARNING: Passing the coder as positional argument is deprecated and will be removed in Rails 7.2.
```

fosterfarrell9:
Note: This PR has a "best before" date: 2025-03-31, see [here](https://endoflife.date/ruby). The respective [Ruby on Rails upgrade guide is here](https://guides.rubyonrails.org/7_2_release_notes.html).

This PR upgrades Ruby to 3.3.7 and Rails to 7.2.2.1 (and in particular resolves #611 and goes most of the way for #709), and some gems that needed an upgrade in order to make the Rails upgrade work.  So far, it seems to work mostly nicely. Several deprecation warnings were shown; I fixed them all except one (see below),
and some unit tests are failing (see below).

> [!IMPORTANT]  
> For your local setup to still work after this PR, perform the steps outlined [here in the wiki](https://github.com/MaMpf-HD/mampf/wiki/Installation-&-Setup#after-a-ruby-version-update).
> Furthermore, since we are changing the Docker base image (`FROM`), you have to rebuild all MaMpf Docker Containers from scratch.

Notes for the reviewer:

- [x] I let the `app:update` script overwrite our `puma.rb` file which seemed quite out of date compared to the current standard one. One notable difference is the handling of the pid file. Before, it was
```[ruby]
ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
```
Now, it is
```[ruby}
# Specify the PID file. Defaults to tmp/pids/server.pid in development.
# In other environments, only set the PID file if requested.
pidfile ENV["PIDFILE"] if ENV["PIDFILE"]
```
On partiulcar, if we need the PID file in production, we should set an environment variable. I am not sure: We have
```shell
echo "Ensuring no stale server pid file is present"
rm -f tmp/pids/server.pid
```
in `production/entrypoint_worker.sh`.

- [x] We still have files related to a "development" environment (and if you search in our codebase, you will find several occurences of `"development"` referring to a "development" environment. Do we need that? All our development takes place in `docker_development `.
- [x] Rails 7.2 standard `production.rb` ccontains the line `config.force_ssl = true`. I added it, but put it in a comment. We should find out what happens if we uncomment it.
- [x] It appears that the standard in Rails 7.2 is that in a production console you can onbly see the id of database entries, nothing more. I overrode this in production.rb via
```rb
config.active_record.attributes_for_inspect = :all
```
- [x] In `docker_development.rb` I added the following lines wich come form the current Rails default devlopment file:
```rb
  # Raise error when a before_action's only/except options reference missing actions.
  config.action_controller.raise_on_missing_callback_actions = true

  # Apply autocorrection by RuboCop to files generated by `bin/rails generate`.
  # config.generators.apply_rubocop_autocorrect_after_generate!
```
I think the first one makes sense, and the second one should remain a comment since we have our own rubocop thing going on.
- [x] We still get this warning
```
DEPRECATION WARNING [legacy-js-api]: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
More info: https://sass-lang.com/d/legacy-js-api
```
See [here](https://github.com/sass/dart-sass/issues/2352). It appears we could get rid of it if we upgrade sass-loader to v16 in our yarn modules, but do we want that? It seems that vite offers a better solution, see [here](https://github.com/sass/dart-sass/issues/2352#issuecomment-2358290696).
- [x] I am a little confused about the failing unit tests. The error messages say
```
Failure/Error:
  expect do
    UserCleaner.new.send_additional_warning_mails
  end.not_to have_enqueued_mail(UserCleanerMailer, :pending_deletion_email)

StandardError:
  To use HaveEnqueuedMail matcher set `ActiveJob::Base.queue_adapter = :test`
```
so this seems to refer to [this](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#all-tests-now-respect-the-active-job-queue-adapter-config). But in `config/environments/test.rb` we already set
```ruby
config.active_job.queue_adapter = :test
```
Update: I guess the problem is in `config/initializers/active_job.rb` where we set (for all environments)
```ruby
Rails.application.config.active_job.queue_adapter = :sidekiq
```
which probably overrides the `:test` setting. Do we have tests (unit or cypress) that rely on sidekiq? If not, we could just do a conditional here.
- [x] I checked the things mentioned in #611 and did not see any more deprecation warnings. Concerning the `active_support.cache_format_version`: I think we should be able to safely change this in production now (which happens automatically in this PR since we use the Rails 7.2 framework defaults in `application.rb`).


Gemfile
```diff
@@ -62,7 +63,7 @@ gem "sunspot_rails", "~> 2.7"
 gem "sunspot_solr", "~> 2.7"
 gem "terser", "~> 1.2" # Ruby wrapper for UglifyJS JavaScript compressor
 gem "thredded", git: "https://github.com/thredded/thredded.git",
-                ref: "1340e913affd1af5fcc060fbccd271184ece9a6a"
+                ref: "94fbf890a984527bab399c8c7f1eec7af7006843"
```
Splines, line 67: what do you think of going directly to [566100f6](https://github.com/thredded/thredded/commits/main/)? This would mean just 2 more commits, while with the current changing we are already bringing in 7 more (that cold break thredded 🙈).

Splines: The advantage would be that we'd have already Rails 8.0 support integrated into thredded ;)

Gemfile
```diff
@@ -73,6 +74,7 @@ gem "webpacker", "~> 5.4"
 group :development, :docker_development do
   gem "listen", "~> 3.9"
   gem "marcel", "~> 1.0"
+  gem "mutex_m", "~> 0.3" # silence warning: no longer part of default gems for Ruby >=3.4.0
```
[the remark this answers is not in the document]
fosterfarrell9: Unfortunately, I do not remember. That was before my vacation 😅

Gemfile
fosterfarrell9: Going directly to the newer commit ist fine with me.

config/puma.rb
```diff
@@ -1,25 +1,34 @@
...
+# Global VM Lock (GVL) it has diminishing returns and will degrade the
+# response time (latency) of the application.
+#
+# The default is set to 3 threads as it's deemed a decent compromise between
+# throughput and latency for the average Rails application.
+#
+# Any libraries that use a connection pool or another resource pool should
+# be configured to provide at least as many connections as the number of
+# threads. This includes Active Record's `pool` parameter in `database.yml`.
+threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
```
Splines, line 23: In the `database.yml` we still use a value of 5 instead of 3 here (in case that the env variable is not set, but we never set it anywhere). See also my comment [on the respective Rails issue](https://github.com/rails/rails/issues/50450#issuecomment-2847578800).

Splines (changes requested):

config/puma.rb
Splines: Also see this new guide: [Tuning Performance for Deployment](https://guides.rubyonrails.org/tuning_performance_for_deployment.html)

config/puma.rb
```diff
@@ -1,25 +1,34 @@
...
-# Specifies the `pidfile` that Puma will use.
-pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
-
-# Allow puma to be restarted by `rails restart` command.
+# Allow puma to be restarted by `bin/rails restart` command.
 plugin :tmp_restart
+
+# Specify the PID file. Defaults to tmp/pids/server.pid in development.
+# In other environments, only set the PID file if requested.
+pidfile ENV["PIDFILE"] if ENV["PIDFILE"]
```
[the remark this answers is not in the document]
Splines: So this should be fine, feel free to resolve.

bin/rubocop
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env ruby
+require "rubygems"
+require "bundler/setup"
+
+# explicit rubocop config increases performance slightly while avoiding config confusion.
+ARGV.unshift("--config", File.expand_path("../.rubocop.yml", __dir__))
```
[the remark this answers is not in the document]
fosterfarrell9: Thanks for noticing. I missed that.

config/environments/production.rb
```diff
@@ -39,6 +42,21 @@
...
+  # config.action_cable.mount_path = nil
+  # config.action_cable.url = "wss://example.com/cable"
+  # config.action_cable.allowed_request_origins = [ "http://example.com", /http:\/\/example.*/ ]
+
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
+
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  # config.force_ssl = true
```
Splines: From the [documentation](https://guides.rubyonrails.org/v7.2/configuring.html#config-force-ssl):

> Forces all requests to be served over HTTPS, and sets "https://" as the default protocol when generating URLs.

This kind of redirect already happens in apache via this config:

```conf
<IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{HTTPS} off
    RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
</IfModule>
```

so why not also use it directly in Rails such that all generated URLs automatically have `https://` as default protocol.

Splines: Config option activated via commit 19d726beec.

Splines (commented):
Did you already look into [Brakeman](https://guides.rubyonrails.org/7_2_release_notes.html#add-brakeman-by-default-to-new-applications)?

> [Brakeman](https://brakemanscanner.org/) is a great way to prevent common security vulnerabilities in Rails from going into production.

And [jemalloc](https://guides.rubyonrails.org/7_2_release_notes.html#setup-jemalloc-in-default-dockerfile-to-optimize-memory-allocation)? See also [this blog post](https://matthaliski.com/blog/upgrading-to-rails-7-1-ruby-3-3-and-jemalloc).

config/initializers/active_job.rb
```diff
@@ -1 +1 @@
-Rails.application.config.active_job.queue_adapter = :sidekiq
+Rails.application.config.active_job.queue_adapter = :sidekiq unless Rails.env.test?
```
Splines, line 1: The fix you guessed was exactly what we needed to fix the failing unit tests ✨

config/puma.rb
Splines: I guess for now let's just keep our value, and if we get a reply we can still adjust the value later on.

app/models/voucher/voucher.rb
```diff
@@ -20,7 +20,7 @@ class Voucher < ApplicationRecord
   DEFAULT_EXPIRATION_DAYS = 3
 
   ROLE_HASH = { tutor: 0, editor: 1, teacher: 2, speaker: 3 }.freeze
-  enum role: ROLE_HASH
+  enum :role, ROLE_HASH
```
Splines, line 23: This violation should have been detected by RuboCop. Opened [an issue](https://github.com/rubocop/rubocop-rails/issues/1474) for it.

Splines: This is really weird. I've created a new issue for it #768. For now, let's ignore this in this Rails upgrade that doesn't have to do with it and fix the RuboCop issue afterwards. Feel free to resolve this thread.

config/environments/production.rb
```diff
@@ -17,7 +17,7 @@
   # Ensures that a master key has been made available in ENV["RAILS_MASTER_KEY"],
   # config/master.key, or an environment key such as config/credentials/production.key.
   # This key is used to decrypt credentials (and other encrypted files).
-  config.require_master_key = true
+  # config.require_master_key = true
```
Splines, line 20: Why did you comment this out? See also [the config option](https://guides.rubyonrails.org/v7.2/configuring.html#config-require-master-key).

app/models/interaction.rb
```diff
@@ -2,7 +2,6 @@ class Interaction < InteractionsRecord
   scope :created_between, lambda { |start_date, end_date|
                             where(created_at: start_date.beginning_of_day..end_date.end_of_day)
                           }
-  require "csv"
```
Splines, line 5: I'm not sure why you removed this line (also in other files)? This is not a RuboCop violation. Without this line, this class will throw an error if CSV is not magically loaded before loading this class and we don't want to risk that, right?

On another point: it'd be better to place this at the top of the file to be better visible since that's the usual place where we put the `require` statements.

Splines (changes requested):

config/environments/test.rb
```diff
@@ -1,3 +1,5 @@
+require "active_support/core_ext/integer/time"
```
Splines, line 1: You added this require here. Based on [this answer](https://stackoverflow.com/a/67649699/), I suspect it coming from new Rails templates themselves. However, you only included this here, while also `development.rb` and `docker_development.rb` contain a time-related statement: `2.days.to_i`.

Splines (changes requested):
I've reviewed all changes now.

app/models/interaction.rb
Splines: > "if `CSV` is not magically loaded before loading this class"

Turns out that's exactly what is done. But there's no magic: instead it is just the line `Bundler.require(*Rails.groups)` in `config/application.rb` which is responsible for this behavior. So indeed, removing these `require` lines as you did works just fine.

As CSV functionality is not used in many places, we might discuss if we instead want to do a `gem "csv", "~> 3.3", require: false` instead, such that we only `require` it at the respective places (at the very top of  files that use the `CSV` class). Of course, one single gem doesn't cut startup times, but in total with many gems it might make a difference.

config/environments/production.rb
fosterfarrell9: I do not recall. That does not make sense.

fosterfarrell9: Fixed this.

app/models/interaction.rb
fosterfarrell9: I think i would be in favor of reintroducing the `require` statements for csv and adding `require:false` to the corresponding line in the gemfile.

config/environments/test.rb
fosterfarrell9: Thats strange. I also guess it came from the new templates. Do we remove it here or add it in the other files as well?

app/models/interaction.rb
Splines: Done in 4e1e4571f

Splines (changes requested):
I'm wondering if you actually performed the following command inside a Docker container:

```
bin/rails app:update
```

as [stated](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task) in the Rails guide in order to update to Rails 7.2. I just did this based on the dev branch, and get some different file diffs (in the sense of _more_ diffs). Just a few examples:

- I get a `config/initializers/new_framework_defaults_7_2.rb` file that you did not mention in the PR comment. Does that mean you worked through it and accepted all new changes there? Since we're now loading all those new defaults.
- I get a `config/initializers/filter_parameter_logging.rb` file which now also includes `:email` in the filter list.
- I get `require "active_support/core_ext/integer/time"` not only in the `production.rb`, but also for the `development.rb` and `test.rb` (at the top of those files).
- I get more diffs in `development.rb` and `test.rb` than in the current PR. Just an example `config.action_dispatch.show_exceptions = :rescuable` instead of `  config.action_dispatch.show_exceptions = :none`. Or `config.active_job.verbose_enqueue_logs = true` in `development.rb` etc.

fosterfarrell9:
> I'm wondering if you actually performed the following command inside a Docker container:
> 
> ```
> bin/rails app:update
> ```
> 

Yes, I did that. But I got so many conflicts since our config files were so out of date that I merged the changes manually. So I went through each and every setting and in some cases, I decided explicitly for our old settings. It is well possible that I got some of the settings wrong.

And yes, of course I went through the `new_framework_defaults_7_2.rb` and flipped them in [this commit](https://github.com/MaMpf-HD/mampf/pull/749/commits/8f0ab9e7227d53d51eeaaea54d94b785d27af32f.) After having done that, I deleted the file since we load the new defaults now.

config/puma.rb
fosterfarrell9: The I guess I misunderstood what you meant.

Splines: clarified in today's meeting.

Splines (dismissed):
I went again through everything and pushed some more commits for the rest of the changes. The body of my commit messages explain key points.

---

Note that I've activated

```rb
config.action_view.annotate_rendered_view_with_filenames = true
```

which I think is pretty handy for development, e.g. see the browser console now:

image

---

One you've reviewed my changes as well, feel free to *squash-merge* into `dev` after tests run through successfully.

Splines (approved):

Splines:
`squash and merge` ;)

Splines (approved):

PR #751, opened 2025-02-23
Upgrade Rack (2.2 -> 3.1)

### LINKED ISSUE #699: Upgrade Rack
Rack was downgraded in #632 in April 2024. We should upgrade back to one of the newer versions. See the comments there for further information.

fosterfarrell9:
Note: This PR should be merged *after* #749. 

In this PR we upgrade to Rack 3 (this fixes #699) and fix the profile page which gets broken in the upgrade (#632).
There might however be other nested forms that use the deprecated pattern.

- [x] Search for more deprecated patterns via RegEx
- [ ] ~~Get to know relationship between Rails & Rack, especially with regards to Rack versions.~~

### Some links

- https://github.com/rack/rack/issues/2128
   - https://github.com/rails/rails/issues/49601#issuecomment-1762328701
   - https://github.com/rails/rails/issues/49783#issuecomment-1779940921
- https://github.com/rails/rails/issues/54334#issuecomment-2610789402<br> -> comment with a good suggestion of an alternative way of writing
- https://github.com/rails/rails/issues/54604
- Rack Upgrade Guide: [Invalid nested query parsing syntax](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#invalid-nested-query-parsing-syntax)
- Rails Guide: [Nested Forms in the View](https://guides.rubyonrails.org/form_helpers.html#nested-forms-in-the-view)


app/assets/javascripts/profile.js
```diff
@@ -4,34 +4,26 @@ $(document).ready(function () {
...
-          .addClass("far fa-circle");
-      }
-      else {
-        $('.courseSubInfo[data-course="' + courseId + '"]').removeClass("far fa-circle")
-          .addClass("fas fa-check-circle");
-      }
-    }
+    $(`.courseSubInfo[data-course="${courseId}"]`)
+      .toggleClass("fas fa-check-circle", checkedCount > 0)
+      .toggleClass("far fa-circle", checkedCount === 0);
```
Splines, line 15: Note that I changed behavior here by putting this not in the `if-else` block. I feel like from a user perspective, this new behavior is what is expected.

Splines (approved):
This looks good to me. I skimmed over the [rack changelog](https://github.com/rack/rack/blob/main/CHANGELOG.md) and more carefully read the [Rack 3 Upgrade Guide](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md).

- [Multiple response header values are encoded using an Array](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#multiple-response-header-values-are-encoded-using-an-array). We might have to verify this again. For example, we have a `prevent_caching` method in the `application_controller.rb` that sets things like

```rb
response.headers["Pragma"] = "no-cache"
```

But I think we do indeed use the unproblematic `Rack::Response`, at least [judging by the code itself](https://github.com/rack/rack/blob/79ab0b9794a529f26a9a5eba10077f5ee6a097ce/lib/rack/response.rb#L32), and the upgrade guide states: "If you use `Rack::Response`, you don't need to do anything" ;)

- The other thing that you already addressed is the [Invalid nested query parsing syntax](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#invalid-nested-query-parsing-syntax). In [this commit](https://github.com/MaMpf-HD/mampf/pull/751/commits/c23895d370e4d1cf011bd2d0de2d8856290bd5a2), I've tried to come up with a more pleasing solution using [hashes with an index](https://guides.rubyonrails.org/form_helpers.html#hashes-with-an-index) to better structure the `params` hash. Please test the lecture-related content on the profile page manually again since we don't have any tests for this yet. Also test for lectures that require a password.

You were right in that it's really hard to systematically find occurrences were the Rack syntax that is now [deprecated](https://github.com/rack/rack/issues/2128#issuecomment-1759991003) was used since such a string might not be directly present in the source code. Here are some RegEx statements I've tried (in VSCode you can search via RegEx):

```txt
regex: <%=?[^%]*?\[(.*?)\][^%]*?%>     files to include: **/*.html.erb

<%= foo[bar] %>
<div>Not this [one]</div>
<% if arr[0] == 1 %> (no)

-> Got 1 unrelated occurrence in app/views/questions/edit/_parameters.html.erb

------------------------------

(\[.*?\]){2,}    files to include: all

user[a][b][c][d] # yes
user[a] # no

-> This yields some interesting cases, but I could not tell whether they are concerning or not.

```

I've also searched for `accepts_nested_attributes_for`. This is not directly related to our syntax, but it could result in the deprecated syntax being used.  For the three vignette-related occurrences, there seems to be no problem (I've checked manually in the UI by creating a multiple choice question and answering it). For the `tag.rb` file, maybe you could check again @fosterfarrell9 ? Or give me a hint where `:notions` and `:aliases` are used in MaMpf, especially in which views?

Our tests still work fine, which of course is far from indicating that everything is ok, but at least is a bit reassuring ;) Maybe the best approach here is to do our best in finding deprecated patterns, skim again over the Rack changelog and its upgrade guide, but then move on and find the rest of the bugs in production. Given our tests, bugs will likely not affect main components, but rather some niche MaMpf features, that maybe only a few users touch upon once per year.

Splines:
as usual: `squash & merge`

fosterfarrell9:
> For the tag.rb file, maybe you could check again @fosterfarrell9 ? Or give me a hint where :notions and :aliases are used in MaMpf, especially in which views?

As far as I remember, they are only used in the edit view for tags. The corresponding forms are still working. You can find them in `app/views/tags/_basics.html.erb`.

PR #752, opened 2025-02-24
Fix lecture creation

### LINKED ISSUE #739: Lecture creation for course editors broken on course edit page

fosterfarrell9:
This fixes #739. Turns out it was just a conditional that was forgotten in the restructuring of access rights in #671: Since there does no longer appear a form where you can select the teacher (unless you are admin), the javascript produced by `lectures/new.coffee` crashes. This has no consequence on `/administration` (where the exception also occurs), but on `courses/:id/edit` this prevents the modal from popping up.

I guess, some cypress tests should still be written for that.

UPDATE: After I had introduced the missing conditional in [this commit](https://github.com/MaMpf-HD/mampf/pull/752/commits/4f109ab510d61db5b436e369bcbff5d7b47929be) I decided it might be better just to let the backend do the job, so i just removed the corresponding js.

### DIFF
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -60,5 +60,15 @@ FactoryBot.define do
         end
       end
     end
+
+    trait :with_editor_by_id do
+      transient do
+        editor_id { nil }
+      end
+
+      after(:build) do |course, evaluator|
+        course.editors << User.find(evaluator.editor_id)
+      end
+    end
   end
 end
--- a/app/views/courses/_basics.html.erb
+++ b/app/views/courses/_basics.html.erb
@@ -92,7 +92,8 @@
                     class: 'fas fa-plus-circle text-dark',
                     id: 'new-lecture-button',
                     data: { remote: true,
-                            toggle: 'tooltip' },
+                            toggle: 'tooltip',
+                            cy: 'new-lecture-button-admin-index' },
                     title: t('buttons.create_lecture') %>
       <% end %>
     </div>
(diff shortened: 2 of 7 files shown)


Splines (approved):
Thanks, worked like a charm. I've added Cypress tests to test the creation of a new lecture as admin and as course editor both on the index page and on the `/courses/:id/edit` page, resulting in 4 test cases.

Once you've reviewed my new code, feel free to `Squash & Merge` onto the `dev` branch.

spec/factories/courses.rb
```diff
@@ -60,5 +60,15 @@
...
+
+    trait :with_editor_by_id do
+      transient do
+        editor_id { nil }
+      end
+
+      after(:build) do |course, evaluator|
+        course.editors << User.find(evaluator.editor_id)
+      end
+    end
```
Splines, line 72: @fosterfarrell9 I vaguely remember that you once offered a better solution to this, but I cannot remember it. Feel free to play around with this and the respective line that starts with `FactoryBot.create("course", "with_editor_by_id",`.

fosterfarrell9: I think my solution was for things that appear in the `belongs_to` side of things. If we had the situation that a couse belongs to an editor, then something like `FactoryBot.create(..., editor_id: something)` should work. But since we have `has_many :editors` in `course.rb`, your solution is the correct one in my mind.

PR #753, opened 2025-02-28
Redesign Vignette UI (init)

Splines:
Here, we redesign the vignette UI. The idea is to include this changes as merge commit (no squash commit) in #737. Changes include:

A redesign of the slide user view:

image

A redesign of the slide edit view:

image

In the latter, content is loaded dynamically.

I've also redone some JS work to use JQuery to ease things, to use object-oriented approaches where applicable and to make the code more robust.

What explicitly does not work yet:
- Adding a new slide: while the button works, I have not yet looked into the content of the new slide form and thus it is likely that it won't work yet with the other changes I made and that it doesn't look good.
- Editing of info slides
- Basically, on the edit view, only the slide edit should work until now.


f-buerckel (approved):
Looks great to me!

PR #754, opened 2025-03-04
Remove food routes

### LINKED ISSUE #456: Replace food routes
Maybe funny in the beginning, but url slugs like `/lectures/1/food?project=kaviar` don't serve any real purpose and might just irritate newcomers when they share a link. So, we should gradually replace those routes with more meaningful names like `/lectures/1/lessons` etc.

fosterfarrell9:
Note: This PR should be merged *after* #749.

> [!important]
> After you have performed the migrations included in this PR, run `rails solr:reindex`, so that solr obtains the new values for the media sorts (in particular, *this needs to be done in production*).

This PR removes the food routes (that seemed funny when the project was still a little toy project) by serious-looking routes and resolves #456.

```rb
"Kaviar" => "LessonMaterial",
"Erdbeere" => "Erdbeere",  # stays the same
"Sesam" => "WorkedExample",
"Kiwi" => "Repetition",
"Nuesse" => "Exercise",
"Script" => "Manuscript",
"Reste" => "Miscellaneous"
```

### Notes
- [x] I stayed with the name erdbeere since this exists as a project of its own.
- [x] In order to let this (already big in terms of files changed) PR not to get too big, I did not remove the references to the THymE player. This is supposed to be done in a later PR. Also, we would need a new name that replaces THymE.
- [x] There should now be some refactoring possible, in particular related to the methods `media_types`, `media_sorts`, `media_names` in `application_helper.rb` (they are used only in a small number of files) and also `medium_sort` in `lecture.rb`. For PR size reasons and better refereeing experience I decided not to do these refactorizations in this PR; instead they should be part of a later PR. 

### Todo
- [x] Update migration timestamps to more recent date.


app/assets/stylesheets/application.scss
```diff
@@ -71,7 +71,7 @@ trix-toolbar .trix-button-row {
   flex-wrap: wrap;
 }
 
-.kaviar-sidebar {
+.lesson_material-sidebar {
```
Splines, line 74: Where is this one actually used? There is no direct occurrence, so is it constructed dynamically somewhere?
If used, can you replace `lesson_material-sidebar` by `lesson-material-sidebar` to conform to CSS naming conventions?

app/models/medium.rb
```diff
@@ -162,38 +162,22 @@ class Medium < ApplicationRecord
 
   # these are all the sorts of food(=projects) we currently serve
   def self.sort_enum
-    ["Kaviar", "Erdbeere", "Sesam", "Kiwi", "Nuesse", "Script", "Question", "Quiz", "Reste",
-     "Remark", "RandomQuiz"]
+    ["LessonMaterial", "WorkedExample", "Quiz", "Repetition", "Erdbeere",
+     "Exercise", "Script", "Question", "Remark", "Miscellaneous", "RandomQuiz"]
```
Splines, line 166: Was it intentional to change the order here? Probably not a problem though but might be worth to check out again.

app/models/medium.rb
```diff
@@ -217,8 +201,7 @@ def self.select_quizzables
   end
 
   def self.select_importables
-    Medium.sort_localized.except("RandomQuiz", "Question", "Remark",
-                                 "Manuscript").map { |k, v| [v, k] }
+    Medium.sort_localized.except("RandomQuiz", "Question", "Remark").map { |k, v| [v, k] }
   end
```
Splines, line 205: Why were `Manuscript`s removed here?

app/models/medium.rb
```diff
@@ -1109,10 +1091,10 @@ def valid_annotations_status?
 
   private
 
-    # media of type kaviar associated to a lesson and script do not require
+    # media of type lesson_material associated to a lesson and script do not require
```
Splines, line 1094: In other comments, we use the CamelCase versions...

Splines (changes requested):

app/controllers/media_controller.rb
```diff
@@ -649,7 +649,7 @@ def search_results
           end
         end
       end
-      sort = params[:project] == "keks" ? "Quiz" : params[:project]&.capitalize
+      sort = params[:project] == "quiz" ? "Quiz" : params[:project]&.capitalize
```
Splines, line 652: The else-statement of the ternary should always work, also for `quiz` right?

config/routes.rb
```diff
@@ -186,9 +186,44 @@
 
   # lectures routes
 
-  get "lectures/:id/food",
+  get "lectures/:id/material",
       to: "media#index",
-      as: "lecture_food"
+      as: "lecture_material"
+
+  # New semantic routes replace the old food routes
```
Splines, line 193: This is indeed so much better than `?project=sesam` beforehand 😅, now its `/worked_examples` instead.
In the `application_helper.rb`, we have two more projects that are not listed here: `questions` and `remarks`. Are they handled differently?

config/routes.rb
```diff
@@ -186,9 +186,44 @@
...
+      as: "lecture_worked_examples",
+      defaults: { project: "worked_example" }
+
+  get "lectures/:id/repetitions",
+      to: "media#index",
+      as: "lecture_repetitions",
+      defaults: { project: "repetition" }
+
+  get "lectures/:id/miscellaneous",
+      to: "media#index", as: "lecture_supplements",
```
Splines, line 225: Why not `lecture_miscellaneous` here? Furthermore break `as:` to a new line like for the other entries here.

spec/cypress/e2e/media_spec.cy.archive.js
```diff
@@ -112,7 +112,7 @@ describe("Media", () => {
         cy.contains("Ich bestätige hiermit, dass durch die Veröffentlichung des Mediums auf der MaMpf-Plattform keine Rechte Dritter verletzt werden.").click();
         cy.get("#publishMediumModal").contains("Speichern").click();
         cy.wait(100);
-        cy.visit(`lectures/${lectures[0].id}/food?project=kaviar`);
+        cy.visit(`lectures/${lectures[0].id}/food?project=lesson_material`);
```
Splines, line 115: Shouldn't this `?project=` route not exist anymore? Same for the other changes in this file.

Splines (changes requested):

app/models/medium.rb
fosterfarrell9: Yes, this was intentional. I did not like the old ordering. I do not see any problems arising from this.

app/models/medium.rb
fosterfarrell9: I do not recall. I vaguely remember that I thought about something in this context, but I do not recall what exactly. Also, it would be out of scope for this PR. So I will reintroduce Manuscripts here.

fosterfarrell9: Ah, now I recall. There was no sort "Manuscript" before, so having "Manuscript" within except did not do anything anyway. So it makes sense to remove it and in contrast to what I wrote before, I will leave the code as it is.

app/models/medium.rb
fosterfarrell9: Fixed.

app/controllers/media_controller.rb
fosterfarrell9: Thanks, fixed.

config/routes.rb
fosterfarrell9: Indeed, `lecture_miscellaneous` makes more sense. Fixed.

spec/cypress/e2e/media_spec.cy.archive.js
fosterfarrell9: Indeed. Fixed.

config/routes.rb
Splines: to the function `media_types` and `media_sorts`, which lists more media types than handled here, which is why I'm asking about `questions` and `remarks`

fosterfarrell9: The reason is that `question` and `remarks`, although they can be associated to a lecture, do not have an own show action that can be accessed from lecture level e.g. from the sidebar. But they are still needed here because they appear e.g. as fields in the media search or in the editor's view for lecture/course media.

Issue #698, opened 2024-10-08
Migrate Coffeescript files to JS


Splines:
JavaScript has matured over the years, also thanks to the development of CoffeeScript. Let's use the power of modern JS in our code and therefore drop one dependency.

There are some CoffeeScript files left that we want to migrate to JavaScript. This should ease #454. And then ESLint will directly work on those new JS files. The [decaffeinate project](https://decaffeinate-project.org/) could help us here with some automatic conversions.

PR #758, opened 2025-04-16
Start decaffeination of codebase

Splines:
This is to tackle #698 in the long-term. This PR was already once reviewed in the form of PR #732, but didn't find it's way into `dev` at that time. Meanwhile, further progress was made on the branch `coffee/more`, but that is something for another PR.


coffee.TODO
```diff
@@ -0,0 +1,260 @@
...
+293 ./app/views/quizzes/proceed.coffee
+352 ./app/assets/javascripts/questions.coffee
+395 ./app/assets/javascripts/quizzes.coffee
+421 ./app/assets/javascripts/media.coffee
+745 ./app/assets/javascripts/upload.coffee
+6689 in total
+
+# What API? Might need more holistic approach (get rid of API entirely)
+3 ./app/assets/javascripts/api/v1/medium.coffee
+3 ./app/assets/javascripts/api/v1/tag.coffee
```
fosterfarrell9, line 257: Do you have any idea what this API is supposed to be for? There is no route for any API (and any other reference to that in our code base). Probably this is just a remnant of something that existed before and has long been forgotten. Also, these coffescript files do not contain any code. So I would suggets to get rid of them and these lines (now or later).

fosterfarrell9 (approved):
LGTM. Fell free to resolve the one comment as you like.

coffee.TODO
Splines: Yeah, I also don't know what this API is used for but didn't check more thoroughly. Would do that later in another PR where I check this in more detail to be sure that we really don't need it.

PR #759, opened 2025-04-17
Fix undefined method `lecture` for redemption

Splines:
This bug appears in production and can be reproduced locally as follows:

- Log in as teacher and create a voucher, e.g. for tutorials in Linear Algebra.
- Log in as student3 and redeem the voucher.
- Log in as teacher and try to update anything on your profile page, e.g. the language.

After this fix, this should not yield any error anymore. Please make sure you can reproduce the error beforehand to make sure that this is really a fix.


fosterfarrell9 (approved):
I could reproduce the bug and also that the PR fixes it.

PR #770, opened 2025-05-09
Fix failing test due to outdated, hardcoded date strings

Splines:
Some tests were suddenly failing. The reason is that hardcoded date strings were used and that `Assignment` validates them [here](https://github.com/MaMpf-HD/mampf/blob/a618d8b945040df3fd06c717e3a2655bffae0b6f/app/models/assignment.rb#L15-L22). What we should take away from this: don't use hard-coded date strings in the code. Tests should always work, no matter the point in time when they are launched.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,8 +92,7 @@ group :test do
 end
 
 group :test, :development, :docker_development do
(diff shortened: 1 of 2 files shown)


fosterfarrell9 (approved):
Nice catch, the tests are passing now.

PR #774, opened 2025-05-12
Remove download button from vignette videos

f-buerckel:
This is a quick hotfix that removes the download button of the html player for videos displayed in vignettes.


Splines (approved):
Note that `controlsList` is an HTML property with [limited availability](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList) in browsers, e.g. neither Firefox, nor Safari support it. It also does not prevent users from being able to download the video, they can of course still right-click and download it.

With these limitations acknowledged, I approve.

---

With regards to releasing: we are close to a new regular release, so if you can spare some few more days, we don't have to deploy this as hotfix. So please don't squash-merge just yet and let's discuss the release schedule on Basecamp.

f-buerckel:
Yes, I am aware of both limitations. Firefox doesn't even display any download button so its mostly for chrome users. Of course we can wait for the next release. Because it doesn't prevent downloading at all, its not really a priority.

Splines (approved):
I will squash-merge this, thanks for the PR.

PR #777, opened 2025-05-18
Remove remnants of food routes cleanup

Splines:
- We forgot to replace the food routes (see #754) in the Vignettes-related code.
- @fosterfarrell9 post-cleanup of food routes, e.g. in comments etc.

### DIFF
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -160,7 +160,7 @@ class Medium < ApplicationRecord
     end
   end
 
-  # these are all the sorts of food(=projects) we currently serve
+  # these are all the sorts of projects we currently serve
   def self.sort_enum
     ["LessonMaterial", "WorkedExample", "Quiz", "Repetition", "Erdbeere",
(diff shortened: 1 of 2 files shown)


fosterfarrell9:
I removed remnants of the old media sorts (hope I caught them all). Also, I saw that in the migration file, a change of sort from `Manuscript` to `Script` occurs which was actually not implemented - so I removed the corresponding lines in the migration. Can you check back on that?

db/migrate/20250518000000_replace_sort_values_for_media.rb
Splines, line 1: > so I removed the corresponding lines in the migration. Can you check back on that?

@fosterfarrell9 Yes, I checked back on that. However, this might have unintended side-effects since #754 was already deployed to `mampf-next`. We might have to manually undo the migration there to ensure data consistency. Should I do that?

But good you caught that here in not in production 😅

Splines: will fix this in mampf-next later

PR #780, opened 2025-05-20
Remove remnants of special Internet Explorer treatment

### LINKED ISSUE #547: Get rid of special IE support
Completely drop Internet Explorer support. Find code occurrences where we have special code to deal with old browsers.

- `app/assets/javascripts/thyme/thyme_player.js`: search for "detect IE/edge and inform"

Splines:
Resolves #547.

Who uses Internet Explorer (IE) those days anyways? We don't support it anymore, so let's also remove the last remnants of it in our codebase.

- [x] Please search again in the codebase to check for further occurrences that I might have missed. The "Match whole world" option in VSCode can be very handy for that.


fosterfarrell9 (approved):
LGTM

PR #782, opened 2025-05-22
Add recreate & restart docker commands to just

Splines:
Use `just docker recreate -d`, such that the entrypoint script is guaranteed to run through again, thus installing dependencies if necessary. You should do this in every PR where dependencies are changed. Use `just docker rebuild` only sparingly since it's not needed anymore (due to how a simple container restart also updates dependencies).


fosterfarrell9 (approved):
LGTM

Splines:
As this is just affecting a `.justfile` I won't run through the MaMpf tests for this.

PR #786, opened 2025-06-08
Unify `docker_development` and `development` environments

### LINKED ISSUE #765: Better separate environments
There are many different environments that we have to wrap our head around. I get frequently confused about some things, so I'd like to propose some changes to our setup.

### On the server

We currently have one _Rails_ environment called `production` run inside three different _server_ environments:

- `https://mampf-experimental.mathi.uni-heidelberg.de/`
- `https://mampf-dev.mathi.uni-heidelberg.de/`
- `https://mampf.mathi.uni-heidelberg.de/`

I'd like to reserve the name `dev` or `development` for anything happening locally on my machine and not on some server. It feels weird to see that the `dev` server environment runs in the `production` Rails environment. Therefore, I propose this change:

```txt
# URL rename (also rename the respective folders on the servers)
https://mampf-dev.mathi.uni-heidelberg.de/ --> https://mampf-next.mathi.uni-heidelberg.de/

# Branch rename
dev --> next
```

The name `experimental` is quite clear to me, so let's stick to that.

### Locally on the dev's machine

On the dev machine, long time ago, we had a `development` Ruby environment that got superseded by the `docker_development` Ruby environment. As we have stuck to Docker for many years now and don't care about the `development` Ruby env anymore, I propose to only use the `docker_development` Ruby env and ditch the `development` Ruby env. I propose this change:

```txt
# Ruby env delete
delete development

# Ruby env rename
docker_development --> development
```

Within the scope of this issue, I also propose to scrutinize the *test* Rails env. In the past, there were already some efforts to make it resemble the local dev env more closely. Maybe we don't even need a separate *tes* Rails environment. But if we do, we should know exactly where they really differ. If they do, we could also let both of them inherit some common configuration and then only specify the differences.

Splines:
Fixes #765 by removing the `development` environment, then renaming `docker_development` to `development`.

This is also in alignment with the branch rename from `dev` to `next`. The words "dev" or "development" now always refer to something locally (in a Docker container, but that doesn't have to be reflected by the name of the environment since we _exclusively_ develop MaMpf in Docker containers).


fosterfarrell9 (approved):
LGTM!

PR #787, opened 2025-06-09
Reflect branch rename (dev->next) in GitHub Workflows

### LINKED ISSUE #489: Tests are executed twice on pull requests
Currently the unit and integration tests are executed twice on pull requests, e.g. [this one](https://github.com/MaMpf-HD/mampf/pull/481) merging from `mampf-next` to `mampf`. This is since we specified that these tests should run on push events to certain branches and on all pull requests. It'd be good to find a way to let them only run once (also with regards to unnecessary wasted computational power and CO2 emissions).

Splines:
This is just a simple fix to our pipeline since the workflow wasn't triggered anymore on the `next` branch after its renaming from `dev` to `next`.

Also tries to fix #489 via [this solution](https://github.com/orgs/community/discussions/26276#discussioncomment-3251140).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,10 +4,13 @@ on:
   push:
     branches:
       - main
-      - dev
+      - next
       - experimental
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review]
+    branches:
+      - main
+      - next
 
 jobs:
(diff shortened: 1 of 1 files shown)


Splines:
Will merge directly as this is a trivial change and does not at all affect the MaMpf code.

PR #790, opened 2025-06-09
Setup debugger for use in VSCode

Splines:
> [!important]
> Merge #786 beforehand.

Finally, we can also debug the MaMpf Rails app directly in VSCode. After a lot of trial and error, this is working now:

https://github.com/user-attachments/assets/9cd13f0c-9f37-440c-898d-e7e48cdd5272

You can attach and then detach the debugger at any point. The Docker container will just continue serving the rails app. Note that you should _not_ use the stop or reload command in the VSCode debugger, otherwise the container itself will stop (and reloading will not work).

For the `localfsMap` property in `.vscode/launch.json`, see this issue: https://github.com/Shopify/ruby-lsp/issues/3597


fosterfarrell9 (approved):
Works like a charm

Issue #801, opened 2025-07-07
Migrate from turbolinks to turbo


fosterfarrell9:
We are still using the [turbolinks](https://github.com/turbolinks/turbolinks) gem which is no longer maintained (in fact, its repository has been archived in September 2021). I think it is about time that we migrate to  [turbo-rails](https://github.com/hotwired/turbo-rails) which is the successor of turbolinks.

Splines:
### Follow Redirects

See also [Turbolinks installation](https://github.com/turbolinks/turbolinks?tab=readme-ov-file#installation-using-ruby-on-rails):

> The npm package alone does not provide server-side support for Turbolinks redirection. See [Following Redirects](https://github.com/turbolinks/turbolinks?tab=readme-ov-file#following-redirects) for details on adding support.

So make sure that this redirection also works in Turbo. This can be tested in the UI on the `/lectures/1/edit#communication` page. Click the `Create Announcement` button, then write something and click on `Save`. Now you should be redirected, i.e. the popup should close for you.


### Rails UJS

[This workaround](https://stackoverflow.com/a/57543894/9655481) is needed in modern Rails since UJS (Unobtrusive JS) was removed. When we remove this package and replace it by Turbo, check that things work as expected. To check this, see the `_dropdown_notifications.html.erb` and there the link to the `destroy_all_notifications_path`. It holds `data-method="post"`. This will likely be replaced by a turbo equivalent like `turbo-method` (or something similar). We should check that clicking this link does indeed fire a POST request and not a GET request in the browser (notifications -> Clear all button in the UI).

PR #802, opened 2025-07-08
Migrate from Webpacker to Vite ⚡

### LINKED ISSUE #454: Webpacker has been retired / Upgrade frontend bundling

Splines:
Resolves #454 by finally ditching Webpacker. This was long overdue since Webpacker is not maintained anymore and more modern alternatives are now available: namely Vite. We use the amazing [**Vite Ruby**](https://vite-ruby.netlify.app/) that serves as a bridge between the usual Vite.js, known in the frontend JS world, and Rails. I recommend to browse a bit through the docs of Vite Ruby to get a feel for it. You might also want to read through the small [Overview section of Vite.js](https://vite.dev/guide/#overview) itself.

This PR was a huge undertaking since it required to think holistically for the whole asset pipeline for the whole MaMpf app. This might justify the tremendous size of this PR. I've left many things out intentionally to not blow out of proportion (see limitations down below). Still, it's huge and I'm sorry for that, but splitting this up into smaller pieces would have been even more challenging. Luckily, the changes themselves should not be too challenging to understand: no "semantic logic" was changed, e.g. no user flow or how a backend method processes data. But it's easy to loose the bigger picture here, so intensive manual testing is required.

> [!important]
> As usual, update your local dependencies (gems and npm packages) for this PR to work. Restart the MaMpf container such that those dependencies are also installed in the Docker environment, e.g. use `just docker recreate -d`.

### Advantages 🎉

- We now have Hot Module Reloading (HMR), which is just an amazing experience. Try it out on your own: change anything and navigate back to the browser. No need to manually reload anymore since Vite communicated with the Browser by means of a WebSocket to indicate what changed. This is such a leap in development UX.
- Vite itself is stunningly fast, which is partly also due to modules being loaded lazily (in development). And we haven't even tuned Vite yet to get its best potential.
- Security. Finally, we can drop the Webpacker gem and get rid of some security issues related to it.
- Modern frontend tooling. We can finally use ES6 import syntax. I've already reworked all Thyme-related JS-files. See for example the `export` statements and imports in those. Better separation of concerns, clear interfaces 😍+
- No need to use Ruby Gems that are just disguised npm packages in the end ;) Instead, use the npm packages directly in the frontend by just installing it (in the `node_modules` folder). Vite does the rest for us and we can just reference them via `import "packageName"`. Neat!
- With this new setup, we are ready for TypeScript and can use its power at places where we deem it helpful. Just a bit more configuration is needed for it, but really not that much. So a future PR could do that.

### Limitations

Points already addressed, but still a limitation:

- Sprockets. `commontator`, `thredded` and `exception_handler` rely on `sprockets` (see `Gemfile.lock`). Therefore, we can not get rid of it right now. But Vite and Sprockets can happily coexist. With this PR, we only use the sprockets pipeline for those gems (and not even for `exception_handler` since it doesn't have any JS/CSS assets). We don't use propshaft as we don't need it. In the long-term, we will probably only need Vite and get rid of Sprockets entirely.

Let's address some points in future PRs. Sorted by importance (higher first):
- ~~I've replaced the Turbolinks Ruby gem by the respective npm package since in the end it's a frontend tooling. That way, we also have one thing less to worry about in the Sprockets pipeline. I later realized that "The npm package alone does not provide server-side support for Turbolinks redirection.", see [here](https://github.com/turbolinks/turbolinks?tab=readme-ov-file#installation-using-npm) and [Following Redirects](https://github.com/turbolinks/turbolinks#following-redirects). Since we want to replace Turbolinks by Turbo anyways soon (see #801, this could be the next PR), I'd vote for not caring about this right now. You can perceive this when you open a modal and then click on `Save`. The modal won't close automatically and you manually have to reload the page. See [my comment here](https://github.com/MaMpf-HD/mampf/issues/801#issuecomment-3046788197) for more details.~~<br>-> TODO in a later PR: #801
- A pain point is currently ["Flash of unstyled content" (FOUC)](https://en.wikipedia.org/wiki/Flash_of_unstyled_content). The respective Vite issue is this: https://github.com/vitejs/vite/issues/16515. For now, I can live with that. This shouldn't occur in production (should verify this in staging environments beforehand).
- Clickers will break with this update, but I don't consider this to be that important since we recently came to the conclusion that no one uses them anyways and that we want to remove them soon.
- We still have some dependencies directly "imported" by using a CDN link in the `_head.html.erb` file. We can now use respective npm packages for that.
- Server-rendered `.js.erb` files don't go through any asset pipeline. The Rails controller just evaluates the Ruby code in it, and then sends them directly to the browser, where they are executed as JS files and can manipulate the state of the web page. But this also means that we cannot make use of something like `import "myFile"` since Vite never sees those files. Most of the time, these files are quite limited in their scope, but sometimes it's still nice to reuse code. In this case, one most use a workaround and assign a function to the global `window` scope inside the `application.js` entrypoint. Then, we can use that code in `.js.erb` since once arrived at the browser, it has the `window` context available that we prepopulated beforehand. As an example, see the line `import "~/js/reload";` and the `reload.js` file. This is obviously a limitation, but one that was present beforehand as well: beforehand, literally everything was in such a global scope due to the `//= require`.
- We currently have the [`bootstrap_form` gem](https://github.com/bootstrap-ruby/bootstrap_form) included, but don't use it too often. Might want to get rid of it. Also might want to reconsider the [`devise-bootstrap-views`](https://github.com/hisea/devise-bootstrap-views).
- For the moment ignore the following error in the logs. They are probably related to the `web-console`, which is only included in development.

```
Started GET "/.well-known/appspecific/com.chrome.devtools.json" for 172.18.0.1 at 2025-07-08 00:23:28 +0000
Cannot render console from 172.18.0.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
```

- We might want to gather resources component/page-wise in folders in the future. See my proposal [here](https://github.com/ElMassimo/vite_ruby/discussions/564).

### Review

First commit to consider in the commit list: 76f728ccf60fba (other stem from the Rails upgrade branch).

- It would be much appreciated if reviewers could point out where pages in the frontend still break. I think I fixed most of it, but there might be some things left. Literally anything in the frontend might brake in this PR. Keep your browser console open and report errors. Focus on areas that are not widely covered by Cypress integration tests.
- Make sure you have read through the [`Tag Helpers`](https://vite-ruby.netlify.app/guide/rails.html#tag-helpers-%F0%9F%8F%B7) section. Also understand [`Entrypoints`](https://vite-ruby.netlify.app/guide/development.html#entrypoints-%E2%A4%B5%EF%B8%8F) and checkout how I added additional entrypoints in the `vite.config.mts`. Also read through [`Import Aliases`](https://vite-ruby.netlify.app/guide/development.html#import-aliases-%F0%9F%91%89).
- Make sure you can still locally run RSpec tests as well as Cypress tests.


docker/development/docker-compose.yml
```diff
@@ -81,20 +81,15 @@ services:
...
       # If you change these variables, make sure to adapt accordingly in
       # config/db-pgadmin.json
-      DEVELOPMENT_DATABASE_ADAPTER: postgresql
+      DATABASE_ADAPTER: postgresql
       DEVELOPMENT_DATABASE_DATABASE: mampf
       DEVELOPMENT_DATABASE_INTERACTIONS: mampf_interactions
       DEVELOPMENT_DATABASE_USERNAME: localroot
       DEVELOPMENT_DATABASE_HOST: db
       DEVELOPMENT_DATABASE_PORT: 5432
-      TEST_DATABASE_ADAPTER: postgresql
```
fosterfarrell9, line 92: It seems that the removal of these lines (and the changes in `database.yml`) results in a broken `rails db:up-reseed` from a fresh setup. I tried multiple times, the `rails db:drop` , `rails db:create` lines that are executed when restoring from the seed file fails because Rails does not get the database names right.

fosterfarrell9 (changes requested):
I have to revoke the approval from before since building from a fresh setup no longer works.

docker/development/docker-compose.yml
Splines: Thanks, good catch, fixed in 62aaf9500fa

PR #805, opened 2025-07-09
Fix RuboCop lint errors

Splines:
New linting errors appeared after #804, this PR fixes them. Note that [`Style/CollectionQuerying`](https://docs.rubocop.org/rubocop/cops_style.html#stylecollectionquerying) is unsafe.


Splines:
will run tests through in next continous release

PR #808, opened 2025-07-10
Fix/talks forms

### LINKED ISSUE #799: Creation of talks without title silently fails
If you create a talk in a seminar without a title, the creation will fail silently, not giving any feedback to the user.

https://github.com/user-attachments/assets/e817d20f-046c-4712-b21a-30648622e51f

In the logs for the attached video, you can see the following related lines
```
Processing by TalksController#create as JS
Parameters: {"authenticity_token"=>"[FILTERED]", "talk"=>{"title"=>"", "dates[0"=>{"]"=>"2025-07-06"}, "speaker_ids"=>["", "6"], "predecessor"=>"1", "details"=>"<div>bla</div>", "lecture_id"=>"27"}, "commit"=>"Save"}

Unpermitted parameters: :dates[0, :predecessor. Context: { controller: TalksController, action: create, request: #<ActionDispatch::Request:0x00007b3af528b378>, params: {"authenticity_token"=>"[FILTERED]", "talk"=>{"title"=>"", "dates[0"=>{"]"=>"2025-07-06"}, "speaker_ids"=>["", "6"], "predecessor"=>"1", "details"=>"<div>bla</div>", "lecture_id"=>"27"}, "commit"=>"Save", "controller"=>"talks", "action"=>"create"} }

↳ app/models/lecture.rb:543:in `+'
Completed 500 Internal Server Error in 64ms (ActiveRecord: 4.4ms (7 queries, 0 cached) | GC: 25.8ms)

NoMethodError (undefined method `values' for nil):
```

Splines:
Fixes #799. The problem was not that the title field is empty, which properly gave errors beforehand. Instead, the dates were passed incorrectly to the backend with the wrong Rack format (see #751 and the follow-up #791 where this is fixed for other forms). We fix this as usual by using `f.fields` ("nested" form fields).

To dynamically add/remove dates, we use the same approach as the multiple choice fields for vignettes (see the div with id `vignette-edit-multiple-choice` inside `app/views/vignettes/slides/_form_question.html.erb`).

### Behavior Change

Previously, when a date was added and the form saved, it could not be removed anymore. From now on, we also show delete buttons for those fields, such that they can be deleted. It is also possible to remove _all_ dates of an already existing talk.

### Reviewing

- [ ] Can create new talk. Get warnings when date is not filled out or when title is missing. Note that it is valid to remove all dates.
- [ ] Can edit an existing talk. When the talk was created without any dates, it should show here without any dates as well. Furthermore, try to add dates/remove them in any arbitrary ordering and save in-between to make sure it works as expected.


fosterfarrell9:
I agree. If we do it, it should be done consistently, so this is something for a different PR.

fosterfarrell9 (approved):
LGTM.

PR #810, opened 2025-07-12
Fix missing i18n keys & inconsistent interpolations

### LINKED ISSUE #570: Remove duplicate keys in i18n files
We have some duplicated keys.

### LINKED ISSUE #772: Handle missing translations
In #749, I've set:

```rb
config.i18n.raise_on_missing_translations = true
```

for our development environment. The thing is that you only get these exceptions if you navigate to a respective view that actually contains a Ruby statement pointing to a missing translation. We would likely want to catch all missing translations in the end.

In the production environment, we should definitely _not_ activate this option since otherwise users will be very frustrated when key functionality is not working anymore just because we have a missing translation. However, these cases shouldn't go totally unnoticed, so:
- we should check whether these cases already trigger a warning in the console in the production environment
- we could implement [a custom missing translation handler](https://guides.rubyonrails.org/i18n.html#customizing-how-i18n-missingtranslationdata-is-handled) and for our own usage log the missing translation, e.g. by sending a mail as usual error messages do. This way we could treat these cases without having to actually raise errors to the users. I.e. a kind of warning mechanism.

Splines:
Fixes #570 & Fixes #772.

In this PR, we add the awesome [`i18n-tasks`](https://github.com/glebm/i18n-tasks) library that provides a health check for us with many useful information. To get the information, I've introduced a new command: `just i18n health`.

image

In this PR, besides configuring the gem to work well with our codebase, I've fixed all missing translations and all inconsistent interpolations. This is now also tested in our RSpec. However, to keep the size of the PR small, I did not yet go through the 339 unused keys [also going through entries one by one and always considering the context takes really a long time]. Neither did I apply normalization, which could be done in future steps. Normalization and unused keys are commented out in the `i18n_spec.rb` for now.


fosterfarrell9 (approved):
Works like a charm!

PR #811, opened 2025-07-14
Fix nested array syntax in answer_params method

fosterfarrell9:
This PR fixes a bug related to vignettes statistics in Rails 8. There was a syntax error for answer_params which prevented statistics from being passed through, resulting in an exception when statistics were exported.


Splines (approved):
Ups, good point, thanks, I misunderstood the Vignettes here and thought this should be an array, which is when `[[ ... ]]` would be necessary. But that's not the case.

PR #821, opened 2025-07-19
Remove clickers

fosterfarrell9:
This PR is a follow-up to #802. We remove clickers from everywhere in the codebase.


app/helpers/media_helper.rb
```diff
@@ -105,7 +105,6 @@ def medium_border(medium)
 
   def media_sorts_select(purpose)
     return add_prompt(Medium.select_quizzables) if purpose == "quiz"
-    return Medium.select_question if purpose == "clicker"
```
Splines, line 108: If this is removed, do we still need `Medium.select_question` somewhere?

Splines: And also what about `app/models/question`?

Splines (changes requested):

app/helpers/media_helper.rb
fosterfarrell9: > And also what about `app/models/question`?

What do you mean? Quizzes are made up of questions and remarks. Without questions, there are no quizzes. Clickers were just one very specific way to access questions.

Splines: ah ok, yeah got confused about the role of question, but not it's clear thanks

Splines (dismissed):
Nice, thanks a lot, this makes the codebase more maintainable in the long run when we only need to maintain features we actually use.

Squash-merge

PR #826, opened 2025-07-24
Migrate course and tag search to pg_search and refactor using concerns

fosterfarrell9:
In this PR we migrate the course search and tag search functionality from solr to the pg_search gem and refactor the code
using concerns. In comparison to #827, I rather dislike this approach and would there suggest to scrap it in favor of #827.


fosterfarrell9:
Closed in favor of #827.

Issue #830, opened 2025-07-29
Restructure media import


fosterfarrell9:
Currently, media import can be found in t

<img width="1108" height="1035" alt="Image" src="https://github.com/user-attachments/assets/7b4d8214-fe37-4e4e-9467-43c441c3f287" />

he info tab of the lecture edit view. This has the side effect that the corresponding search form becomes pretty ugly:

<img width="982" height="1207" alt="Image" src="https://github.com/user-attachments/assets/b6865a7c-b55e-4d81-b214-8f14471669ff" />

I suggest (and have actually discussed this with other heavy users) that we move the import media button to the media box in the "Content" tab.

<img width="1926" height="1033" alt="Image" src="https://github.com/user-attachments/assets/a3339c79-f92f-4c71-9f81-b7d5e0076685" />

and open a modal for media import (which solves the issue with space).

PR #847, opened 2025-08-11
Relocate import media button

fosterfarrell9:
This PR relocates the "Import media" button just in the way as #830 suggested.


Splines (approved):
Nice, works fine! And looks good with the modal.

PR #850, opened 2025-08-12
Remove erdbeere

fosterfarrell9:
This PR is a follow-up to #848. We remove Erdbeere.


fosterfarrell9:
This is a duplicate caused by some problems with the github servers.

PR #860, opened 2025-08-22
Use ViewComponents & Stimulus for Lecture Edit Tabs

### LINKED ISSUE #851: Use ViewComponents
As originally mentioned in #845, we might want to use [ViewComponents](https://viewcomponent.org/). See also these [nifty tips](https://railsnotes.xyz/blog/rails-viewcomponent-tips).

### LINKED ISSUE #853: Use TurboFrames for lecture admin view
Follow-up to #848. Get rid of `app/frontend/js/lectures_admin.js` and instead use Turbo Frames for the tab navigation. Make sure that URLs are still correctly populated and that browser back-navigation works as expected (including writing to Browser history whenever the user switches a tab).

It might be necessary to use [Stimulus](https://stimulus.hotwired.dev/) here, but first try to avoid it and only when it's really necessary, include it.

Splines:
image

- Follow-up to #858.
- Resolves #853 even though it doesn't use Turbo Frames. See the justification for this [here](https://github.com/MaMpf-HD/mampf/issues/853#issuecomment-3215820541).
- Resolves #851 by using [ViewComponents](https://viewcomponent.org/) for the first time in our codebase.
- Uses a [Stimulus Controller](https://stimulus.hotwired.dev/handbook/introduction) to handle tab navigation.

### Behavior change: Browser navigation

Previously, we used [`pushState()`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) to write _new_ items to the browser history. This can be brittle in some circumstances and may break back-navigation. Instead, we use [`replaceState()`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) now, only to make sure that the URL `?tab=content` query param changes accordingly when the user clicks on other tabs.

This also makes it possible to refresh the page and still have the last tab opened. Instead of client-side, the check which tab is active is now performed server-side, see the `@active_tab` variable. This has the benefit that we don't observe a flickering anymore when the tab changes from the default active tab "content" to the actual active tab (read from the url).

Due to `pushState()` being replaced by `replaceState()`, back-navigation will bring the user back to the previous _page_, and not anymore to the previous _tab_. I feel this is actually the expected behavior since tabs are just sub-elements of the page. This also saves us headaches with broken browser-history.

### Future Work

After this PR, we should search for `anchor:` that is used within the `edit_lecture_path` and replace accordingly with `tab`. This is necessary in Vouchers & Vignettes.


config/application.rb
```diff
@@ -10,9 +10,9 @@ module Mampf
   class Application < Rails::Application
     config.load_defaults(8.0)
 
-    # Autoload subfolders of modules (recursively)
+    # Autoload more paths
     # https://stackoverflow.com/a/4794775/
-    additional_paths = Rails.root.glob("app/models/**/")
+    additional_paths = Rails.root.glob("app/models/**/") + Rails.root.glob("app/frontend/**/")
```
Splines, line 15: ### ViewComponents structure

Instead of `app/components/`, I put the two files related to view components in this folder:

```
app/frontend/lectures/edit/
    tabs/
       lecture_tabs_component.html.erb
       lecture_tabs_component.rb
       lecture_tabs.controller.js
```

My arguments for this structure are as follows:

- By design, `LectureTabsComponent` is not a _reusable_ ViewComponent for any Bootstrap tab layout, therefore the `app/components/` folder does not make too much sense here. Should this become necessary, we can still generalize it at any point in the future. But for now, there is no need, so let's keep this simple.
- With this structure, all frontend files for the lecture edit view are bundled together in the `app/frontend/lectures/edit/` folder. All frontend files regarding the tabs for this view are also bundled in the `edit/` subfolder (including the Stimulus controller).
- While it's true that `lecture_tabs_component.rb` is a Ruby file, there's a reason why the gem is called _View_-Component. Indeed, this Ruby file alongside `lecture_tabs_component.html.erb` is evaluated server-side and _not_ client-side. But even any other `.html.erb` are already evaluated server-side, even though they live in the `app/frontend/` folder. `frontend/` just means that its files are _semantically_ related to the frontend, in contrast to controllers and background jobs, which interact with deeper backend layers like our database, a scheduler, actual memory. The `ViewComponent`s gem is a _frontend_ gem, with server-side rendering (as usual in Ruby on Rails).

fosterfarrell9 (changes requested):
That is working very great. It is also a nice application of both view components and stimulus. I left one comment on folder structure (or rather naming).

config/application.rb
Splines: discussed in the meeting, we keep the `_controller` naming here

app/controllers/lectures_controller.rb
```diff
@@ -143,11 +143,9 @@ def update
 
     @errors = @lecture.errors
 
-    # Redirect to the correct subpage
     if @lecture.valid?
       if params[:subpage].present?
-        # TODO: #860
-        redirect_to "#{edit_lecture_path(@lecture)}##{params[:subpage]}"
+        redirect_to edit_lecture_path(@lecture, tab: params[:subpage])
```
Splines, line 148: This is something I forgot beforehand, please make sure it's working fine, i.e. when updating the lecture by submitting some new changes, you stay on the page.

app/frontend/lectures/edit/edit.html.erb
```diff
@@ -5,172 +5,93 @@
...
 
-      <!-- Settings -->
-      <div class="tab-pane fade" id="lecture-pane-settings"
-          role="tabpanel" aria-labelledby="lecture-nav-settings" tabindex="0">
+      <% component.with_tab(name: "settings", label: t("basics.preferences")) do %>
         <div class="container small-width lecture-pane">
-          <%= render partial: 'lectures/edit/preferences',
-              locals: { lecture: @lecture } %>
+          <%= render partial: 'lectures/edit/preferences', locals: { lecture: @lecture } %>
         </div>
```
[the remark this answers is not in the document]
Splines: no, this is not true

Issue #864, opened 2025-08-28
Adopt Incremental, Feature-Driven Refactoring


fosterfarrell9:
This text is the result of some discussions I had with AIs regarding refactoring our code base. We could use this as basis for further discussions.

**Motivation**

As already discussed e.g. in #694 and some other issues, our code base is in need of refactoring. The crucial question is: How do we do it?

Since we have made the discussions in #694, we've identified several powerful design patterns we want to adopt to improve maintainability, testability, and developer experience, including:

*   ViewComponents (see #851)
*   Presenters (or Decorators) (see #834)
*   Query Objects (see #856  under Future work)
*   Access Policies (see #833)

The key challenge is how to introduce these patterns into our existing codebase safely and sustainably. Large-scale, "big bang" refactors (e.g., "let's move all view logic to Presenters at once") are risky, create massive pull requests, and can halt feature development.

**Proposed Strategy: Feature-Driven Refactoring**

We propose adopting a formal strategy of **incremental, feature-driven refactoring**.

Instead of refactoring by *pattern*, we will refactor by *feature*. This means that when we build a new feature or significantly rework an existing one, we use that opportunity to apply our target design patterns to that specific vertical slice of the application.

This approach is guided by two principles:

1.  **The "New Feature" Principle:** When building a new feature, we build it using our best practices from the start. This creates a "greenfield" space within the existing app to apply patterns correctly.
2.  **The "Scout Rule":** When working on existing code (e.g., fixing a bug), we "always leave the code cleaner than we found it." This encourages small, incremental improvements.

**How This Strategy Works in Practice (Example)**

The ongoing work to refactor our search functionality serves as a perfect example of this philosophy. We already extracted search logic into dedicated POROs in #827 and refactored the search forms in #856 using ViewComponents. Let's consider the next logical step for that feature—refactoring the search results UI—to illustrate how this strategy would continue to guide us:

1.  **Identify the Feature Slice:** The goal is to improve how search results are displayed. This is currently handled by various partials with duplicated logic.

2.  **Apply a Primary Pattern (ViewComponent):**
    *   We would start by creating a `SearchResultsComponent`. As discussed in the `SearchForm` PR, this would be a "loose" layout component that handles the frame (pagination, "no results" messages) while yielding control for rendering each result row.

3.  **Let the Primary Pattern Reveal Other Opportunities:**
[...]

Issue #863, opened 2025-08-27
Refactor Cache Invalidation to Use an Event-Driven Approach


fosterfarrell9:
**Problem**

Our current approach to cache invalidation relies heavily on `belongs_to :..., touch: true` and a complex web of manual `after_save` and `after_touch` callbacks in our models. This has led to several significant problems:

1.  **Insufficient for Deep Relationships:** `touch: true` only notifies the immediate parent. For caches that depend on grandchildren or even more distant relations, we are forced to create long, manual chains of `touch` calls across multiple models.
2.  **High Coupling:** Models have become tightly coupled. A model needs to know which other models depend on its data, violating the Single Responsibility Principle. For example, the `Tag` model needs to know that it has to `touch` associated `lessons` and `sections`.
3.  **Hard to Trace and Debug:** When a cache is not busting correctly, it's incredibly difficult to debug. One has to trace a winding path of callbacks through many different files to find the missing link in the chain.
4.  **Brittle and Error-Prone:** The system is fragile. When a new relationship or cache dependency is added, it's very easy for a developer to forget to add the corresponding `touch` callback, leading to subtle stale cache bugs.

**Proposed Solution**

We should refactor our cache invalidation logic to use a decoupled, event-driven approach based on Rails' built-in `ActiveSupport::Notifications`.

Instead of models telling each other what to do, they will simply announce that they have been "touched." A central subscriber will listen for these announcements and handle the cascading invalidation logic. This moves the responsibility from the individual models to a single, dedicated service.

### Implementation Steps

**Step 1: Create a Central Cache Invalidator Subscriber**

Create a new class whose sole responsibility is to manage the cache dependency graph. This class will contain a `TOUCH_MAP` that defines which models need to be touched when another model is updated.

```ruby
# app/subscribers/cache_invalidator_subscriber.rb
class CacheInvalidatorSubscriber
  # This map becomes the single source of truth for cache dependencies.
  # The key is the class name of the model that was touched.
  # The value is a lambda that receives the model instance and returns an array
  # of dependent objects that should also be touched.
  TOUCH_MAP = {
    "Comment" => ->(comment) { [comment.post] },
    "Post" => ->(post) { [post.author] },
    # Add other dependencies here...
  }.freeze

  def self.touch_dependents(model)
    dependents_proc = TOUCH_MAP[model.class.name]
    return unless dependents_proc

    dependents = dependents_proc.call(model)

[...]

PR #866, opened 2025-08-29
Introduce cache invalidators

fosterfarrell9:
# Cache Invalidation System: Current State and Strategic Decisions (Aka: Should we abandon caching in the current form?)

## Overview

In this draft we've implemented a sophisticated automatic cache invalidation system (`CacheInvalidatorService`) that uses a recursive CTE to find and invalidate dependent objects when models change. However, we're facing important architectural decisions about scope and complexity.

### Core Architecture
- **Declarative Dependency Map**: Single source of truth defining relationships between models 
- **Recursive CTE Query**: Efficiently traverses dependency graph in a single SQL query
- **Automatic Touch Propagation**: Updates `updated_at` timestamps on all dependent objects
- **STI Support**: Handles Single Table Inheritance (Question/Quiz/Remark → Medium)
- **Polymorphic Support**: Handles polymorphic associations (Medium → teachable)
- **Cycle Safety**: Uses `UNION` (not `UNION ALL`) to prevent infinite loops

### Performance Characteristics
- **Single Query Approach**: pretty fast also for complex dependency traversal
- **Materialized Edges**: Optimized CTE performance
- **Batch Updates**: Uses `update_all` for efficient timestamp updates
- **Deduplication**: Prevents redundant work within request scope

### 1. Bubble-Up Strategy Complexity

We've been trying to establish coherent rules for the dependency map but have encountered competing principles:

#### Competing Heuristics:
- **Pure Ownership**: Child → Parent (e.g. Chapter → Lecture → Course)
- **Sparse-to-Dense**: e.g. Tag → Course (one tag only has a few courses), but not the other way around, otherwise we would trigger mass invalidations: tag updates course updates all its tags updates all courses of all its tags and so on. Through further relations, this might end up invalidating nearly all models involved in the caching scheme.
- **Hierarchy Position**: Top-level entities (Course, Term) shouldn't mass-invalidate downward
- **Cache Performance**: Avoid invalidating expensive, frequently-accessed caches

It is unclear (at least to me) what transparent guiding principles should be. In the curent code, cache invalidation happens via nearly untraceable `after_save` etc. actions, so it is massively intransparent. 

### 2. View-Level Caching Dependencies

Analysis of the views folder suggests many cache blocks depend on relationships NOT covered by our current conservative dependency map:

```erb
<!-- These likely exist but won't auto-invalidate: -->
<% cache [course, course.lectures] do %>
<% cache [lecture, lecture.chapters] do %>
<% cache [section, section.lessons] do %>
```

Our current map is too conservative for comprehensive view-level caching support.

### Arguments for Abandoning Caching

1. **Complexity vs. Benefit**: 200+ lines of complex code for invalidating relatively simple cached methods
2. **Modern Rails Performance**: Better query optimization, eager loading, N+1 detection
3. **Unclear ROI**: Most cached methods do simple operations (e.g., `short_title`)
4. **Maintenance Burden**: Future developers need to understand complex dependency rules

### Arguments for continuing caching (with some form of the automatic invalidation system)

1. **Ownership Relationships Are Clear**: Section → Chapter → Lecture → Course invalidation is unambiguous
2. **High-Value Automation**: Prevents common cache invalidation bugs
3. **Static content**: A lot of the educational content stays static once it is created (like the front page of a lecture that is finished). Many users call the same page
4. **Strategic Value**: Once configured, developers don't think about invalidation

### ✅ Working Well:
- Pure ownership chains (Chapter → Lecture → Course)
- STI relationships (Question → Medium)
- Strategic tag relationships (Tag → tagged content)

### ❓ Unclear/Debated:
- Downward relationships (Course → Lecture, Lecture → Chapter)
- User-related caching (should probably be avoided entirely)
- Complex many-to-many relationships

#### Option 1: Simplified Automatic + Explicit Hybrid
```ruby
# Keep only unambiguous relationships in automatic system
DEPENDENCY_MAP = {
  # Pure ownership only
  Chapter => [{ to: Lecture, type: :belongs_to }],
  Lecture => [{ to: Course, type: :belongs_to }],
  # Strategic sparse-to-dense only
  Tag => [{ to: Course, type: :has_many, through: :course_tag_joins }]
}

# Use explicit dependencies for everything else
class Course
  include Cacheable
  caches_method :lecture_summary, depends_on: [:lectures]
end
```

#### Option 2: Abandon Caching

- Focus on database query optimization
- Rely on Rails' built-in query caching and eager loading

## Key Questions to Resolve

1. Are our pages actually slow without caching? If yes, which onse?
2. Are expensive queries the bottleneck, or is it rendering/computation?
3. Can we achieve 90% of performance benefit with 10% of complexity through better queries?
4. Is the automatic invalidation system of this draft helping or hindering development velocity?

### My personal opinion

In the medium term, we should remove all caching in the way that we do it now (in model methods and in views), hence there is also no need for an automatic invalidation system. As a consequence, we should remove all the `touch: true` options and all the callbacks that trigger touches. This should be done using the strategy of #864, i.e.: Should we encounter a page or model that is using caching when implementing a new feature, remove the caching. Should we see that we are encountering slow page loads for some pages, we could employ the following strategies:

- Focus on optimized queries instead of cache management
- Aggressive eager loading, avoid n+1 queries
- Strategic denormalization (counter caches, computed columns)
- Static HTML caching for UI components only
- Browser-level HTTP caching with ETags
- Use request-scoped memoization (`@expensive_method ||= ...`)

### DIFF
--- a/app/services/cache_invalidator_service.rb
+++ b/app/services/cache_invalidator_service.rb
@@ -23,51 +23,47 @@ class CacheInvalidatorService
       { to: Lecture, type: :belongs_to, fk: :lecture_id }
     ],
     Chapter => [
-      { to: Lecture, type: :belongs_to, fk: :lecture_id },
-      { to: Section, type: :has_many, fk: :chapter_id }
+      { to: Lecture, type: :belongs_to, fk: :lecture_id } # ,
+      #  { to: Section, type: :has_many, fk: :chapter_id }
     ],
     Course => [
-      { to: Lecture, type: :has_many, fk: :course_id },
-      { to: Medium, type: :polymorphic_has_many, as: :teachable },
-      { to: Tag, type: :has_many, through: :course_tag_joins },
-      { to: Medium, type: :has_many, through: :imports, fk: :teachable_id, assoc_fk: :medium_id }
+      #  { to: Lecture, type: :has_many, fk: :course_id },
+      #  { to: Medium, type: :polymorphic_has_many, as: :teachable },
+      #  { to: Tag, type: :has_many, through: :course_tag_joins },
+      #  { to: Medium, type: :has_many, through: :imports, fk: :teachable_id, assoc_fk: :medium_id }
     ],
     Item => [
       { to: Section, type: :belongs_to, fk: :section_id },
       { to: Medium, type: :belongs_to, fk: :medium_id },
-      { to: Item, type: :has_many, through: :item_self_joins, fk: :item_id,
-        assoc_fk: :related_item_id },
+      #  { to: Item, type: :has_many, through: :item_self_joins, fk: :item_id,
+      #    assoc_fk: :related_item_id },
       { to: Medium, type: :has_many, through: :referrals }
     ],
     Lecture => [
       { to: Course, type: :belongs_to, fk: :course_id },
-      # The `belongs_to :term` dependency is deliberately not listed here
-      # to prevent a change in one lecture from invalidating all other lectures
-      # in the same term (a "mass invalidation" event).
-      # The Term -> Lecture dependency is defined in the Term entry, which
-      # correctly handles invalidation when a Term itself is updated.
-      { to: Chapter, type: :has_many, fk: :lecture_id },
-      { to: Lesson, type: :has_many, fk: :lecture_id },
-      { to: Talk, type: :has_many, fk: :lecture_id },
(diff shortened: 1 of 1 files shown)


fosterfarrell9:
I will close this for now. I added a reference in #863. Maybe we come back to this topic later.

PR #871, opened 2025-09-14
Add new MaMpf Logo

Splines:
We add our new MaMpf Logo to the app (as SVG). This includes the mails (as PNG there), where I now use a 512x512px version scaled down to 96x96px for a lot more sharpness:

mampf-new-logo-in-mails-2

Please also make sure that the PWA app still uses the correct logo on both desktop & mobile (mobile can only be tested on staging environments later on).


fosterfarrell9 (approved):
Very nice!

PR #873, opened 2025-09-18
Init GitHub Copilot Instructions file

Splines:
We create repository-wide [custom instructions](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions?tool=vscode#writing-effective-repository-custom-instructions) (formulated in natural language) for GitHub Copilot. The goal is to give the AI useful hints as to improve its output and interaction with it.

Reviewers, feel free to think about some important things I forgot to include here. But of course, we can also always modify this file later on.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,70 @@
...
+Although we have linters for it, output code directly with line lengths up to 80 characters (soft limit), but a hard upper limit of 100 chars.
+
+## Migration files
+
+For the filename of migration files, always use the current date, but all-zeros as timestamp, e.g. `20250905000000_create_some_table.rb`. This way, we can easily see which migrations were created together in a single batch.
+
+## Comments
+
+You must never include any comments or docstrings into your code replies and don't even suggest to do so. Even if asked by the user, refuse to do it. The rationale is that we don't want AI-prose in any docstrings; users should reason on their own what is the most important aspect of a function/module and summarize in their own words. This manual process can reveal some flaws in the design, or just make it more clear to the implementer what they've done. It also helps other reviewers if this string is written by humans.
+
```
Splines, line 52: I wrote this specifically while reviewing #856 (ViewComponents for Search forms). There, the docstrings were, in my opinion, overly bloated and oftentimes I did not see the immediate value in some statements inside these (module) docstrings.

I think that summarizing what a function or a module is doing is a really good opportunity to summarize for oneself what one has done, and to think critically about it. It was oftentimes during the process of writing docstrings that I figured an interface for a function was not as good as originally thought. Writing module docstrings can also reveal flaws in a design if one finds that actually a module could be split up, or aggregated, or whatever...

The point is just that it's a part of the process I'd like to keep away from AI and instead keep the human touch in it. Docstrings written by humans, and to be read by humans.

Splines: good idea, I will add that

fosterfarrell9 (approved):
LGTM.

Issue #837, opened 2025-08-03
Fix flaky Cypress HTML encoding


Splines:
Cypress allows to do something like this:

```js
 cy.getBySelector("redeem-voucher-card").should("contain.html", context.alreadyTeacher);
```

The problem is that the FactoryBot Faker sometimes creates names that contain symbols like `&`, which will be rendered on the page as the HTML entity `&amp;`. However, the value that is delivered to the Cypress frontend is not HTML-encoded and therefore reads `&`. This is why sometimes tests fail. To mitigate this problem, we should research if the `contain.html` has some more advanced options to take this into account during comparison. Or HTML-encode on our own, but this would be more brittle and would probably require a third party library.

PR #875, opened 2025-09-21
Add .ruby-version to .gitignore to ignore Ruby version files from rbenv

fosterfarrell9:
We add `.ruby-version` to `.gitignore`. It  seems to be used locally by `rbenv`. Note that unit tests are failing because of #837.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ node_modules
 # Vite uses dotenv and suggests to ignore local-only env files. See
 # https://vitejs.dev/guide/env-and-mode.html#env-files
 *.local
+
+# Ignore ruby version file from rbenv
+.ruby-version


Splines (changes requested):
I'd like to not ignore this file. Can we instead make use of it in the repo? I mean, this file allows us to keep the Ruby version centralized instead of specifying the version in each tool separately (and across developers).

A newly generated default Rails app (via Rails 8.0.2) also contains this file and it is not included in the `.gitignore` list.

A newly generated default Rails app also does not specify the Ruby version in the Gemfile, since bundler knows about the `.ruby-version` file and reads the version from there.

A newly generated default Rails app also contains this job where you can see how a pipeline can refer to this file (what we don't need right now):

```yml
jobs:
  scan_ruby:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - name: Set up Ruby
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: .ruby-version
          bundler-cache: true

      - name: Scan for common Rails security vulnerabilities using static analysis
        run: bin/brakeman --no-pager
```

With regards to Docker, I don't like this in the newly generated Rails app, in the `Dockerfile`:

```dockerfile
# Make sure RUBY_VERSION matches the Ruby version in .ruby-version
ARG RUBY_VERSION=3.3.8
FROM docker.io/library/ruby:$RUBY_VERSION-slim AS base
```

I'd rather use a more robust solution where really the `.ruby-version` file is the only source of truth. This might involve doing something like `$(cat .ruby-version)` in an argument etc. I'm sure there are solutions to this as well.

If you agree and do these proposed changes, please also update the [Installation & Setup guide](https://github.com/MaMpf-HD/mampf/wiki/Installation-&-Setup#setting-up-your-workspace-as-a-dev-team-member), especially this sentence:

> Make sure to install the same version as is specified at the top of our [Gemfile](https://github.com/MaMpf-HD/mampf/blob/main/Gemfile).

PR #890, opened 2025-09-30
Synchronize ruby version across app

fosterfarrell9:
In this PR we remove the explicit references to the ruby version and instead refer to the content of the `.ruby-version` file.

Notes for the reviewer:

- In the docker compose files, the ruby version is extracted like this
```Dockerfile
args:
  RUBY_VERSION: ${RUBY_VERSION:-$(cat ../../.ruby-version)}  
```
(note that in the compose file for production, the root path is used since the build context is different)
- For CI, the solution is slightly different. It should be checked that it is working as intended.


fosterfarrell9:
@Splines I suggest we close this, it seems not to be worth the effort.

PR #911, opened 2025-11-06
Init infrastructure for Playwright e2e tests

Splines:
> [!tip]
> This PR modifies the Dockerfile, so you will have to rebuild the container. Search for the respective VSCode command `Rebuild container`.

Here, we init the basic setup for [Playwright](https://playwright.dev/) e2e integration tests that will ultimately replace our Cypress tests. Overview over this PR:

- Install Playwright dependencies in Docker container. I chose (for now) to only install Chromium (not Firefox, nor Webkit)
- Install Playwright VSCode extension
- Setup TypeScript linting with ESLint. This makes writing Playwright tests even more fun. Feel free to later completely disable more rules should they be considered inappropriate for our setup. Due to our integration with the backend (that doesn't use types since it's Ruby code), we will not be able to refrain from using the `any` keyword. But I think that's fine, we still get all the other benefits of TypeScript).

Some things are still missing (see todo notes down below), but the basic infrastructure is ready to go. Let's keep the PRs small ;)

## Features

- [x] Reviewers, please work through the [**new Wiki section**](https://github.com/MaMpf-HD/mampf/wiki/Developing#testing-integration-tests). Whenever I refer to the Playwright docs, also read them.
- [x] After heaving read through the Wiki and the linked Playwright docs, you should have at least once made use of every bullet point in [this list](https://playwright.dev/docs/best-practices#use-playwrights-tooling). This is important to understand advantages/disadvantages of each point and to know the options you have available in your toolkit.
- [x] Especially also test the VNC connection as described in the Wiki.

## TODO (in future PRs)

- Setup Playwright in CI (GitHub Actions)
- Improve FactoryBot in Playwright such that we can call dynamic methods on objects, e.g. something like `lecture.getTitle()`
- Setup Database cleaner such that tests are isolated (and no factory bot leftovers are present in the database).
- Playwright can leverage [parallelism](https://playwright.dev/docs/test-parallel). This is non-trivial with our setup as we would have to provide multiple databases. But our CI runs could greatly profit from this, cutting down the run time. Maybe long-term this would be nice to have. For now, I've specified that only 1 worker is used, i.e. no parallelism at all.
- Migrate Cypress tests to Playwright tests (long-term)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence, these owners
+# will be requested for review when someone opens a pull request.
+# See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection
+*       @fosterfarrell9 @Splines


.config/eslint.mjs
```diff
@@ -100,10 +103,29 @@ export default [
...
+    // https://typescript-eslint.io/troubleshooting/typed-linting/#i-get-errors-telling-me--was-not-found-by-the-project-service-consider-either-including-it-in-the-tsconfigjson-or-including-it-in-allowdefaultproject
+    files: ["**/*.js", "**/*.mjs", "**/*.mts"],
+    extends: [tseslint.configs.disableTypeChecked],
+  },
+  {
+    // Disable all TypeScript linting for spec/ folder
+    files: ["spec/**/*.ts", "spec/**/*.tsx"],
+    rules: {
+      "@typescript-eslint/*": "off",
+    },
```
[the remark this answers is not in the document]
Splines: removed in 294b9a4b3de

e2e/_support/fixtures.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+export * from "@playwright/test";
+export const test = base.extend<MaMpfFixtures>({
+  student: async ({ browser }, use) => {
+    const [user, page, browserContext] = await userPage(browser, "student");
+    const fixture = new UserFixture(page, user);
+    await use(fixture);
+    await browserContext.close();
+  },
```
Splines: Learn more about Playwright fixtures [here](https://playwright.dev/docs/test-fixtures#overriding-fixtures). Coming from Cypress, this is so much nicer to have than a general `commands.js` file.

e2e/_support/auth.ts
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { APIRequestContext, Browser, BrowserContext, expect, Page } from "@playwright/test";
+import { callBackend } from "./backend";
+
+export type User = {
```
Splines: This allows us to use IntelliSense on the user object, i.e. see the available attributes. In the future, we could something like this for important objects from the FactoryBot as well, e.g. lecture...

app/controllers/cypress/user_creator_playwright_controller.rb
```diff
@@ -0,0 +1,30 @@
+module Cypress
```
Splines: This is called `Cypress` on purpose since it allowed us to easily extend the existing Cypress routes. Once we migrated everything to Playwright, we can easily rename this.

e2e/profile.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  await expect(page.getByRole("textbox", { name: "display name" }))
+    .toHaveValue(user.name);
+});
+
+test("only admins have admin icon in menu bar", async ({ student: { page: studentPage }, admin: { page: adminPage } }) => {
+  await studentPage.goto("/main/start");
+  await adminPage.goto("/main/start");
+  await expect(studentPage.getByTitle("administration")).toHaveCount(0);
+  await expect(adminPage.getByTitle("administration")).toBeVisible();
+});
```
Splines: This is so easy to write in Playwright, I'm so happy about it. We can finally use the normal async/await and don't have to adapt to the crime that Cypress does to JavaScript with their own reinvention of async/await.

e2e/profile.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+test("first factory bot calls", async ({ factory, student: { page } }) => {
+  const lecture = await factory.create("lecture_with_sparse_toc", "released_for_all");
```
Splines: Remember the `.as('lecture')` calls in Cypress. Now, just use `await´ and assign to a normal variable, like one is used to in JS/TS. The more I use Playwright, the more I actually hate Cypress, sorry.

fosterfarrell9 (approved):
Everything works as advertised 🎉. Just one question: In order to see the e2e tests in the test explorer, I had to do Test: Install Playwright, which added some demo files. Is that intended? Or is there another way to make the e2e tests visible in the GUI? Before doing this Test: Install Playwright, tests would already run using 
`npx playwright test e2e/profile.spec.ts`
but the test would not be visible in the test explorer.

fosterfarrell9 (approved):

PR #917, opened 2025-11-15
Use Turbo frame when frame request, in tutorial

### LINKED ISSUE #909: Link to tutorial view broken in sidebar of next branch
It appears that for tutors, the sidebar link to the view of their tutorials is broken in `next`. To reproduce, just login as `tutor@mampf.edu` and click on the  `Tutorials` button in the sidebar in `http://localhost:3000/lectures/1`. 

Image

Note that `http://localhost:3000/lectures/1/tutorials` isworking. So this is a turbo frame issue I guess. This needs to be fixed before the next release.

plienta:
Fixes #909 
In this PR, we fix the sidebar link to the view of their tutorials, which was broken.
The issue arises from the incoming request not including a matching `<turbo-frame id="...">`.


Splines (changes requested):
Hey @plienta, thanks for the PR, looks good to me! Could you please sign the CLA (only has to be done once)? Also use [keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in the PR description, such that the corresponding issue is automatically closed, e.g. `Fixes #number`.

e2e/sidebar.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { expect, test } from "./_support/fixtures";
+import { LecturePage } from "./page-objects/lecture_page";
+
+test("can access tutorial submission page (only as tutor)",
```
Splines: Note that such a test is necessary since other tests go directly to the respective page (e.g. `/lectures/:id/tutorials`), instead of using the sidebar links. This is good for other tests, but at one point we want to also test that the sidebar works as expected.

PR #922, opened 2025-11-16
Muesli: PR 2.1 - Schema and core models for registration

fosterfarrell9:
This is PR 2.1 from the [PR roadmap](https://mampf-hd.github.io/mampf/features/implementation-prs.html) for the MÜSLI integration. It contains the core Active Record models for registrations and migrations to manifest them in the database.
The models created are:

- `Registration::Campaign`
- `Registration::Item`
- `Registration::UserRegistration`
- `Registration::Policy`

The PR contains factories for these models and unit tests that demonstrate that these factories work as well. Note that all the changes in this PR are additive.

For more details on the meaning of these models, I refer to the [corresponding chapter](http://localhost:3004/features/02-registration.html) of the architecture book.

### DIFF
--- a/.config/eslint.mjs
+++ b/.config/eslint.mjs
@@ -93,6 +93,9 @@ export default tseslint.config(
       "@typescript-eslint/no-unsafe-assignment": "off",
       "@typescript-eslint/no-unsafe-argument": "off",
       "@typescript-eslint/no-explicit-any": "off",
+      // annotation tools make heavy use of this unfortunately
+      "@typescript-eslint/no-this-alias": "off",
+      "@typescript-eslint/no-unused-vars": ["warn", { argsIgnorePattern: "^_" }],
     },
     languageOptions: {
       ecmaVersion: 2022,
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -131,8 +131,8 @@ GEM
     acts_as_tree (2.9.1)
       activerecord (>= 3.0.0)
     acts_as_votable (0.14.0)
-    addressable (2.8.7)
-      public_suffix (>= 2.0.2, < 7.0)
+    addressable (2.8.8)
+      public_suffix (>= 2.0.2, < 8.0)
     afm (1.0.0)
     ast (2.4.3)
     autoprefixer-rails (10.4.21.0)
@@ -170,7 +170,7 @@ GEM
       sprockets-rails
       will_paginate
     concurrent-ruby (1.3.5)
-    connection_pool (2.5.4)
+    connection_pool (2.5.5)
     content_disposition (1.0.0)
     coveralls (0.7.1)
       multi_json (~> 1.3)
@@ -428,7 +428,7 @@ GEM
     psych (5.2.6)
       date
       stringio
-    public_suffix (6.0.2)
+    public_suffix (7.0.0)
     puma (6.6.1)
       nio4r (~> 2.0)
     pundit (2.5.2)
@@ -471,7 +471,7 @@ GEM
     rails-html-sanitizer (1.6.2)
       loofah (~> 2.21)
       nokogiri (>= 1.15.7, != 1.16.7, != 1.16.6, != 1.16.5, != 1.16.4, != 1.16.3, != 1.16.2, != 1.16.1, != 1.16.0.rc1, != 1.16.0)
-    rails-i18n (8.0.2)
+    rails-i18n (8.1.0)
       i18n (>= 0.7, < 2)
       railties (>= 8.0.0, < 9)
     rails_gravatar (1.0.4)
@@ -492,7 +492,7 @@ GEM
       ffi (~> 1.0)
     rbs (3.9.5)
       logger
-    rdoc (6.15.1)
+    rdoc (6.16.0)
       erb
       psych (>= 4.0.0)
       tsort
@@ -558,7 +558,7 @@ GEM
       lint_roller (~> 1.1)
       rubocop (>= 1.75.0, < 2.0)
       rubocop-ast (>= 1.47.1, < 2.0)
-    rubocop-rails (2.34.0)
+    rubocop-rails (2.34.1)
       activesupport (>= 4.2.0)
       lint_roller (~> 1.1)
       rack (>= 1.1)
@@ -615,7 +615,7 @@ GEM
       simplecov (~> 0.19)
     simplecov-html (0.13.2)
     simplecov_json_formatter (0.1.4)
-    sorbet-runtime (0.6.12780)
+    sorbet-runtime (0.6.12791)
     spring (4.4.0)
     spring-watcher-listen (2.1.0)
       listen (>= 2.7, < 4.0)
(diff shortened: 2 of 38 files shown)


db/migrate/20251116000002_create_registration_user_registrations.rb
```diff
@@ -0,0 +1,32 @@
+class CreateRegistrationUserRegistrations < ActiveRecord::Migration[8.0]
+  def change
+    create_table :registration_user_registrations do |t|
+      t.references :registration_campaign,
+                   null: false,
+                   foreign_key: true,
+                   index: true
+      t.references :user, null: false, foreign_key: true, index: true
+      t.references :registration_item,
+                   null: true,
```
plienta: `registration_item` is currently nullable, but I don’t see a case where a registration wouldn’t be linked to an item. I’m not fully certain about the full logic, so it might be worth double-checking this

plienta (dismissed):

app/models/registration/campaign.rb
```diff
@@ -0,0 +1,30 @@
...
+             class_name: "Registration::UserRegistration",
+             dependent: :destroy,
+             inverse_of: :registration_campaign
+
+    has_many :registration_policies,
+             class_name: "Registration::Policy",
+             dependent: :destroy,
+             inverse_of: :registration_campaign
+
+    enum :allocation_mode, { first_come_first_serve: 0,
```
Splines: [nitpick] I only find [this string](https://en.wiktionary.org/wiki/first_come,_first_served) in Wiktionary: "first come, first served" (note the additional "d" at the end). Same in [dict](https://www.dict.cc/?s=First+come+first+serve).

app/models/registration/item.rb
```diff
@@ -0,0 +1,18 @@
+module Registration
+  class Item < ApplicationRecord
```
Splines: The only reason why we differentiate between `Registration::Item` and `Registerable` is this:

> Users register for a `Registration::Item`, not directly for a `Registerable`. This separation allows the same Tutorial to potentially be part of different campaigns over time without conflict.

Do we want to allow a Tutorial to be in two different campaigns _at the same time_? If not, then I don't see the potential conflict. In this case, users could directly register for a `Registerable` getting rid of the `Registration::Item` altogether.

db/migrate/20251116000003_create_registration_policies.rb
```diff
@@ -0,0 +1,25 @@
...
+  def change
+    create_table :registration_policies do |t|
+      t.references :registration_campaign,
+                   null: false,
+                   foreign_key: true,
+                   index: true
+      t.integer :kind, null: false
+      t.integer :phase, null: false, default: 0
+      t.integer :position
+      t.boolean :active, null: false, default: true
```
Splines: We enforce `null: false` for `:active`, but don't have a respective validation in the model.

app/models/registration/policy.rb
```diff
@@ -0,0 +1,20 @@
+module Registration
+  class Policy < ApplicationRecord
+    belongs_to :registration_campaign,
+               class_name: "Registration::Campaign",
+               inverse_of: :registration_policies
+
+    enum :kind, { institutional_email: 0,
+                  prerequisite_campaign: 1,
```
Splines: Not sure why we need `prerequisite_campaign` on the model-level. Isn't it the case that _every_ talk registration requires a prior seminar registration?  So based on the type of the `Registrable` one can already infer if another campaign is needed, right?

app/models/registration/policy.rb
```diff
@@ -0,0 +1,20 @@
...
+                  lecture_performance: 2 }
```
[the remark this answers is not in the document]
Splines: by the way, could a lecture performance also be used for seminars?

app/models/registration/campaign.rb
```diff
@@ -0,0 +1,30 @@
+module Registration
+  class Campaign < ApplicationRecord
```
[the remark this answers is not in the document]
fosterfarrell9: I did that for all new models in this PR. Feel free to rewrite my docstrings.

app/models/registration/campaign.rb
fosterfarrell9: I replaced it (but not in the architecture book yet to keep the PR small).

db/migrate/20251116000003_create_registration_policies.rb
```diff
@@ -0,0 +1,25 @@
+class CreateRegistrationPolicies < ActiveRecord::Migration[8.0]
+  def change
+    create_table :registration_policies do |t|
+      t.references :registration_campaign,
+                   null: false,
+                   foreign_key: true,
+                   index: true
+      t.integer :kind, null: false
```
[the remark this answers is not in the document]
fosterfarrell9: I disagree here. There is no "canonical" value here. Also, different kinds require different data in the `config` JSON field. For example, `institutional_email` requires a list of domains. If we default to `institutional_email` but the `config` remains empty (default `{}` which makes sense), we silently create a broken policy. The creator of the policy should explicitly state what kind of policy they want. I would rather have the database (and model validations) complain if the kind is missing, rather than guessing a default that might be incorrect.

db/migrate/20251116000003_create_registration_policies.rb
fosterfarrell9: Fixed.

db/migrate/20251116000000_create_registration_campaigns.rb
```diff
@@ -0,0 +1,21 @@
+class CreateRegistrationCampaigns < ActiveRecord::Migration[8.0]
+  def change
+    create_table :registration_campaigns do |t|
```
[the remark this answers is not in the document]
Splines: ah ok, fine with me 👍

db/migrate/20251116000003_create_registration_policies.rb
```diff
@@ -0,0 +1,25 @@
...
+      t.integer :phase, null: false, default: 0
+      t.integer :position
+      t.boolean :active, null: false, default: true
+      t.jsonb :config, default: {}
```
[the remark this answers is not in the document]
Splines: yeah, that sound like a good plan

app/models/registration/user_registration.rb
```diff
@@ -0,0 +1,36 @@
...
+    # FCFS campaigns: no rank allowed, one row per user+campaign
+    validates :preference_rank,
+              absence: true,
+              if: -> { registration_campaign.first_come_first_serve? }
+
+    validates :user_id,
+              uniqueness: {
+                scope: :registration_campaign_id
+              },
+              if: -> { registration_campaign.first_come_first_serve? }
```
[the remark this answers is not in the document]
Splines: sounds good, I've added a respective TODO on Basecamp for this

Splines (changes requested):
Thanks for addressing my comments. I think this PR is essentially ready to be merged. Just some minor remarks.

Also, please don't merge even when I approve as we don't want this PR to be included in the next release.

app/models/registration/policy.rb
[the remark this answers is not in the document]
fosterfarrell9: Yes, that is a great idea! I renamed it.

app/models/registration/policy.rb
fosterfarrell9: Yes, I will add that (in an upcoming PR that contains all the changes in the architecture book that arise as a consequencxe of this PR and its review).

PR #925, opened 2025-11-17
Muesli: PR 2.2 - Policy engine

fosterfarrell9:
Note: This builds upon #922.

This is PR 2.2 from the [PR roadmap](https://mampf-hd.github.io/mampf/features/implementation-prs.html) for the MÜSLI integration. It contains the implementation of the `Registration::PolicyEngine` service model and of two policy kinds (`:institutional_email` and `:prerequisite_campaign`). The PR contains unit test for all new methods.

For more details on the meaning of these models, I refer to the [corresponding chapter](http://localhost:3004/features/02-registration.html) of the architecture book.


app/models/registration/policy.rb
```diff
@@ -16,5 +16,51 @@ class Policy < ApplicationRecord
...
+        allowed = domains.any? do |domain|
+          email.end_with?("@#{domain}")
+        end
+
+        { pass: allowed, code: allowed ? :ok : :institutional_email_mismatch }
+      end
+
+      def evaluate_prerequisite_campaign(user)
+        campaign_id = config&.fetch("prerequisite_campaign_id", nil)
+        return { pass: true } if campaign_id.blank?
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/models/registration/policy.rb
```diff
@@ -16,5 +16,51 @@ class Policy < ApplicationRecord
...
+    }
+
+    def evaluate(user)
+      case kind.to_sym
+      when :institutional_email
+        evaluate_institutional_email(user)
+      when :prerequisite_campaign
+        evaluate_prerequisite_campaign(user)
+      else
+        { pass: true }
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/services/registration/policy_engine.rb
```diff
@@ -0,0 +1,54 @@
+module Registration
+  class PolicyEngine
```
Splines: A few sentences as class docstring would be nice to have, describing the overall motivation for this class without going into too much detail.

app/services/registration/policy_engine.rb
Splines: not too fond of the idea of having this in `app/services/`. Would you be opposed to just put this file into `app/models/registration` where it is also being used?

app/models/registration/campaign.rb
```diff
@@ -26,5 +26,17 @@ class Campaign < ApplicationRecord
...
+
+    def evaluate_policies_for(user, phase: :registration)
+      Registration::PolicyEngine.new(self).eligible?(user, phase: phase)
+    end
+
+    def policies_satisfied?(user, phase: :registration)
+      evaluate_policies_for(user, phase: phase).pass
+    end
+
+    def user_registered?(user)
```
Splines: I propose to rename this to something containing "confirmed", e.g. `user_registration_confirmed?` for more clarity.

app/models/registration/campaign.rb
Splines: As a sidenote for future PRs: where is the `confirmed` field in `user_registrations.confirmed` coming from, I can't find it in the architecture book. Don't we need a scope for this?

app/services/registration/policy_engine.rb
```diff
@@ -0,0 +1,54 @@
...
+            failed_policy: policy,
+            trace: trace
+          )
+        end
+      end
+
+      Result.new(pass: true, failed_policy: nil, trace: trace)
+    end
+
+    def full_trace_for(user, phase: :registration)
```
Splines: Where is this method needed? I don't find any code occurrence, nor an occurence in the architecture book.

app/models/registration/campaign.rb
```diff
@@ -26,5 +26,17 @@ class Campaign < ApplicationRecord
                     completed: 3 }
 
     validates :title, :registration_deadline, :allocation_mode, :status, presence: true
+
+    def evaluate_policies_for(user, phase: :registration)
+      Registration::PolicyEngine.new(self).eligible?(user, phase: phase)
```
Splines: I suggest something like this to avoid having to recreate a new object all the time:

```
def policy_engine
  @policy_engine ||= Registration::PolicyEngine.new(self)
end
```

app/models/registration/policy.rb
```diff
@@ -16,5 +16,51 @@ class Policy < ApplicationRecord
...
+        allowed = domains.any? do |domain|
+          email.end_with?("@#{domain}")
+        end
+
+        { pass: allowed, code: allowed ? :ok : :institutional_email_mismatch }
+      end
+
+      def evaluate_prerequisite_campaign(user)
+        campaign_id = config&.fetch("prerequisite_campaign_id", nil)
+        return { pass: true, code: :ok } if campaign_id.blank?
```
Splines: Whenever the config cannot be correctly loaded, this indicates a serious flaw and we shouldn't just pass the policy. But not sure if throwing an error would be better here.

Splines (changes requested):
Really cool concept with the PolicyEngine and quite flexible! Here are some first remarks.

app/services/registration/policy_engine.rb
fosterfarrell9: No, that would be fine with me. Do you prefer  them in `registration` or in a `registration/services` subfolder (to indicate that these are not `ActiveRecords` models)?

app/models/registration/policy.rb
fosterfarrell9: I handled this differently in the rewrite.

app/models/registration/policy_engine.rb
```diff
@@ -0,0 +1,54 @@
...
+            failed_policy: policy,
+            trace: trace
+          )
+        end
+      end
+
+      Result.new(pass: true, failed_policy: nil, trace: trace)
+    end
+
+    def full_trace_for(user, phase: :registration)
```
fosterfarrell9: I removed the method and added the link to the architecture book.

PR #932, opened 2025-11-20
Update dependencies

Splines:
As usual before a new release, we update our dependencies (minor versions). simplecov-cobertura is the exception: here we update the major version.

### Off-topic
- Fix new RuboCop errors due to new rules
- Don't open Playwright's [HTML reporter](https://playwright.dev/docs/test-reporters#html-reporter) in CI


fosterfarrell9 (approved):
LGTM.

PR #934, opened 2025-11-21
Remove outdated guides & improve Readme

Splines:
As the title suggests, guides were outdated and we have the [Wiki](https://github.com/MaMpf-HD/mampf/wiki) instead, which also gets updated more frequently as it is independent of the repo. Also improved the wording of the developing section in the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,14 @@
     </p>
   </div>
   <div align="center">
-    <a href="https://mampf.mathi.uni-heidelberg.de">Website (🇩🇪/🇺🇸)</a>
-    | <a href="https://mampf.blog/">Blog (🇩🇪)</a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,7 @@
   </div>
 </div>
 
+
```
fosterfarrell9: We might want to rename the link to "User docs" to something like "MÜSLI integration" or something like that and remove the sentence: "Note the [user docs](https://mampf-hd.github.io/mampf/) are outdated with respect to the screenshots showcased over there."

fosterfarrell9 (changes requested):
Looks good. We might want to rename the "User docs" link in this PR as well, see the comment.

README.md
Splines: ah good point, fixed in 0ed0c5aa958af6e1

PR #946, opened 2025-11-26
Continuous Release 1.19.1

Splines:
This release is just fixing a wrong Node.js version which stopped production. We already used this new version in development, but forgot to update the respective Dockerfile in production.

Also adds a missing env variable.


fosterfarrell9 (approved):
LGTM.

PR #963, opened 2025-12-06
Downgrade connection pool gem

f-buerckel:
connection_pool is a gem needed by Activesupport. 
In its 3.0 release, connection_pool introduced breaking changes in the signature of the initialize function (see this [PR](https://github.com/mperham/connection_pool/pull/209)) that break activesupport resulting in the error: ArgumentError: wrong number of arguments (given 1, expected 0) (ArgumentError) during build.

Therefore we should currently limit connection_pool to versions before 3.0, until this is resolved.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -170,7 +170,7 @@ GEM
       sprockets-rails
       will_paginate
     concurrent-ruby (1.3.5)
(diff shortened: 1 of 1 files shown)


f-buerckel:
done

PR #979, opened 2025-12-22
Muesli: PR 4.3a - Solver

fosterfarrell9:
This PR implements the core solver logic for preference-based registration campaigns. It realizes one part of PR 4.3 in the architecture book such that 4.3 rather comes in two parts that each have smaller size. The PR introduces the backend machinery required to calculate optimal allocations based on user preferences and capacity constraints. It contains:

- `Registration::Solvers::MinCostFlow`: Implementation of a Minimum Cost Max Flow algorithm using Google OR-Tools. It maps users and items to a flow network to minimize the total "cost" (where lower preference ranks = lower cost). The solver is designed to find the "social optimum", potentially assigning a user their 2nd choice if it prevents another user from being unassigned entirely. It will try to fill all slots even if that means that some users get non-preferred slots (rather than just saying taht there is no solution that meets all conditions).
- support for switching forced assignments off: by default, users will  be assigned to non-preferred slots if their choices are full, rather than leaving them empty-handed. It is possible to switch that off.
- Perturbed Cost Logic: Implements a "scaled cost with noise" strategy by adding small, random noise to preference weights. This ensures fair, randomized tie-breaking for equal-rank contentions (effectively acting as a lottery for single-item campaigns) while strictly preserving rank hierarchy and preventing bias based on database IDs or insertion order.
- `Registration::Allocation Service`: A service layer to interface between the application and the solver strategy. This is only a basic skeleton, the rest will follow in 4.3b.
- `Registration::AllocationStats` to generate metrics (e.g., % top choice, global average rank) for evaluating allocation quality.
- full coverage by unit tests of the above models (in extensive specs for the solver)

You can verify the solver logic interactively in  the rails console using the provided rake tasks (I would recommend this to just get a feeling). First you need to seed the development database using the appropriate task. Then, set up the campaign and registration data:

```shell
# Create a sample campaign with items
rails solver:create_campaign

# Populate it with random user registrations and preferences
rails solver:create_registrations
```

Then, in the rails console, you can do
```ruby
# Load the campaign
campaign = Registration::Campaign.last

# Run the allocation
service = Registration::AllocationService.new(campaign)
allocation = service.allocate!
# => { user_id => item_id, ... }

# Inspect Statistics
stats = Registration::AllocationStats.new(campaign, allocation).calculate# Load the campaign
pp stats
```

Note that the scenario constructed via the above task is a "picky eater" scenario where even an optimal solution will have many forcded assignments. If you want to want to run the allocation with forced assignments disabled, use 
```ruby
service = Registration::AllocationService.new(campaign, force_assignments: false)
```
instead.

Notes for the reviewer:

- This PR adds the [or-tools](https://github.com/ankane/or-tools-ruby) gem. When bundling for the first time, this take a lot of time since it is compiled from C++ sources. I have not yet found a good way to reduce this time.

### DIFF
--- a/app/models/lecture.rb
+++ b/app/models/lecture.rb
@@ -794,6 +794,26 @@ class Lecture < ApplicationRecord
     touch
   end
 
+  def ensure_roster_membership!(user_ids)
+    # Efficiently insert missing memberships using upsert (ignoring duplicates)
+    # Note: Requires a unique index on [:user_id, :lecture_id]
+    attributes = user_ids.map do |uid|
+      { user_id: uid, lecture_id: id, created_at: Time.current,
+        updated_at: Time.current }
+    end
+
+    return if attributes.empty?
+
+    # Skipping validations for performance; ensure data integrity via unique index
+    # rubocop:disable Rails/SkipsModelValidations
+    LectureMembership.upsert_all(
+      attributes,
+      unique_by: [:user_id, :lecture_id],
+      record_timestamps: false # Timestamps handled manually above
+    )
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+
   def eligible_as_tutors
     (tutors + Redemption.tutors_by_redemption_in(self) + editors + [teacher]).uniq
     # the first one should (in the future) actually be contained in the sum of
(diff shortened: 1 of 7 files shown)


Splines:
Closing as this PR was absorbed into #1081.

PR #987, opened 2025-12-26
Update dependencies

Splines:
As usual...

### RuboCop pinned

Note that I temporarily pinned RuboCop during the updating since new versions produced very weird errors like this:

It's probably a RuboCop bug that will soon be fixed and we can wait for that.ixed and we can wait for that. that.


Splines:
I will squash-merge even though the one Playwright test is failing. This is the new one I added in #983 and apparently it is flaky. All other Playwright tests went through fine multiple times. I cannot reproduce the failing test locally, will have to investigate this, but maybe in another PR.

PR #1012, opened 2026-01-14
Decaffeination of lectures

EliaPfl:
This pull request refactors the lectures management UI and backend to use Turbo Frames and Turbo Streams for a more modern, interactive, and maintainable experience. It removes CoffeeScript and AJAX-based flows, simplifies controller logic, and updates the views to fully leverage Hotwire. Additionally, new end-to-end tests are added to ensure correct behavior for creating and editing lectures


e2e/lectures.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+      { medium_id: sourceMedium.id, teachable_id: targetLecture.id, teachable_type: "Lecture" });
+
+    await page.goto(`/lectures/${targetLecture.id}/edit?tab=content`);
+    await page.getByRole("button", { name: "Import Media" }).click();
+    await expect(page.getByRole("heading", { name: "Imported Media ( 1)" })).toBeVisible();
+    await expect(page.locator("#importedMediaTable")).toContainText(sourceMedium.description);
+
+    await page.getByTitle("Delete").click();
+    await expect(page.getByRole("heading", { name: "Imported Media (0)" })).toBeVisible();
+    await expect(page.getByText("There are no imported media.")).toBeVisible();
```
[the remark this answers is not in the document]
EliaPfl: Implemented

app/frontend/js/media.coffee
```diff
@@ -239,14 +239,21 @@ $(document).on 'turbo:load', ->
...
         }
-        error: (jqXHR, textStatus, errorThrown) ->
-          console.log("AJAX Error: #{textStatus}")
+        body: JSON.stringify(media_ids: selected)
+      .then (response) ->
+        response.text()
+      .then (streamMessage) ->
+        Turbo.renderStreamMessage(streamMessage)
+      .catch (error) ->
+        console.log("Turbo Import Error: #{error}")
```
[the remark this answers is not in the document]
PN-1000: fixed

app/controllers/lectures_controller.rb
```diff
@@ -85,31 +85,41 @@ def create
...
+      render turbo_stream: streams
+    else
+      @from = params.dig(:lecture, :from)
+
+      render turbo_stream: turbo_stream.update(turbo_frame_request_id,
+                                               template: "lectures/new/_new", locals: {
+                                                 lecture: @lecture, from: @from
+                                               }),
+             status: :unprocessable_content
     end
```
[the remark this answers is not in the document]
PN-1000: no fallback needed

app/controllers/lectures_controller.rb
```diff
@@ -67,9 +67,9 @@ def new
       @lecture.annotations_status = 0
     end
 
-    respond_to do |format|
-      format.js { render template: "lectures/new/new" }
-    end
+    render turbo_stream: turbo_stream.update(turbo_frame_request_id, template: "lectures/new/_new",
+                                                                     locals: { lecture: @lecture,
+                                                                               from: @from })
```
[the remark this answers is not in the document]
PN-1000: fixed

app/controllers/lectures_controller.rb
```diff
@@ -85,31 +89,44 @@ def create
...
+
+      streams = []
+
+      if params.dig(:lecture, :from) == "course"
+        streams << turbo_stream.update("course_lectures",
+                                       partial: "courses/lectures_list",
+                                       locals: { course: @lecture.course })
+        streams << turbo_stream.update(Lecture.new,
+                                       partial: "lectures/new/new",
+                                       locals: { lecture: @lecture, from: "course" })
```
fosterfarrell9: Why do you pass `@lecture` (whioch contains the saved lecture) to the new partial? I don't understand this semantically. I would have assumed one would take a new lecture (that has the appropriate settings that are related to the course that provides the context).

app/frontend/lectures/new/_new.html.erb
```diff
@@ -1,85 +1,108 @@
...
-             id="new-lecture-term-error">
+    </div>
+  <% end %>
+  <% if Course.count == 0 %>
+    <%= t('admin.lecture.no_courses_yet') %>
+  <% elsif Term.count == 0 %>
+    <%= t('admin.lecture.no_terms_yet') %>
+  <% else %>
+    <%= form_with model: lecture,
+                  data: { action: 'turbo:submit-end->administration--index--create-buttons#submitEnd' } do |f| %>
```
fosterfarrell9: This line makes the partial depend implicitly on a page-specific parent controller. From my understanding this controller is mounted on the admin lectures index card in `_my_lectures.html.erb` but not in the course-modal context. This asymmetry seems brittle: future changes might break one context while the other context still works, and it is also not. I guess the simplest fix would be to compute the form data conditionally. 

Even better fro a design point of view (but would require more work, idk if its really worth it, and I also didn't think this through completely): Create a "neutral" stimulus controller for the shared form which handles form-lifecyle and emits a custom DOM event like `lecture:new:success`, keep the toggling in the `create_buttons.controller.js` but change it to listen for that custom event.

app/frontend/modal/turbo_modal.controller.js
```diff
@@ -0,0 +1,33 @@
...
+      headers: {
+        "Accept": "text/vnd.turbo-stream.html, text/html, application/xhtml+xml",
+        "Turbo-Frame": this.frameIdValue,
+      },
+    });
+
+    if (response.ok) {
+      const html = await response.text();
+      Turbo.renderStreamMessage(html);
+    }
```
fosterfarrell9, line 64: You might consider doing at least some form of error-handling for non-ok status.

fosterfarrell9 (changes requested):
Very nice PR! I just got some minor comments that should be adressed.

app/frontend/modal/turbo_modal.controller.js
PN-1000: added error-handling

app/controllers/lectures_controller.rb
PN-1000: fixed it, form is now reset to spinner

app/frontend/lectures/new/_new.html.erb
PN-1000: removed implicit dependency by adding new form controller

app/helpers/form_unknown_error_helper.rb
```diff
@@ -20,12 +20,11 @@ def form_with(**options, &)
...
     # no field-specific error markup is present.
     def add_whole_form_error_message(form_object, form_html)
-      if !form_object.respond_to?(:errors) || form_object.errors.empty? \
-        || form_html.include?('class="invalid-feedback"')
-        return form_html
-      end
+      return form_html if !form_object.respond_to?(:errors) || form_object.errors.empty?
 
       doc = Nokogiri::HTML::DocumentFragment.parse(form_html)
+      return form_html if doc.css(".invalid-feedback").any?
```
Splines, line 26: Thanks, good point, this is indeed more robust by handling css class checking via `doc.css()` instead of the custom string checking.

Splines: Also updated the blog post accordingly [here](https://splines.me/blog/2025/server-side-validation-rails-turbo#bonus-general-form-errors).

app/frontend/js/_turbo_fix_selectize.js
```diff
@@ -148,5 +148,33 @@ $(document).on("turbo:before-cache", function () {
 });
 
 $(document).on("turbo:load", function () {
-  fillOptionsByAjax($(".selectize"));
+  fillOptionsByAjax($(".selectize").filter(function () {
+    return !this.tomselect;
+  }));
+});
+
+$(document).on("turbo:frame-load", function (event) {
```
Splines, line 157: This files includes some fixes. Could you please add docstrings to explain why they are needed? Especially also the selectize handling for which you use the `turbo:before-stream-render` (a custom event that does not come natively with Turbo)

app/frontend/administration/index/create_buttons.controller.js
```diff
@@ -0,0 +1,28 @@
...
+
+  show() {
+    document.getElementById("new-lecture-button")?.classList.remove("d-none");
+    document.getElementById("new-course-button")?.classList.remove("d-none");
+  }
+
+  cancel(event) {
+    event.preventDefault();
+    this.show();
+    document.getElementById("new_lecture")?.replaceChildren();
```
Splines, line 26: thanks, didn't know about this better alternative to ".innerHTML = ''"

app/frontend/js/media.coffee
```diff
@@ -239,14 +239,23 @@ $(document).on 'turbo:load', ->
...
-          console.log("AJAX Error: #{textStatus}")
+        body: JSON.stringify(media_ids: selected)
+      .then (response) ->
+        unless response.ok
+          throw new Error("Import failed with status #{response.status}")
+        response.text()
+      .then (streamMessage) ->
+        Turbo.renderStreamMessage(streamMessage)
+      .catch (error) ->
+        console.log("Turbo Import Error: #{error}")
```
Splines: This is too low-level, there must be a better way to do all of this without having to parse the csrf token and manually setting the Turbo Stream MIME type.

app/frontend/lectures/edit/_preferences.html.erb
```diff
@@ -28,8 +30,8 @@
                       options_for_select(Lecture.select_sorts, lecture.sort),
                       {},
                       { class: 'form-select' } %>
-        <div class="invalid-feedback"
-              id="lecture-term-error">
+          <div class="invalid-feedback d-block"
```
Splines: Instead of manually rendering the form errors, please refer to my blog post [here](https://splines.me/blog/2025/server-side-validation-rails-turbo) and especially see also [this line in MaMpf](https://github.com/MaMpf-HD/mampf/blob/9abc08262d693dd35c0f1c7b4e48bfac475b074d/config/initializers/form_errors.rb#L12).

app/frontend/lectures/edit/_seminar_content.html.erb
```diff
@@ -16,7 +16,7 @@
                     new_talk_path(lecture_id: lecture.id),
                     remote: true,
                     class: 'btn btn-sm btn-secondary new-in-lecture',
-                    data: { cy: 'new-talk-btn' },
+                    data: { cy: 'new-talk-btn', testid: 'new-talk-btn' },
```
Splines: I don't see any Cypress test anymore that references "new-talk-btn", so can we remove it? Furthermore, for this button, maybe with Playwright try to use `getByRole` instead of hardcoding the test id here. This is in order to rely more on what a user of the site would see (instead of adding "instrumentation" to the DOM itself).

Splines: Same for other occurences of `testid` in this PR. Let's check if we can get rid of that (of course, not always possible).

app/frontend/modal/turbo_modal.controller.js
```diff
@@ -0,0 +1,45 @@
...
+    if (!modalEl) return;
+
+    const modal = Modal.getOrCreateInstance(modalEl);
+    modal.show();
+
+    const response = await fetch(this.urlValue, {
+      headers: {
+        "Accept": "text/vnd.turbo-stream.html, text/html, application/xhtml+xml",
+        "Turbo-Frame": this.frameIdValue,
+      },
```
Splines: Same here, there's probably some better way instead of manually setting the header for a Turbo Stream. But granted that this is a general-purpose Modal controller, I agree that here it might be "more ok" than at other places. Still, let's look for a more elegant solution ;)

app/helpers/lectures_helper.rb
```diff
@@ -148,18 +150,19 @@ def teacher_select(form, is_new_lecture, lecture = nil)
...
                                                               placeholder: t("basics.enter_two_letters"), # rubocop:disable Layout/LineLength
                                                               no_results: t("basics.no_results"),
                                                               modal: true,
-                                                              cy: "teacher-admin-select"
+                                                              testid: "teacher-admin-select"
                                                             } })
 
-      error_div = content_tag(:div, "", class: "invalid-feedback", id: "lecture-teacher-error")
+      error_div = content_tag(:div, teacher_error, class: "invalid-feedback d-block",
+                                                   id: "lecture-teacher-error")
```
Splines: see my other comment about the blog post for error handling

app/views/courses/_basics.html.erb
```diff
@@ -91,37 +91,20 @@
...
                     class: 'fas fa-plus-circle text-dark',
                     id: 'new-lecture-button',
-                    data: { remote: true,
-                            toggle: 'tooltip',
-                            cy: 'new-lecture-button-admin-index' },
+                    data: { controller: 'turbo-modal',
+                            turbo_modal_url_value: new_lecture_path(params: { from: 'course', course: course.id }),
+                            turbo_modal_selector_value: '#newLectureModal',
+                            turbo_modal_frame_id_value: 'new_lecture',
+                            turbo_modal_error_message_value: t('modals.load_error'),
```
Splines: I think it makes more sense to find some way to hardcode this for any model as I don't see the purpose of changing this message on a "per-modal" basis. Just as some ideas: see the "js-routes" gem we use, or maybe even just put a span with data attribute on every modal that the JS can read out.

Splines: Furthermore, I'd be happy if we could get rid of `_value` for these properties, what else than a "value" could it be? I think `turbo_modal_something` is sufficient.

app/views/talks/_basics.html.erb
```diff
@@ -23,7 +23,8 @@
       <div class="col-12">
         <%= f.label :date, t('basics.date'),
                     class: "form-label" %>
-        <a class="fas fa-plus-circle clickable text-dark"
+        <a href="#"
```
Splines: what is `href="#"` used for?

e2e/lecture.new.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  course: any,
+  term: any,
+  isCoursePrefilled: boolean,
+) {
+  const button = "new-lecture-button-admin-index";
+
+  await page.getByTestId(button).click();
+
+  if (isCoursePrefilled) {
+    await page.waitForResponse(response => response.url().includes("/lectures/new") && response.status() === 200);
```
Splines: To avoid flakyness: Please call `waitForResponse` before performing the action, save it in a variable, then perform the action, and only then await the promise. See [the Playwright docs](https://playwright.dev/docs/network#variations).

e2e/lectures.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+        await expectUsersInDropdown(page, "tutor-select-div", student.user, true);
+      });
+  });
+});
+
+test.describe("Seminar speakers (new talk)", () => {
+  async function openTalkForm(page: Page) {
+    await page.getByTestId("new-talk-btn").click();
+    await page.waitForResponse(resp => resp.url().includes("/talks/new"));
+    await page.waitForTimeout(500); // Wait for form initialization
```
Splines, line 110: Please don't use timeouts, unless you went through every other option there exist. This is to avoid flakiness

e2e/lectures.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  test.describe("when logged in as teacher", () => {
+    test("prohibits searching for arbitrary users in the speakers dropdown",
+      async ({ factory, teacher: { page, user: teacher }, student }) => {
+        const seminar = await factory.create("seminar", [], { teacher_id: teacher.id });
+
+        await page.goto(`/lectures/${seminar.id}/edit`);
+        await openTalkForm(page);
```
Splines, line 119: I think some abstraction by using Playwright page objects could help the tests at some points.

e2e/lectures.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+      { lesson_id: sourceLesson2.id });
+
+    const targetLecture = await factory.create("lecture");
+
+    await page.goto(`/lectures/${targetLecture.id}/edit?tab=content`);
+    await page.getByRole("button", { name: "Import Media" }).click();
+    await page.getByRole("combobox", { name: "Types" }).fill("Les");
+    await page.locator("#search_media_types-opt-1").click();
+    await page.getByRole("button", { name: "Search" }).click();
+    await page.waitForResponse(resp => resp.url().includes("/media/search") && resp.status() === 200);
```
Splines: same as mentioned in my other comment regarding the `waitForResponse`

Splines (dismissed):
Thanks a lot for this PR, works great! Here are some comments form my first review round.

app/frontend/lectures/edit/_preferences.html.erb
PN-1000: fixed, by mirroring the errors to corresponding fields

app/helpers/lectures_helper.rb
PN-1000: fixed, by mirroring the errors to corresponding fields

app/views/talks/_basics.html.erb
PN-1000: herb linter requested this

app/frontend/js/_turbo_fix_selectize.js
PN-1000: added comments, I think `turbo:before-stream-render` is a native turbo event (see [here](https://turbo.hotwired.dev/reference/events#turbo%3Abefore-stream-render))

app/views/courses/_basics.html.erb
PN-1000: moved the message to a meta tag, `_value` is needed by [stimulus](https://stimulus.hotwired.dev/reference/values)

app/frontend/js/_turbo_fix_selectize.js
Splines: ah yeah true, I confused that with the custom `turbo:stream-render` that we added in `initHotwire.js`

app/views/talks/_basics.html.erb
Splines: not sure how herb linter could have requested this as currently it shouldn't even lint that file (we've exluded `app/views/**/*.html.erb` in the `.her.yml` config)

I removed the `href="#"` in c91f5162c

app/views/courses/_basics.html.erb
Splines: nice, moving it to a meta tag is a really cool idea!

app/frontend/lectures/edit/_seminar_content.html.erb
Splines: I removed the cypress test key in 3d5849d2c

app/views/talks/_basics.html.erb
Splines: Ah, i see now, it shows this error only in the CI. I've opened an issue for it here: https://github.com/marcoroth/herb/issues/1595

Splines: will ignore this error here for now

e2e/lecture.new.spec.ts
Splines: fixed in 796354982 and 43e0aaa4c

e2e/lectures.spec.ts
Splines: fixed in 43e0aaa4c

e2e/lectures.spec.ts
fosterfarrell9: Fixed in [a4dc5ed](https://github.com/MaMpf-HD/mampf/pull/1012/commits/a4dc5ed913c0cc8d87fe597a4937004974027ec1).

e2e/lectures.spec.ts
fosterfarrell9: Fixed in [a4dc5ed](https://github.com/MaMpf-HD/mampf/pull/1012/commits/a4dc5ed913c0cc8d87fe597a4937004974027ec1).

app/models/lecture.rb
```diff
@@ -100,7 +100,7 @@ class Lecture < ApplicationRecord
 
   # we do not allow that a teacher gives a certain lecture in a given term
   # of the same sort twice
-  validates :course, uniqueness: { scope: [:teacher_id, :term_id, :sort] }
+  validates :course_id, uniqueness: { scope: [:teacher_id, :term_id, :sort] }
```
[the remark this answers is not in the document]
fosterfarrell9: Nope.

app/frontend/js/media.coffee
```diff
@@ -241,12 +241,11 @@ $(document).on 'turbo:load', ->
...
       $.ajax Routes.lecture_import_media_path(lectureId),
         type: 'POST'
-        dataType: 'script'
-        data: {
-          media_ids: selected
-        }
-        error: (jqXHR, textStatus, errorThrown) ->
-          console.log("AJAX Error: #{textStatus}")
+        contentType: 'application/json'
+        headers: { 'Accept': 'text/vnd.turbo-stream.html' }
```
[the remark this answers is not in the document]
fosterfarrell9: Nope.

fosterfarrell9 (dismissed):
Thank you for your effort! It took a while until I found the time to look into it, but I finally manged it.

fosterfarrell9 (dismissed):

fosterfarrell9 (approved):

PR #1017, opened 2026-01-21
🎈 Muesli: Registration and roster management

fosterfarrell9:
Details to follow.

## TODO (will be done in the `bowl-allocation` branch)
- [x] Remove the `purpose` column from cohorts
- [x] Add columns for location and time to the tutorials table
- [x] Right before merging, update all the DB migration dates.
- [x] Right before merging, update versions of newly added Gems to the latest.
- [ ] Remove unused i18n keys -> use script to find unused keys
- [ ] Generate default registration data for local development
- [ ] Add link to feedback pane since the features are quite new. E.g. below the participants tab, but watch out for participants sideview in this case that might wrap below.


Splines:
Closed in favor of #1081

PR #1026, opened 2026-01-24
Import spec files from `muesli-bowl`

Splines:
This branch is on top of the one that only contains the `.rb` files changed in the `muesli-bowl` branch #1017. Now, we also include the spec files here to allow for a separate Copilot review.


Splines:
Closing as we created this PR only to help with review of #1081, which is now merged.

PR #1043, opened 2026-02-14
Lint HTML & CSS files

Splines:
To have more fun during review of the Müsli Bowl, I add HTML & CSS linting (no SCSS linting) to our codebase.

Note that ERB indentation is not respected in `.html.erb` files. If you'd like to have that, vote up for this issue: https://github.com/yeonjuan/html-eslint/issues/422

I've also fixed a wrong JS setup for the brace style rule. With the new change, `else` should always come on the same line as the closing `}` of the `if` statement.


Splines:
The output looks somewhat ugly due to the ERB linting. I forgot that I wanted to rather try [Herb](https://herb-tools.dev/overview), wow 😅

PR #1052, opened 2026-02-17
🗳️ Müsli: Student registration

plienta:
In this PR, we introduce the student-facing views for the new registration flows (see the Müsli integration).

### Registration for campaigns (see also Müsli 3.3, 3.4 and 4.1 #1051):

Introduces a redesigned student registration flow supporting both First-Come-First-Served (FCFS) and Preference-Based (PB) campaign modes.

-  Added RESTful actions (`index`, `create`, `update`, `destroy`, etc.) to `UserRegistrationsController` for handling student registration, preference management, and campaign details

### Self-enrollment (see Müsli 5.4 #1010): 

Enables students to self-add or self-remove themselves from rosterable entities (Tutorials, Talks, Cohorts, Lectures)
- Introduces the `Roster::SelfMaterializationController`
- Updated `LectureAbility` to grant the `self_materialize` permission to users who are not tutors or the teacher for a lecture.
- Configurable permissions.
  - Self-add: allowed when `(add_only OR add_and_remove) AND !locked AND !full`
  - Self-remove: allowed when `(remove_only OR add_and_remove) AND !locked`

### UI Access

Both registration via campaigns and self-materialization can be accessed through the **"Enrolment"** tab in the lecture.
- Added an "Enrollment" sidebar link in the lecture view, gated by the `registration_campaigns` feature flag, to provide easy access to the new registration interface.

### Off-topic

As per [this security implication](https://github.com/MaMpf-HD/mampf/pull/1052#discussion_r3254380429), I replaced `.constantize` with a map-construct, but also for other register-related controllers.

---

image


app/frontend/registration/main/index.html.erb
```diff
@@ -0,0 +1,99 @@
...
+        <%= turbo_frame_tag dom_id(campaign_details.campaign, :main_student_registration_campaign) do %>
+          <%= render partial: 'registration/main/campaign_card',
+                  locals: {
+                    details: campaign_details,
+                    campaign: campaign_details.campaign
+                  } %>
+        <% end %>
+      </div>
+    <% end %>
+    <!-- Collapsible section for completed campaigns -->
```
plienta: add if here

app/models/registration/campaign.rb
```diff
@@ -63,14 +63,43 @@ def policies_satisfied?(user, phase: :registration)
       evaluate_policies_for(user, phase: phase).pass
     end
 
+    def evaluate_full_trace_for(user, phase: :registration)
```
plienta: check if it is used, if not then remove

app/views/talks/_base_data.html.erb
```diff
@@ -1,9 +1,7 @@
-<p>
```
plienta, line 1: revert this file change

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -21,8 +26,118 @@ def destroy_for_user
...
+      @rosterized_entries = Rosters::StudentMaterializedResultResolver
+                            .new(current_user)
+                            .all_rosterized_for_lecture(@lecture)
+      @self_rosterables = Rosters::SelfRosterOptionsQuery.new(@lecture, current_user)
+                                                         .call
+      render template: "registration/main/index",
+             layout: turbo_frame_request? ? "turbo_frame" : "application"
+    end
+
+    def create
```
fosterfarrell9, line 41: Both the `create` and `update` student endpoints do not enforce campaign-mode serverside. `create` is using `LectureFCfsEditService` and `update` is using `LecturePreferenceEditService`, but neither of these services seem to check the actual registration mode for the campaign. Since the routes are public, a crafted `POST` to `register_item` against a preference-based campaign might possibly create an immediate `confirmed` registration with no rank, bypassing the preference workflow completely.

app/models/registration/user_registration.rb
```diff
@@ -45,14 +45,6 @@ class UserRegistration < ApplicationRecord
...
               if: -> { registration_campaign.first_come_first_served? }
 
-    # FCFS campaigns: one row per user+campaign
-    # There is also a DB index to enforce it at the database level (see the schema).
-    validates :user_id,
-              uniqueness: {
-                scope: :registration_campaign_id
-              },
-              if: -> { registration_campaign.first_come_first_served? }
-
```
fosterfarrell9, line 55: Why did you remove this (and the DB index as well)? What remains now is an application check in `registration/user_registration/handler.rb`, but `LectureFcfsEditService` appears to lock only the chosen item row. I guess that two concurrent requests for two tutorials in the same campaign can therefore both pass the check and insert two `confirmed` registrations for the same user and group type.

db/migrate/20260223163358_remove_unique_unranked_index_from_registration_user_registrations.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveUniqueUnrankedIndexFromRegistrationUserRegistrations < ActiveRecord::Migration[8.0]
+  def change
+    remove_index :registration_user_registrations, name: "index_reg_user_regs_unique_unranked"
```
fosterfarrell9, line 3: Why did you remove this index (and the corresponding model validation as well)? See my comment on the removed model validation in the `UserRegistration` model.

spec/helpers/registration/user_registrations_helper_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+      expect(helper.get_mode_info(0)).to include(:mode_name, :abbr, :badge_class)
+    end
+
+    it "returns unknown mode for missing key" do
+      expect(helper.get_mode_info(999)).to eq(Registration::UserRegistrationsHelper::MODE_MAP[-1])
+    end
+  end
+
+  describe "#get_policy_config_info" do
+    context "lecture_performance" do
```
fosterfarrell9: Note that this has been renamed to `student_performance`

app/helpers/roster_helper.rb
```diff
@@ -205,4 +205,87 @@ def roster_group_badge(group, group_type)
...
   end
+
+  def rosterable_display_type(rosterable)
+    case rosterable.class.name
+    when "Tutorial"
+      t("registration.item.types.tutorial")
+    when "Talk"
+      t("registration.item.types.talk")
+    when "Cohort"
+      base_type = case rosterable.purpose.to_sym
```
fosterfarrell9: `purpose` has been deprecated for cohorts in the latest version of `bowl-allocation` (which is the successor of muesli-bowl).

config/locales/roster/de.yml
```diff
@@ -287,6 +287,8 @@ de:
       user_not_found: "Studierende/r nicht gefunden."
```
fosterfarrell9, line 279: There are missing translations (here and in other locale files):

`roster.errors.lecture_has_other_roster_entry`
(missing in de)

`roster.warnings.lecture_has_other_roster_entry`
(missing in en)

`registration.user_registration.options_title`
(missing in de)

`registration.user_registration.option_title` 
(missing in en)

(just run `i18n-tasks missing` to see these).

spec/controllers/user_registrations_spec.rb
```diff
@@ -0,0 +1,176 @@
+require "rails_helper"
```
fosterfarrell9: Why is there both a controller spec and a request spec for the `UserRegistrationController`? I would think that a request spec is enough.

spec/helpers/registration/user_registrations_helper_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+      let(:policy) do
+        { kind: "institutional_email", config: { "allowed_domains" => ["a.com", "b.com"] } }
+      end
+      it { expect(helper.get_policy_config_info(policy)).to eq("a.com, b.com") }
+    end
+
+    context "prerequisite_campaign" do
+      let(:policy) do
+        { kind: "prerequisite_campaign", config: { "prerequisite_campaign_id" => 42 } }
+      end
```
fosterfarrell9, line 41: This one fails.

spec/requests/roster/self_materialization_controller_spec.rb
```diff
@@ -0,0 +1,156 @@
+require "rails_helper"
+
```
fosterfarrell9, line 2: All of these fail.

app/abilities/lecture_ability.rb
```diff
@@ -34,5 +34,9 @@ def initialize(user)
     can :subscribe_page, Lecture do |lecture|
       lecture.published? || user.active_teachable_editor?
     end
+
+    can :self_materialize, Lecture do |lecture|
+      !user.in?(lecture.tutors) && (user != lecture.teacher)
```
fosterfarrell9: What do we do about users who can edit the lecture? I think they should also be barred from self materialization. Of course, some time a teacher might even give seminar talk themselves, so we should discuss this in our team.

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -21,8 +26,118 @@ def destroy_for_user
...
+      @campaign = @item.user_registrations.find_by!(user_id: current_user.id,
+                                                    status: :confirmed)
+                       .registration_campaign
+      result = Registration::UserRegistration::LectureFcfsEditService
+               .new(@campaign, current_user).withdraw!(@item)
+      respond_to_student_registration(result,
+                                      I18n.t("registration.user_registration.messages.withdrawn"))
+    end
+
+    def render_not_found(exception)
```
fosterfarrell9: This does not correspond to a route for this controller, so I think this should better be a private method.

app/helpers/registration/user_registrations_helper.rb
```diff
@@ -12,5 +12,163 @@ def sort_registrations_by_rank(registrations)
...
+             badge_class: "bg-light-blue-lighten-4 text-darkblue" },
+      1 => { mode_name: "Preference Based", abbr: "PB",
+             badge_class: "bg-mdb-color-lighten-3 text-white" }
+    }.freeze
+    def get_mode_info(mode)
+      MODE_MAP.fetch(mode, MODE_MAP[-1])
+    end
+
+    # Examples of policy config:
+    # lecture_performance -> config: { certification_status: :pending }
```
fosterfarrell9: `lecture_performance` has been renamed to `student_performance`.

app/models/registration/campaign/campaign_details_service.rb
```diff
@@ -0,0 +1,75 @@
...
+        trace.each do |policy_result|
+          next unless policy_result[:kind] == "prerequisite_campaign"
+
+          id = policy_result[:config]["prerequisite_campaign_id"]
+          campaign = Registration::Campaign.find_by(id: id)
+          policy_result[:config]["prerequisite_campaign"] =
+            if campaign
+              "#{campaign&.campaignable&.title}: #{campaign&.description}"
+            else
+              "Campaign not found"
```
fosterfarrell9: Can this be visible to users? Then it should be i18ned.

app/models/registration/campaign.rb
```diff
@@ -67,10 +67,35 @@ def open_for_registrations?
       open? && registration_deadline > Time.current
     end
 
+    def open_for_withdrawals?
+      open? && registration_deadline > Time.current
```
fosterfarrell9: Instead, you could also just write `open_for_registrations?` in this line. This makes it more apparent that as of now, this is the same condition.

app/models/registration/item.rb
```diff
@@ -41,6 +41,30 @@ class Item < ApplicationRecord
     validate :validate_capacity_reduction, on: :update
     before_destroy :ensure_campaign_is_draft
 
+    def item_capacity_used
+      confirmed_registrations_count
+    end
+
+    def capacity_remained
```
fosterfarrell9: This sounds strange to me in english. Maybe `remaining_capacity` might be better.

app/models/rosters/rosterable.rb
```diff
@@ -61,6 +61,24 @@ def locked?
...
+
+    def config_allow_self_remove?
+      ["remove_only", "add_and_remove"].include?(self_materialization_mode)
+    end
+
+    # guard in FE for self-removal possibility
+    def allow_self_remove?(user)
+      config_allow_self_remove? && !locked? && user_allocated?(user)
+    end
+
```
fosterfarrell9, line 90: The code block from l64-80 could be rewritten in a more "Rails way" as

```
def config_allow_self_add?
  self_materialization_mode_add_only? ||
    self_materialization_mode_add_and_remove?
end

def config_allow_self_remove?
  self_materialization_mode_remove_only? ||
    self_materialization_mode_add_and_remove?
end

def allow_self_add?(user)
  return false unless config_allow_self_add?
  return false if locked?
  return false if user_allocated?(user)

  !full?
end

def allow_self_remove?(user)
  return false unless config_allow_self_remove?
  return false if locked?

  user_allocated?(user)
end
```

(I think, better check). This makes it a lot more readable.

app/models/rosters/self_roster_options_query.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def lecture_rosterables
+      # get all rosterables for the lecture
+      rosterables = []
+      rosterables.concat(@lecture.talks)
+      rosterables.concat(@lecture.tutorials)
+      rosterables.concat(@lecture.cohorts)
+      # filter rosterables that allow self add or self remove
+      rosterables.select do |rosterable|
+        rosterable.self_materialization_mode in ["add_only", "remove_only", "add_and_remove"]
```
fosterfarrell9: Since you already have the corresponding methods in `Rosterable` you could also replace this line by

```
rosterable.config_allow_self_add? || rosterable.config_allow_self_remove?
```

Then you can also remove the comment above.

app/models/rosters/student_materialized_result_resolver.rb
```diff
@@ -0,0 +1,44 @@
...
+      registerables = tutorials + cohorts + talks
+      name = []
+      if registerables.any?
+        registerables.each do |registerable|
+          name << "#{registerable.class.name.humanize} #{registerable.title}"
+        end
+      end
+      name.any? ? name.join(", ") : nil
+    end
+  end
```
fosterfarrell9, line 50: I think, semantically `joins` makes more sense here than `includes` (although the latter works as well). If you want to write this more in a "Rails way", you could also do it like this:


```ruby
def all_rosterized_for_lecture(lecture)
  names =
    lecture.tutorials
           .joins(:tutorial_memberships)
           .where(tutorial_memberships: { user_id: @user.id })
           .map { |tutorial| rosterable_label(tutorial) } +
    lecture.cohorts
           .joins(:cohort_memberships)
           .where(cohort_memberships: { user_id: @user.id })
           .map { |cohort| rosterable_label(cohort) } +
    lecture.talks
           .joins(:speaker_talk_joins)
           .where(speaker_talk_joins: { speaker_id: @user.id })
           .map { |talk| rosterable_label(talk) }

  names.presence&.join(", ")
end

private

  def rosterable_label(rosterable)
    "#{rosterable.class.model_name.human} #{rosterable.title}"
  end
```

fosterfarrell9 (changes requested):
Thank you for this nice PR! As usual, I have some comments 😉

app/controllers/registration/user_registrations_controller.rb
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55 , I have added the check in service

app/helpers/roster_helper.rb
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55

spec/requests/roster/self_materialization_controller_spec.rb
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55

app/models/registration/item.rb
plienta: resolved in ced7cb1f403d4c95e3f8d2b7b6aa96ed14b74a8d

spec/helpers/registration/user_registrations_helper_spec.rb
plienta: resolved in 4634e431b59a3420240adcb133fa6badc258d99f

app/abilities/lecture_ability.rb
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55

app/controllers/registration/user_registrations_controller.rb
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55

app/helpers/registration/user_registrations_helper.rb
plienta: resolved in 4634e431b59a3420240adcb133fa6badc258d99f

app/models/rosters/rosterable.rb
plienta: resolved in 4f7115e1ce12f360ffea63a991bf0e31dd0fadcc

app/models/rosters/student_materialized_result_resolver.rb
plienta: resolved in 4f7115e1ce12f360ffea63a991bf0e31dd0fadcc

app/models/rosters/self_roster_options_query.rb
plienta: resolved in 4f7115e1ce12f360ffea63a991bf0e31dd0fadcc

app/models/registration/campaign.rb
plienta: resolved in 4f7115e1ce12f360ffea63a991bf0e31dd0fadcc

config/locales/roster/de.yml
plienta: resolved in be97f99d31a8a2250d10d290f64132145baea312

app/models/registration/campaign/campaign_details_service.rb
plienta: resolved in be97f99d31a8a2250d10d290f64132145baea312

spec/controllers/user_registrations_spec.rb
plienta: resolved in fd9c7915cb4f144b9bfd46db72e8db6fab303980

spec/helpers/registration/user_registrations_helper_spec.rb
plienta: resolved in 4634e431b59a3420240adcb133fa6badc258d99f

app/models/registration/user_registration.rb
plienta: In FCFS campaigns users can register for multiple options within the same campaign, depending on the type. For example, they can select multiple cohorts, but only one tutorial or talk. So if a campaign has 2 tutorials and 2 cohorts, a user should be able to register for 1 tutorial and 2 cohorts at the same time. The previous uniqueness UnrankedIndex constraint on `user_id + campaign` would block this behavior. 
So I update this constraint a little, enforce the checking only happens for Tutorial and Talk, not for Cohort. However the new constraint can only be implemented on entity level, not in DB level

db/migrate/20260501000000_remove_unique_unranked_index_from_registration_user_registrations.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveUniqueUnrankedIndexFromRegistrationUserRegistrations < ActiveRecord::Migration[8.0]
+  def change
+    remove_index :registration_user_registrations, name: "index_reg_user_regs_unique_unranked"
```
plienta: please refer to https://github.com/MaMpf-HD/mampf/pull/1052#discussion_r3129755452

app/controllers/roster/self_materialization_controller.rb
```diff
@@ -0,0 +1,150 @@
+module Roster
+  class SelfMaterializationController < ApplicationController
+    class RosterLockedError < StandardError; end
+    class RosterFullError < StandardError; end
+    class SelfAddNotAllowedError < StandardError; end
+    class SelfRemoveNotAllowedError < StandardError; end
+    class UserNotFoundError < StandardError; end
+
+    ALLOWED_ROSTERABLE_TYPES = ["Tutorial", "Talk", "Cohort", "Lecture"].freeze
+
```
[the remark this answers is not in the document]
plienta: fixed in e35faf1bbf51eb5077c76e3f67cba8c97d83e1b5

spec/requests/roster/self_materialization_controller_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        let(:talk) do
+          create(:talk, lecture: seminar, capacity: 1, self_materialization_mode: "add_and_remove")
+        end
+
+        it("allows a user to self-add to a talk if not full") do
+          post self_add_talk_path(talk,
+                                  params: { type: "Talk",
+                                            partial: "talks/talker",
+                                            variable: "talk" })
+          expect(talk.allocated_user_ids).to include(student.id)
```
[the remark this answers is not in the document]
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55

config/locales/roster/de.yml
```diff
@@ -267,6 +267,9 @@ de:
...
       user_already_in_bundle: "Studierende/r ist bereits %{group} zugewiesen."
       item_locked: "Diese Gruppe wird durch ein aktives Anmeldeverfahren verwaltet und kann nicht manuell bearbeitet werden."
+      self_add_not_allowed: "Die Selbstanmeldung zu diesem %{type} ist für Studierende nicht erlaubt."
+      self_remove_not_allowed: "Das selbstständige Verlassen dieses %{type} ist für Studierende nicht erlaubt."
+      campaign_associated: "Diese Gruppe ist Teil eines Anmeldeverfahrens und kann nicht manuell verwaltet werden."
       campaign_associated: "Selbst-Einschreibung ist nicht verfügbar, da diese Gruppe Teil eines Anmeldeverfahrens ist."
```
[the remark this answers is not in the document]
plienta: resolved in e35faf1bbf51eb5077c76e3f67cba8c97d83e1b5

config/locales/roster/en.yml
```diff
@@ -263,6 +263,9 @@ en:
       user_not_found: "Student not found."
       user_already_in_bundle: "Student is already assigned to %{group}."
       item_locked: "This group is managed by an active registration campaign and cannot be modified manually."
+      self_add_not_allowed: "Self-enrollment in this %{type} is not permitted for students."
+      self_remove_not_allowed: "Self-leaving from this %{type} is not permitted for students."
```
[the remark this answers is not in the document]
plienta: reject, as I want to highlight the  "self"

app/controllers/roster/self_materialization_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+    private
+
+      def ensure_rosterable_unlocked!
+        raise(RosterLockedError) if @rosterable.locked?
+      end
+
+      def ensure_rosterable_not_full!
+        raise(RosterLockedError) if @rosterable.full?
+      end
```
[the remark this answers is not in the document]
plienta: resolved in e35faf1bbf51eb5077c76e3f67cba8c97d83e1b5

config/locales/roster/en.yml
```diff
@@ -263,6 +263,9 @@ en:
...
+      campaign_associated: "This group is part of a registration campaign and cannot be managed manually."
```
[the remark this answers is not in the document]
plienta: resolved in e35faf1bbf51eb5077c76e3f67cba8c97d83e1b5

app/models/rosters/rosterable.rb
```diff
@@ -61,6 +61,24 @@ def locked?
...
+    def config_allow_self_remove?
+      ["remove_only", "add_and_remove"].include?(self_materialization_mode)
+    end
+
+    # guard in FE for self-removal possibility
+    def allow_self_remove?(user)
+      config_allow_self_remove? && !locked? && allocated_user_ids.include?(user.id)
+    end
```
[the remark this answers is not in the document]
plienta: this part has been considerably rewritten in 4f7115e1ce12f360ffea63a991bf0e31dd0fadcc

spec/requests/roster/self_materialization_controller_spec.rb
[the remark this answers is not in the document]
plienta: resolved in 88148377348a556b2669c1a2a7c0d87d1e20aa55

app/controllers/roster/self_materialization_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+      ensure_rosterable_allow_self_remove!
+
+      user = current_user
+      Rosters::MaintenanceService.new.remove_user!(user, @rosterable)
+
+      flash.now[:notice] = t("roster.messages.user_removed")
+
+      render_user_update(params[:turbo_frame],
+                         params[:partial],
+                         { params[:variable].to_sym => @rosterable })
```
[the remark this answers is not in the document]
plienta: move logic into a service in e35faf1bbf51eb5077c76e3f67cba8c97d83e1b5

app/controllers/roster/self_materialization_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+    end
+
+    rescue_from SelfRemoveNotAllowedError do
+      respond_with_error(t("roster.errors.self_remove_not_allowed",
+                           type: @rosterable.class.model_name.human))
+    end
+
+    def current_ability
+      @current_ability ||= LectureAbility.new(current_user)
+    end
```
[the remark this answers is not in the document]
plienta: added `authorize_lecture` in e35faf1bbf51eb5077c76e3f67cba8c97d83e1b5

app/models/registration/user_registration.rb
fosterfarrell9: This makes sense but the solution in [cc4b388](https://github.com/MaMpf-HD/mampf/pull/1052/commits/cc4b3889c015144649b308dcbee6520b4ad213e5) seems not quite correct:

```ruby
# FCFS campaigns: one row per user+campaign for tutorial + talk items
# NOTE: DB cannot enforce this via nested conditions/subqueries in indexes
# (so enforcement is handled partially in Rails)
validates :user_id,
          uniqueness: {
            scope: :registration_campaign_id
          },
          if: lambda {
            registration_campaign&.first_come_first_served? &&
              registration_item&.registerable_type.in?(["Tutorial", "Talk"])
          }
```

From my understanding of this validation, this will have the effect that if you register for a cohort and then for a tutorial (which should be allowed), then the registration for the tutorial will be blocked because the check looks for any existing registration.

So I guess t he code needs to be a bit more complex. I would suggest to rather go with a custom validation using a private method instead of using the built-in uniqueness validator.

Also: Once you have this, please add unit tests that verify correct behaviour (and remove the stale unit tests for the old validation).

Now that I think more about it, I guess a database level backed validation could actually work and I would argue that it makes sense to use one (since students could indeed try to write scripts to "mass register" and try to exploit a race condition). We could do the following in the database (add a boolean flag on the user registration that contains whether the registration item is an "exclusive one" and a uniqueness index for that)

```ruby
add_column :registration_user_registrations,
           :exclusive_assignment,
           :boolean,
           null: false,
           default: false

add_index :registration_user_registrations,
          [:registration_campaign_id, :user_id],
          unique: true,
          where: "exclusive_assignment",
          name: "index_reg_user_regs_unique_exclusive_assignment"
```

and in the UserRegistration model

```ruby
before_validation :set_exclusive_assignment
```

where we have this private method

```ruby
def set_exclusive_assignment
  self.exclusive_assignment =
    registration_campaign&.first_come_first_served? &&
    registration_item&.exclusive_assignment?
end
```

which would need this new method in the registration item class

```ruby
def exclusive_assignment?
  registerable.class.exclusive_assignment?
end
```

and corresponding methods for `Tutorial`, `Talk` and `Cohort`.

Splines: > if you register for a cohort and then for a tutorial (which should be allowed), then the registration for the tutorial will be blocked because the check looks for any existing registration.

that's indeed a good point!

>  I would suggest to rather go with a custom validation

I agree, would probably be a lot nicer to read, and especially if it's getting more complex this is needed.

> Also: Once you have this, please add unit tests that verify correct behaviour (and remove the stale unit tests for the old validation).

Yes, we definitely need unit tests for this.

> Now that I think more about it, I guess a database level backed validation could actually work

I like this approach a lot. Additionally, we should also have unit tests which validate (not on DB-level but at least in Rails) that there can exist at most 3 preference-based registrations.

Splines: Please also refer to Copilot's comment here: https://github.com/MaMpf-HD/mampf/pull/1052#discussion_r3030540827

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -21,8 +26,118 @@ def destroy_for_user
...
+      locals = Registration::Campaign::CampaignDetailsService.new(@campaign, current_user)
+                                                             .preferences_info
+      rerender_preferences(locals)
+    end
+
     private
 
+      def render_not_found(exception)
+        render json: { error: exception.message }, status: :unprocessable_content
+      end
```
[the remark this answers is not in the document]
plienta: render_not_found has been removed

config/locales/roster/de.yml
fosterfarrell9: `i18n-tasks missing` now display this key as missing:

```
|  all   | buttons.add | app/frontend/registration/main/pb/_options_table.html.erb:37
```

app/frontend/registration/main/pb/_options_table.html.erb
```diff
@@ -0,0 +1,57 @@
...
+              <div class="small text-muted">(<%= item_display_type(item) %>)</div>
+            </div>
+
+            <!-- Right: Add button -->
+            <button type="submit" name="add_<%= campaign.id %>"
+                    class="material-icons clickable playlist_add border-0 bg-transparent"
+                    <%= "disabled" if items_selected.include?(item.id) %>
+                    formaction="<%= add_preference_path(item.id) %>">
+              playlist_add
+            </button>
```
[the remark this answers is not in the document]
plienta: has been removed

app/frontend/registration/main/pb/_select_preference.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= form_with method: :post,
+              data: { turbo_frame: "preferences_workspace",
```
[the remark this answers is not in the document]
plienta: resolved in afac8a7b3849af40d4dd4b4a16054f57796ab1d0

app/helpers/registration/user_registrations_helper.rb
```diff
@@ -0,0 +1,162 @@
...
+             badge_class: "bg-mdb-color-lighten-3 text-white" }
+    }.freeze
+    def get_mode_info(mode)
+      MODE_MAP.fetch(mode, MODE_MAP[-1])
+    end
+
+    # Examples of policy config:
+    # student_performance -> config: { certification_status: :pending }
+    # institutional_email	-> config:	{ allowed_domains: ["uni-heidelberg.de "] }
+    # prerequisite_campaign	-> config:	{ prerequisite_campaign_id: 42 }
```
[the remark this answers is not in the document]
plienta: resolved in 2b74469d96059c241082b514ee3aec3b05e7974a

app/helpers/roster_helper.rb
```diff
@@ -112,4 +112,59 @@ def roster_group_badge(group, _group_type)
...
+      base_type = t("registration.item.types.other_group")
+
+      if cohort.propagate_to_lecture
+        base_type
+      else
+        icon = tag.i(class: "bi bi-person-x ms-1",
+                     style: "color: #495057;",
+                     data: { bs_toggle: "tooltip",
+                             bs_title: t("registration.item.hints.no_propagation") })
+        safe_join([base_type, " ", icon])
```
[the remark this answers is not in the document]
plienta: resolved in afac8a7b3849af40d4dd4b4a16054f57796ab1d0

app/helpers/roster_helper.rb
```diff
@@ -112,4 +112,59 @@ def roster_group_badge(group, _group_type)
...
+      end
+    end
+  end
+
+  def format_date(time)
+    return "" if time.nil?
+
+    time.strftime("%b %d, %H:%M")
+  end
+  SELF_ROSTER_TABLE_CONFIG = {
```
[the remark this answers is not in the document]
plienta: removed in 6541218da450b62acf37b28f24d853666912721a

app/frontend/registration/touched_check/touched_check.controller.js
```diff
@@ -0,0 +1,39 @@
...
+
+  handleMarkState(event) {
+    const [typeAction, campaignId] = event.submitter.name.split("_");
+    if (typeAction === "reset" || typeAction === "save") {
+      this.reset(campaignId);
+    }
+    else if (typeAction === "up" || typeAction === "down"
+      || typeAction === "remove" || typeAction === "add") {
+      this.mark(campaignId);
+    }
```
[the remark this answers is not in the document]
plienta: resolved in cc14531b7a98bcefc2dfba43b7ad70cfe8ceafee

app/models/registration/user_registration/lecture_fcfs_edit_service.rb
```diff
@@ -0,0 +1,65 @@
...
+          errors = validate_register(item)
+          return Result.new(false, errors) unless errors.empty?
+
+          Registration::UserRegistration.create!(
+            registration_campaign: @campaign,
+            registration_item: item,
+            user: @user,
+            status: :confirmed
+          )
+        end
```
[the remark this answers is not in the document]
plienta: add index for uniqueness in f6b1aceafe323a84174853f2efcfee67d5d2cba4

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -21,8 +25,114 @@ def destroy_for_user
...
+                           .new(@lecture, current_user)
+                           .call
+      @rosterized_entries = Rosters::StudentMaterializedResultResolver
+                            .new(current_user)
+                            .all_rosterized_for_lecture(@lecture)
+      @self_rosterables = Rosters::SelfRosterOptionsQuery.new(@lecture, current_user)
+                                                         .call
+      render template: "registration/main/index",
+             layout: turbo_frame_request? ? "turbo_frame" : "application"
+    end
```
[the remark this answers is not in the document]
plienta: resolve in aea9e90a66925b91f1f3c51510a3d074b25e6824

config/locales/roster/de.yml
plienta: resolve in aea9e90a66925b91f1f3c51510a3d074b25e6824

app/controllers/registration/user_registrations/campaign_details_service.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      def results_roster
+        items_selected = @campaign.registration_items
+                                  .includes(:user_registrations)
+                                  .where(user_registrations: { user_id: @user.id })
+        items_succeed = Rosters::StudentMaterializedResultResolver
+                        .new(@user).succeed_items(@campaign)
+        status_items_selected = items_selected.each_with_object({}) do |i, hash|
+          hash[i.id] = items_succeed.pluck(:id).include?(i.id) ? "confirmed" : "dismissed"
+        end
```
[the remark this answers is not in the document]
plienta: resolve in  aea9e90a66925b91f1f3c51510a3d074b25e6824

app/models/registration/user_registration.rb
plienta: I added `index_reg_user_regs_unique_exclusive_assignment_unranked` following your suggestion in commit f6b1aceafe323a84174853f2efcfee67d5d2cba4, so that a user can only have one unranked registration per campaign for exclusive registerables (Tutorial/Talk).

In the same commit, I also added `index_reg_user_regs_unique_item_user` to prevent duplicate rows for the same campaign + user + item combination.

Unit tests are included in commits f6b1aceafe323a84174853f2efcfee67d5d2cba4, 33aa549191306a5b040a9040e04285bd6c5e8dad, and e2b30e4dbbc8b5f9ce429c33bb0cdb0a1b1ad426.

app/controllers/roster/self_materialization_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+        end
+      end
+
+      def set_rosterable
+        unless Rosters::Rosterable::TYPES.include?(params[:type])
+          redirect_to root_path, alert: t("roster.errors.invalid_type")
+          return
+        end
+
+        klass = params[:type].constantize
```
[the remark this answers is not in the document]
Splines: Oh good point, fixed in ed529bcfc.

fosterfarrell9:
I found two issues in the student registration flow and fixed them in [1439cae](https://github.com/MaMpf-HD/mampf/pull/1052/commits/1439cae07a5580a45afea0d84c69897707e5b0a2)  and [a4baeec](https://github.com/MaMpf-HD/mampf/pull/1052/commits/a4baeecc17998400f29d85fc93b8aa87c50c24bd). The student registration actions `index`, `create`, `destroy`, and `add` lacked server-side ability checks, so users intentionally excluded from the student flow, such as teachers, tutors, or editors, could still call them directly. I fixed that by adding explicit authorization and denial request specs, and also fixed a separate withdraw bug where `campaign_id` from the route could differ from the item’s actual campaign, causing validation against one campaign while deleting from another.

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -21,8 +26,112 @@
...
+                                 .all_rosterized_for_lecture(student_registration_lecture),
+            lecture: student_registration_lecture,
+            user: current_user
+          ).render_in(view_context)
+        )
+      end
+
+      def preference_params
+        raw_preferences = params.fetch(:preferences, {})
+        return raw_preferences.permit!.to_h if raw_preferences.respond_to?(:permit!)
```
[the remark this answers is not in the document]
Splines: fixed in 6a209b25e658

e2e/_support/fixtures.ts
```diff
@@ -73,3 +73,8 @@ export const test = base.extend<MaMpfFixtures>({
     await use(new FactoryBot(request));
   },
 });
+
+test.beforeEach(async ({ request }) => {
+  // Clean database before every test (brutal, but effective for good test isolation)
+  await callBackend(request, "database_cleaner", {});
+});
```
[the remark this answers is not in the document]
Splines: oh good point, fixed in 73e035a5

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -104,5 +215,22 @@ def destroy_registrations(registrations)
...
         end
       end
+
+      def set_item
+        @item = Registration::Item.find(params[:item_id])
+        return if @item
+
+        respond_with_flash(:alert, t("registration.item.not_found"),
+                           fallback_location: root_path)
+      end
```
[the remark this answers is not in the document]
Splines: good point, fixed in 6c6e823baa

app/controllers/user_registrations/campaign_details_service.rb
```diff
@@ -0,0 +1,75 @@
...
+        next unless policy_result[:kind] == "prerequisite_campaign"
+
+        id = policy_result[:config]["prerequisite_campaign_id"]
+        campaign = Registration::Campaign.find_by(id: id)
+        policy_result[:config]["prerequisite_campaign"] =
+          if campaign
+            "#{campaign&.campaignable&.title}: #{campaign&.description}"
+          else
+            I18n.t("registration.user_registration.messages.campaign_not_found")
+          end
```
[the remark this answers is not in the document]
Splines: oh, that key was not even existing, fixed in 9c248a7a7b

config/locales/registration/en.yml
```diff
@@ -259,6 +272,87 @@ en:
...
         remove_user: "Remove user from registration process"
+        rank_option_tooltip: "Set as %{rank} preference"
+      registrations: "Registrations"
+      title: "Your Registration"
+      planning: "Planning"
+      course_code: "Course code"
+      campaign_main: "Registration process"
+      eligibility_title: "Eligibility"
+      fcfs_instruction: "Melde Dich für eine Gruppe an. Die endgültige Zuteilung erfolgt nach der Anmeldefrist."
+      fcfs_instruction: "Register for a group. Final allocation will automatically take place after the registration deadline."
```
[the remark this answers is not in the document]
Splines: fixed in 34a7ccea375

Splines:
@coderabbitai review

app/frontend/user_registrations/eligibility_helper.rb
```diff
@@ -0,0 +1,117 @@
+module EligibilityHelper
+  POLICY_CONFIG_UNAVAILABLE = "No configuration available".freeze
```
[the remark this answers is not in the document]
Splines: This is just a fallback for something that should never happen, no need to translate this.

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -21,8 +26,117 @@ def destroy_for_user
...
+        )
+      end
+
+      def preference_params
+        raw_preferences = params.fetch(:preferences, {})
+        allowed_keys = preference_param_keys
+        return raw_preferences.permit(*allowed_keys).to_h if raw_preferences.respond_to?(:permit)
+
+        raw_preferences.to_h.slice(*allowed_keys)
+      end
```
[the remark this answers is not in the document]
Splines: good catch, fixed in 1320c5fd9a

app/frontend/lectures/show/_sidebar.html.erb
```diff
@@ -88,7 +102,7 @@
   <% if lecture %>
     <%# Müsli (external link) %>
     <li class="sidebar-item nav-link <%= !lecture.muesli ? "disabled" : ""%>">
-      <a href="<%= DefaultSetting::MUESLI_LINK %>" target="_blank" %>
+      <a href="<%= DefaultSetting::MUESLI_LINK %>" target="_blank">
```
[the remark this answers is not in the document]
Splines: fixed in abe639f6c

app/frontend/registration/campaigns/_card_header.html.erb
```diff
@@ -48,7 +48,7 @@
         </small>
 
         <div>
-          <%= l(campaign.registration_deadline, format: :short) if campaign.registration_deadline %>
+          <%= l(campaign.registration_deadline, format: :short) %>
```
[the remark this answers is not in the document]
Splines: fixed in a4bb349898b537

app/frontend/user_registrations/eligibility_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    def policy_config_info(policy)
+      case policy[:kind]
+      when "student_performance"
+        policy.dig(:config, "certification_status").to_s.capitalize
+      when "institutional_email"
+        Array(policy.dig(:config, "allowed_domains")).join(", ")
+      when "prerequisite_campaign"
+        policy.dig(:config, "prerequisite_campaign")
```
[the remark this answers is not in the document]
Splines: fixed in 83567fd00270

spec/requests/roster/self_materialization_controller_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+    create(:tutorial,
+           lecture: lecture,
+           skip_campaigns: false)
+  end
+  let(:campaign) { create(:registration_campaign, campaignable: lecture) }
+  let(:item) do
+    create(:registration_item,
+           registration_campaign: campaign,
+           registerable: camp_tutorial)
+  end
```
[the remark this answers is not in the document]
Splines: no, the branch doesn't depend on `item` being created or not. Bot with `let` and `let!`, the tests will past. so won't fix

app/models/user_registrations/preferences_handler.rb
```diff
@@ -0,0 +1,35 @@
...
+                                   .where(user_id: user.id)
+                                   .where(status: [:confirmed, :pending])
+      user_registrations.includes(:registration_item)
+                        .map(&:registration_item)
+                        .flatten
+                        .sort_by { |i| i.preference_rank(user) }
+                        .map do |item|
+        ItemPreference.new(item,
+                           item.preference_rank(user))
+      end
```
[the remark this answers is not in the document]
Splines: oh good catch, fixed in 4c49ac9e9

app/models/user_registrations/lecture_preference_edit_service.rb
```diff
@@ -0,0 +1,68 @@
...
+  # with preference policy.
+  class LecturePreferenceEditService < Handler
+    def update!(pref_items)
+      ActiveRecord::Base.transaction do
+        errors = validate_update(pref_items)
+        return Result.new(false, errors) unless errors.empty?
+
+        @campaign.user_registrations.where(user_id: @user.id).destroy_all
+        pref_items.each do |pref_item|
+          item = Registration::Item.find(pref_item.id)
```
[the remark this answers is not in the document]
Splines: good catch, fixed in 8b85abd36

app/models/user_registrations/lecture_fcfs_edit_service.rb
```diff
@@ -0,0 +1,73 @@
...
+    # the downside effect is that we cannot trace the withdraw time of user
+    def withdraw!(item)
+      ActiveRecord::Base.transaction do
+        item.lock!
+        errors = validate_withdraw
+        return Result.new(false, errors) unless errors.empty?
+
+        registration = item.user_registrations.find_by!(user: @user, status: :confirmed)
+        registration.destroy!
+      end
```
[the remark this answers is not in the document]
Splines: fixed in 8e76947f2

app/frontend/user_registrations/user_registrations_helper.rb
```diff
@@ -0,0 +1,228 @@
...
+        user,
+        item.registerable_type
+      )
+    end
+
+    def gtile_icon_for(icon_name)
+      case icon_name
+      when "person"   then "bi-person"
+      when "location" then "bi-geo-alt"
+      end
```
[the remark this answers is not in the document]
Splines: fixed in 8492a101b

app/frontend/user_registrations/user_registrations_helper.rb
```diff
@@ -0,0 +1,228 @@
+module UserRegistrationsHelper
+  MODE_MAP = {
+    -1 => { mode_name: "Unknown", abbr: "UNK",
+            badge_class: "bg-grey-lighten-4 text-grey" },
+    0 => { mode_name: "First-come, first-served", abbr: "FCFS",
+           badge_class: "bg-light-blue-lighten-4 text-darkblue" },
+    1 => { mode_name: "Preference Based", abbr: "PB",
+           badge_class: "bg-mdb-color-lighten-3 text-white" }
```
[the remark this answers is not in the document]
Splines: This mode map was not even used anymore, so I removed it in 03d45a85b75085

app/controllers/user_registrations/lecture_first_come_first_served_edit_service.rb
```diff
@@ -0,0 +1,73 @@
...
+        return Result.new(false, errors) unless errors.empty?
+
+        Registration::UserRegistration.create!(
+          registration_campaign: @campaign,
+          registration_item: item,
+          user: @user,
+          status: :confirmed
+        )
+      end
+      Result.new(true, [])
```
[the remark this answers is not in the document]
Splines: good catch, fixed in c35b533d329:

> without serialization, two concurrent `register!` calls on different tutorial/talk items for the same user and campaign can both pass validation, then the loser raises `ActiveRecord::RecordNotUnique` on insert. That still bypasses the service `Result` contract.

.github/copilot-instructions.md
```diff
@@ -48,18 +48,16 @@ For the filename of migration files, always use the current date, but all-zeros
...
-
-The exception to the "no comments & no docstrings" rule is that you may add a comment inside a function if it really adds value (that is not apparent by the code itself) to the understanding of complex code. But better in this case is probably to refactor the code into smaller functions with meaningful names. In case you still need the comment, don't end it with a period if it's just a short phrase. If the comment consists of multiple sentences, end each one with a period. Break comments at 80 chars hard limit.
-
-Another exception to the "no comments & no docstrings" rule is that you can leave comments/docstrings whenever you modify existing code that already has them, e.g. when you refactor a function.
+Never delete any existing comments. For newly generated code, prefer self-explanatory code with docstrings over individual line comments.
```
fosterfarrell9, line 51: @Splines Just to be sure since were an adamant defender of the "no comment" rule: This is a change you want? Feel free to resolve. Just wanted to be sure.

app/abilities/lecture_ability.rb
```diff
@@ -34,5 +34,19 @@ def initialize(user)
...
     end
+
+    can [:self_materialize, :enroll], Lecture do |lecture|
+      student_registration_participant?(user, lecture)
+    end
   end
+
+  private
+
+    def student_registration_participant?(user, lecture)
```
fosterfarrell9: This is verbatim the same method as in `registration_user_registration_ability`. Maybe it could be beneficial to DRY this up in order to avoid divergence when one of these gets  changed (and the othe other forgotten).

app/abilities/registration_user_registration_ability.rb
```diff
@@ -7,5 +7,19 @@ def initialize(user)
...
+
+    can [:index, :create, :destroy, :add], Lecture do |lecture|
+      student_registration_participant?(user, lecture)
+    end
   end
+
+  private
+
+    def student_registration_participant?(user, lecture)
+      lecture.visible_for_user?(user) &&
```
fosterfarrell9: Seem my comment in `lecture_ability`.

app/controllers/user_registrations/lecture_fcfs_edit_service.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    private
+
+      # Validation for creating registration in lecture based registration
+      # 0. Check open for registration
+      # 1. Check if user has already registered for this campaign this group type
+      # 2. Check if item still has capacity
+      # 3. Check if user satisfies all policies (phase: registration and both)
+      # 4. Check if items are valid for this campaign
+      # 5. Check if campaign is in FCFS mode
```
fosterfarrell9: Maybe nitpicking here but maybe dont use 0., 1. etc when the numbering doesn't refer to the numbering in the array.

app/controllers/user_registrations/lecture_preference_edit_service.rb
```diff
@@ -0,0 +1,68 @@
...
+            preference_rank: pref_item.rank
+          )
+        end
+      end
+      Result.new(true, [])
+    end
+
+    private
+
+      # Validation for creating registration in lecture based registration
```
fosterfarrell9: See my comment in `lecture_fcfs_edit_service`

app/controllers/user_registrations/preferences_handler.rb
```diff
@@ -0,0 +1,37 @@
+module UserRegistrations
+  class PreferencesHandler
+    MAX_PREFERENCES = 3
+
+    # struct for store temporary preference items during modification process,
+    # will be used in form
+    SimpleItemPreference = Struct.new(:id, :rank)
+
+    # struct for render preference items in FE
+    ItemPreference       = Struct.new(:item, :rank)
```
fosterfarrell9: Is this really what rubocop wants (this tab before =). I wonder why.

app/frontend/user_registrations/preference_based/_options_table.html.erb
```diff
@@ -0,0 +1,59 @@
+<div>
```
fosterfarrell9, line 1: I don't like t hat the folder is called `pb`. This is not an intuitive name.What does it stand for? I would prefer a more descriptive name.

app/frontend/user_registrations/eligibility_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+          "registration.user_registration.eligibility_failures.institutional_email_mismatch_html",
+          {
+            domains: policy_config_info(policy),
+            profile: link_to(t("navbar.profile"), edit_profile_path,
+                             class: "fw-semibold", target: "_top")
+          }
+        ]
+      when "configuration_error"
+        [
+          "registration.user_registration.eligibility_failures.institutional_email_configuration_error", # rubocop:disable Layout/LineLength
```
fosterfarrell9: Manual line break  via "... " \ "..." would be a valid alternative over disabling the cop here.

app/models/rosters/rosterable.rb
```diff
@@ -61,6 +72,33 @@ def locked?
       !in_completed_campaign?
     end
 
+    def config_allow_self_add?
+      self_materialization_mode_add_only? ||
+        self_materialization_mode_add_and_remove?
+    end
+
+    # guard in FE for self-assignment possibility
```
fosterfarrell9: What is `FE`? Without knowing what `FE`is the comment becomes hard to decipher

app/models/rosters/rosterable.rb
```diff
@@ -61,6 +72,33 @@ def locked?
...
+
+      !full?
+    end
+
+    def config_allow_self_remove?
+      self_materialization_mode_remove_only? ||
+        self_materialization_mode_add_and_remove?
+    end
+
+    # guard in FE for self-removal possibility
```
fosterfarrell9: What is `FE`

fosterfarrell9 (changes requested):
Made some nitpick comments.

.github/copilot-instructions.md
Splines: Haha yeah 😅 It turned out AI frequently removed existing comments due to this, which was a bit annoying. I'm still against too many comments in the code and prefer docstrings over individual line comments. Let's see how that goes with these updated instructions ;)

app/controllers/user_registrations/preferences_handler.rb
Splines: Fixed in 01d782791.

This can be flagged as error by setting the following config:

```yml
# .rubocop.yml
Layout/ExtraSpacing:
  AllowForAlignment: false
```

(which was by default `true`). As this would flag many errors in the codebase, I don't do this in this PR.

app/controllers/user_registrations/lecture_preference_edit_service.rb
Splines: The comment didn't add any value, so I just removed it completely in 55b30428f95

app/controllers/user_registrations/lecture_fcfs_edit_service.rb
Splines: The comment didn't add any value, so I just removed it completely in 55b30428f95

app/abilities/registration_user_registration_ability.rb
Splines: fixed in 79c6559ea

app/abilities/lecture_ability.rb
Splines: oh good catch, fixed in 79c6559ea

app/frontend/user_registrations/preference_based/_options_table.html.erb
Splines: indeed, fixed in bb0a8cb9fa4

app/frontend/user_registrations/eligibility_helper.rb
Splines: good point, fixed in 5f310930971b

app/models/rosters/rosterable.rb
Splines: Maybe Lien meant "frontend" with it? Removed it in bb0a8cb9fa4.

app/models/rosters/rosterable.rb
Splines: Maybe Lien meant "frontend" with it? Removed it in bb0a8cb9fa4.

fosterfarrell9 (dismissed):
LGTM

Splines:
@coderabbitai full review

app/controllers/user_registrations/lecture_preference_edit_service.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      # Each available rank must be covered exactly once by a distinct item.
+      def check_preferences(pref_items)
+        ranks = pref_items.map(&:rank).sort
+        item_ids = pref_items.map(&:id)
+        return if ranks == (1..expected_preference_count).to_a &&
+                  item_ids.uniq.size == item_ids.size
+
+        I18n.t("registration.user_registration.messages.invalid_preferences")
+      end
```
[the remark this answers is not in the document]
Splines: However, we have changed that along the way and want that exactly the expected_preference_count number of options are given in each request. So the code is exactly what we try to do, thus I won't tackle this request.

app/frontend/roster/components/group_tile_component.html.erb
```diff
@@ -1,143 +1,226 @@
...
+                tutorial-gtile-sm-menu
+              "
+            >
+              <li>
+                <h6 class="dropdown-header">
+                  <%= t("roster.self_materialization.label",
                       default: "Self-Enrollment") %>
-              </h6>
-            </li>
+                </h6>
```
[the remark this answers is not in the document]
Splines: fixed in 2d975c1a35

app/frontend/user_registrations/components/rosterized_entries_component.rb
```diff
@@ -0,0 +1,245 @@
...
+  # "confirmed cases" notice when no preference-based messaging applies.
+  def notice_messages
+    return [unassigned_notice] if rosterized_entries.blank?
+
+    messages = rosterized_entries.filter_map do |rosterable|
+      preference_notice_message(rosterable)
+    end
+
+    messages.presence || [default_notice]
+  end
```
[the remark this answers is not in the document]
Splines: This is more something for another PR if we experience slow-downs due to this.

app/helpers/registration/campaigns_helper.rb
```diff
@@ -112,5 +112,9 @@ def allocate_campaign_button(campaign, size: nil)
                 class: classes,
                 form: { data: form_data })
     end
+
+    def closed_early?(campaign)
+      !campaign.open_for_registrations? && campaign.registration_deadline > Time.current
+    end
```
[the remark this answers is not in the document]
Splines: It is enforced at model and database level that `registration_deadline` is never null, so this is not a problem.

app/models/rosters/self_roster_availability.rb
```diff
@@ -0,0 +1,24 @@
...
+    def initialize(lecture, user)
+      @lecture = lecture
+      @user = user
+    end
+
+    def blocked_by_unremovable_assignment?
+      rosterized_entries.any? do |rosterable|
+        rosterable.user_allocated?(@user) && rosterable.config_allow_self_add? &&
+          !rosterable.config_allow_self_remove?
+      end
```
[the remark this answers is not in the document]
Splines: oh, good point, fixed in ca5d34269

db/migrate/20260501000001_add_exclusive_assignment_to_registration_user_registrations.rb
```diff
@@ -0,0 +1,20 @@
...
+    add_index :registration_user_registrations,
+              [:registration_campaign_id, :user_id],
+              unique: true,
+              where: "exclusive_assignment = true AND preference_rank IS NULL",
+              name: "index_reg_user_regs_unique_exclusive_assignment_unranked"
+
+    add_index :registration_user_registrations,
+              [:registration_campaign_id, :user_id, :registration_item_id],
+              unique: true,
+              name: "index_reg_user_regs_unique_item_user"
```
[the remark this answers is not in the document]
Splines: not a problem as this is not yet deployed to production

e2e/user_registration/campaign_registration.spec.ts
```diff
@@ -0,0 +1,599 @@
...
+      factory,
+      lecture,
+      "first_come_first_served",
+      "Tutorial registration",
+    );
+
+    await new CampaignRegistrationPage(student.page, lecture.id).goto();
+
+    const firstTile = student.page.locator(".tutorial-gtile").first();
+    await firstTile.getByRole("button", { name: "Register now" }).click();
```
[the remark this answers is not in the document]
Splines: fixed in 0fb21e819e0

spec/controllers/user_registrations/lecture_first_come_first_served_edit_service_spec.rb
```diff
@@ -0,0 +1,337 @@
...
+      expect(registration.registration_item).to eq(item)
+      expect(registration.status).to eq("confirmed")
+    end
+
+    it "fully deletes a registration on withdraw" do
+      service = described_class.new(campaign, user)
+      service.register!(item)
+      service.withdraw!(item)
+      registration = Registration::UserRegistration.first
+      expect(registration).to be_nil
```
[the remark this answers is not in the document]
Splines: oh good point, fixed in fcd9171d35509a4

app/frontend/user_registrations/user_registrations_helper.rb
```diff
@@ -0,0 +1,224 @@
...
+
+  def registration_blocked_by_unremovable_assignment?(lecture)
+    return false if lecture.blank?
+    return @registration_blocked_by_unremovable_assignment \
+      unless @registration_blocked_by_unremovable_assignment.nil?
+
+    @registration_blocked_by_unremovable_assignment =
+      Rosters::SelfRosterAvailability.new(lecture, current_user)
+                                     .blocked_by_unremovable_assignment?
+  end
```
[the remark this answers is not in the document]
Splines: won't fix as this is only used for the student registration page which is by design specific to the lecture, so what you describe cannot actually occur

spec/requests/roster/self_materialization_controller_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+      post self_add_tutorial_path(tutorial), as: :turbo_stream
+
+      expect(response.body).to include('target="student_registration_rosterized_entries"')
+    end
+
+    it "updates the self materialization zone turbo frame" do
+      post self_add_tutorial_path(tutorial), as: :turbo_stream
+
+      expect(response.body).to include('target="self_materialization_zone"')
+    end
```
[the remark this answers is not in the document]
Splines: fixed in 6dd1c121db5c

app/helpers/roster_helper.rb
```diff
@@ -112,4 +103,86 @@ def roster_group_badge(group, _group_type)
...
+        cell_class: "text-end",
+        icon: "looks_one",
+        field: ->(rosterable) { rosterable.position } },
+      { header: "basics.description",
+        icon: "description",
+        cell_class: "text-center",
+        field: ->(rosterable) { rosterable.description } },
+      { header: "basics.date",
+        icon: "event",
+        field: lambda { |rosterable|
```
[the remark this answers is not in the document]
Splines: fixed in 19756ec0e4

fosterfarrell9 (approved):
Huge chunk of work, awesome.

PR #1054, opened 2026-02-18
Update dependencies

Splines:
As usual and especially since there are some CVEs.

- I've also added `.env` to the gitignore file.
- RuboCop still has weird linting problems in the newest version `1.84.2`, so I leave it pinned at version `1.81.7`. See also https://github.com/rubocop/rubocop/issues/14946
- `faraday` was only used for Erdbeere, which is not part of MaMpf anymore, so we remove the gem here.


fosterfarrell9 (commented):
LGTM. By the way, I don't think we need the faraday gem any more. It was used for communication with the now-gone erdbeere. I see no other use if it any longer. Maybe we could get rid of it.

fosterfarrell9 (approved):
LGTM.

PR #1076, opened 2026-03-24
Update dependencies (Release v1.20.3)

Splines:
We update our dependencies as there were many CVEs in the last time.

With regards to RuboCop, we still stay on `1.81.7` since newer versions break some method indentation for us, see [this comment](https://github.com/rubocop/rubocop/issues/14916#issuecomment-4116165665).


fosterfarrell9 (approved):
LGTM!

PR #1078, opened 2026-04-04
Update dependencies (Release 1.20.4)

Splines:
Update Gems and npm packages. By default, we update their minor versions according to our lock files.

Additionally:

### Gems

I ran `bundle outdated` to also see where we should update major versions. There are quite a few entries, here I only tackle these ones:
- Dalli (from 3 to 5). See how we use it [in Rails](https://github.com/petergoldstein/dalli/wiki/Using-Dalli-with-Rails) with the `:mem_cache_store`. I removed a duplicate key from `docker.env` for this and also updated connection_pool (Dalli relies on connection_pool when used in conjunction with Rails). Florian pinned connection pool [in this commit](https://github.com/MaMpf-HD/mampf/commit/5ee6df53c6ab1b98e3a9e0c5bdded499d8265ebb), but this is already fixed by [this PR](https://github.com/rails/rails/pull/56292#issuecomment-3730324376).
- Puma (from 6 to 7). I didn't find an upgrade guide, so I just upgraded the version.

In development, I've upgraded:
- rspec-github (2 to 3) (RuboCop doesn't show any additional offenses)
- rspec-rails (6 to 8) (RuboCop doesn't show any additional offenses)
- i18n-tasks (I have unwantedly pinned to the hotfix version, now a bit looser to minor versions just like other dependencies do)

### npm

I ran `npm outdated` to also get to know if there are new major versions for our npm packages.

For development deps, I've upgraded:
- Typescript (5 to 6). We make use of TypeScript in our Playwright tests.
- Globals (15 to 17)
- ESLint (9 to 10). I've looked through the [Migration guide](https://eslint.org/docs/latest/use/migrate-to-10.0.0) and fixed some things.

### Off-topic

- I've also removed the `SOLR_PATH` variable in the env files on the servers as we don't use it anymore.
- Update Herb version in config file to reflect the version that comes bundled with the VSCode extension.


docker/production/docker.env
```diff
@@ -56,6 +55,7 @@ URL_HOST=mampf.mathi.uni-heidelberg.de
 MAMPF_PORT=3000
 
 # Fast caching and indexing need further services
+SOLR_PATH=/my/path/to/solr
```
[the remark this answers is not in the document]
Splines: ah good catch, then I should remove it from the the real production env files, will do that

Gemfile
```diff
@@ -18,10 +18,10 @@ gem "bootstrap_form", "~> 5.4"
 gem "cancancan", "~> 3.6"
 gem "coffee-rails", "~> 5.0"
 gem "commontator", "~> 7.0.1"
-gem "connection_pool", "< 3.0" # ActiveSupport currently breaks with connection_pool 3.x
+gem "connection_pool", "~> 3.0"
```
fosterfarrell9, line 21: Did we update ActiveSupport in the meantime such that this is now possible (sorry, didn't research this)?

Splines: ActiveSupport is not an extra dependency, it comes with pure Rails. With this PR, notice how we went from `8.0.4` to `8.0.5`, which includes the fix.

- Problem: https://github.com/rails/rails/issues/56786
- PR that fixed it for the `8.1` version of Rails: https://github.com/rails/rails/pull/56292
- Backport to `8.0`: https://github.com/rails/rails/commit/6a1207aed9df1df5e97f0e46504cdba5de91d0fe

I already built `mampf-experimental` and everything worked fine.

Gemfile.lock
```diff
@@ -430,20 +432,20 @@ GEM
...
-      actionmailer (= 8.0.4)
-      actionpack (= 8.0.4)
-      actiontext (= 8.0.4)
-      actionview (= 8.0.4)
-      activejob (= 8.0.4)
-      activemodel (= 8.0.4)
-      activerecord (= 8.0.4)
-      activestorage (= 8.0.4)
-      activesupport (= 8.0.4)
+    rails (8.0.5)
```
Splines, line 435: see here @fosterfarrell9

fosterfarrell9 (approved):
LGTM!

PR #1081, opened 2026-04-04
🚀 Müsli: Registration & Roster Management

fosterfarrell9:
This PR is a collector PR that includes all admin-facing stuff related to registration and roster management. It is an extension of https://github.com/MaMpf-HD/mampf/pull/1017 and supersedes it.

Regarding the default registration data: I set up a rake task `demo:setup` that will set the needed feature flags `registration_campaigns` and `roster_maintenance` and sets up one lecture and one seminar up with sveral groups and registrations, such that you can play around with it.

Since #1017 had a to do list, I copied this one here:

### TODO
- [x] Remove the `purpose` column from cohorts
- [x] Add columns for location and time to the tutorials table
- [x] Right before merging, update all the DB migration dates.
- [x] Right before merging, update versions of newly added Gems to the latest.
- [x] Html linting via herb
- [x] Generate default registration data for local development

### Won't do (here)
- [ ] Remove unused i18n keys -> use script to find unused keys
- [ ] Add link to feedback pane since the features are quite new. E.g. below the participants tab, but watch out for participants sideview in this case that might wrap below.

[Test Coverage report](https://github.com/MaMpf-HD/mampf/pull/1081#issuecomment-4189161081)


Splines:
@coderabbitai full review

db/migrate/20260404000005_add_unique_index_to_registration_items.rb
```diff
@@ -0,0 +1,12 @@
...
+    remove_index :registration_items,
+                 name: "index_registration_items_on_registerable"
+    remove_index :registration_items,
+                 name: "index_registration_items_uniqueness"
+
+    add_index :registration_items, [:registerable_type, :registerable_id],
+              unique: true,
+              name: "index_registration_items_on_unique_registerable"
+  end
+end
```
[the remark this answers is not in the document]
fosterfarrell9: The global uniqueness is intentional — the architecture doc specifies it just like that. Registerables belong to one campaign at a time. Cross-campaign sharing is a documented future extension that would require relaxing this index to per-campaign uniqueness when item-level capacity splitting is implemented.

app/models/registration/item.rb
```diff
@@ -22,15 +22,85 @@ class Item < ApplicationRecord
...
              class_name: "Registration::UserRegistration",
-             dependent: :destroy
+             foreign_key: :registration_item_id,
+             dependent: :destroy,
+             inverse_of: :registration_item
 
     validates :registerable_id,
               uniqueness: {
-                scope: [:registration_campaign_id, :registerable_type]
+                scope: :registerable_type
```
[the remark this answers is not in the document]
fosterfarrell9: This validation intentionally mirrors the global-uniqueness DB index added in the same PR. Both enforce the architecture's design rule that a registerable belongs to at most one campaign. Cross-campaign reuse is a documented future extension (item-level capacity splitting) — when that's needed, both the index and the validation would be relaxed together.

app/models/registration/policy.rb
```diff
@@ -27,68 +30,74 @@ class Policy < ApplicationRecord
...
+      val = config&.fetch("allowed_domains", nil)
+      return val if val.is_a?(String)
+
+      Array(val).join(", ")
     end
 
-    private
+    def allowed_domains_display
+      Array(config&.fetch("allowed_domains", [])).join(", @")
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed and added a dedicated spec for it.

app/models/registration/policy.rb
```diff
@@ -27,68 +30,74 @@ class Policy < ApplicationRecord
...
+                   when :institutional_email
+                     Registration::Policy::InstitutionalEmailHandler.new(self)
+                   when :prerequisite_campaign
+                     Registration::Policy::PrerequisiteCampaignHandler.new(self)
+                   when :student_performance
+                     Registration::Policy::Handler.new(self)
+                   else
+                     raise(ArgumentError, "Unknown policy kind: #{kind}")
       end
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Memoization was removed.

app/frontend/controllers/selectize.controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  connect() {
+    if (window.fillOptionsByAjax) {
+      window.fillOptionsByAjax($(this.element));
+    }
+  }
+}
```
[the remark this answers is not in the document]
fosterfarrell9: Fair enough. There were a few other with the same old naming pattern, I renamed them all and updated the import in initHotwire.js.

db/migrate/20260404000011_add_self_materialization_mode_to_rosterables.rb
```diff
@@ -0,0 +1,14 @@
+class AddSelfMaterializationModeToRosterables < ActiveRecord::Migration[8.0]
```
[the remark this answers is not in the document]
fosterfarrell9: Indeed, the last one used 7.2, fixed.

.herb.yml
```diff
@@ -23,12 +23,12 @@ linter:
     - "app/frontend/login/**/*.html.erb"
     - "app/frontend/media/**/*.html.erb"
     - "app/frontend/modal/**/*.html.erb"
-    - "app/frontend/roster/**/*.html.erb"
     - "app/frontend/search/**/*.html.erb"
     - "app/frontend/spinner/**/*.html.erb"
     - "app/frontend/vignettes/**/*.html.erb"
     # not excluded on purpose, since we want to try out herb on these files:
     # - app/frontend/registration/**/*.html.erb
+    - "!app/frontend/roster/components/roster_overview_component.html.erb"
```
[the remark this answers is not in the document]
fosterfarrell9: The roster overview component has been removed anyway, so this line is moot. Fixed.

app/helpers/application_helper.rb
```diff
@@ -48,6 +48,46 @@ def hide(value)
...
+    end
+
+    tag.div(class: container_class, style: [style, "height: #{height}"].compact.join("; ")) do
+      tag.div(class: "progress-bar #{color_class}",
+              role: "progressbar",
+              style: "width: #{percentage}%",
+              "aria-valuenow": value,
+              "aria-valuemin": 0,
+              "aria-valuemax": max) do
+        label || "#{percentage.round}%" if show_label
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

spec/requests/rosters/stream_builder_spec.rb
```diff
@@ -0,0 +1,122 @@
+require "rails_helper"
+
+RSpec.describe(Rosters::StreamBuilder, type: :request) do
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/models/registration/policy/prerequisite_campaign_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+    class PrerequisiteCampaignHandler < Handler
+      def evaluate(user)
+        if campaign_id.blank?
+          return fail_result(:configuration_error,
+                             "Prerequisite campaign not configured")
+        end
+        unless campaign
+          return fail_result(:prerequisite_campaign_not_found,
+                             "Prerequisite campaign missing")
+        end
```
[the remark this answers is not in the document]
fosterfarrell9: They don't surface in the UI yet, but will be via #1052, I guess. So better be on the safe side here. Fixed.

app/models/registration/policy/prerequisite_campaign_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+          @confirmed_user_ids.include?(user.id)
+        else
+          campaign.user_registration_confirmed?(user)
+        end
+
+        if confirmed
+          pass_result(:prerequisite_met)
+        else
+          fail_result(:prerequisite_not_met,
+                      "Prerequisite campaign not completed")
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

Splines:
@coderabbitai full review

app/controllers/cohorts_controller.rb
```diff
@@ -0,0 +1,179 @@
...
+    end
+
+    flash.now[:notice] = t("controllers.cohorts.created") if persisted
+    @errors = @cohort.errors
+
+    respond_to do |format|
+      format.turbo_stream do
+        group_type = parse_group_type
+        streams = create_turbo_streams(group_type)
+        render turbo_stream: streams, status: @cohort.persisted? ? :ok : :unprocessable_content
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/controllers/tutorials_controller.rb
```diff
@@ -55,28 +59,129 @@ def new
...
+    respond_to do |format|
+      format.js do
+        Rails.logger.warn("[MUESLI-DEPRECATION] Legacy JS format accessed in " \
+                          "TutorialsController#create")
+      end
+      format.turbo_stream do
+        group_type = parse_group_type
+
+        streams = create_turbo_streams(group_type)
+        render turbo_stream: streams, status: @tutorial.persisted? ? :ok : :unprocessable_content
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/controllers/registration/allocations_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+          )
+        end
+      end
+    end
+
+    def create
+      authorize! :allocate, @campaign
+      if @campaign.closed? || @campaign.processing?
+        Registration::AllocationService.new(@campaign).allocate!
+        @dashboard = Registration::AllocationDashboard.new(@campaign)
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/controllers/registration/campaigns_controller.rb
```diff
@@ -0,0 +1,294 @@
...
+      was_processing = @campaign.processing?
+
+      attributes = { status: :open }
+      if params[:registration_deadline].present?
+        attributes[:registration_deadline] = params[:registration_deadline]
+      end
+
+      if @campaign.update(attributes)
+        @campaign.reset_allocation_results! if was_processing
+        respond_with_success(t("registration.campaign.reopened"))
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/models/registration/campaign.rb
```diff
@@ -47,14 +63,238 @@ def policies_satisfied?(user, phase: :registration)
...
+    end
+
     private
 
+      def ensure_editable
+        return unless completed?
+        return unless changed?
+
+        errors.add(:base, :already_finalized) unless status_changed?
+      end
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/controllers/roster/maintenance_controller.rb
```diff
@@ -0,0 +1,294 @@
...
+    end
+
+    def bulk_update_self_materialization
+      mode = @mparams.mode
+      query = Rosters::NoCampaignRegisterablesQuery.new(@lecture)
+
+      ActiveRecord::Base.transaction do
+        query.scopes_by_type.each do |scope|
+          scope.update_all(self_materialization_mode: mode) # rubocop:disable Rails/SkipsModelValidations
+        end
```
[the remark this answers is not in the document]
fosterfarrell9: Disagree. For a bulk operation this is an acceptable trade-off. Note that `Rosters::NoCampaignRegisterablesQuery` already returns only tutorials/talks/cohorts that are not in any campaign — so the "campaigns are locked" validation can't fire anyway. Lectures are never in the query results — so the "Lectures can't enable it" validation is irrelevant.

spec/models/registration/campaign_spec.rb
```diff
@@ -97,6 +83,116 @@
...
+    it "validates registration_deadline is in the future if open" do
+      campaign = create(:registration_campaign, :open)
+      campaign.registration_deadline = 1.day.ago
+      expect(campaign).not_to be_valid
+      expect(campaign.errors.added?(:registration_deadline, :must_be_in_future)).to be(true)
+    end
+
+    it "validates prerequisites are not draft if open" do
+      prereq = create(:registration_campaign, :draft)
+      campaign = create(:registration_campaign, :draft)
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

spec/models/registration/campaign_spec.rb
```diff
@@ -153,4 +249,391 @@
...
+      campaign.reset_allocation_results!
+
+      ranks = campaign.user_registrations.reload.order(:preference_rank)
+                      .pluck(:preference_rank)
+      expect(ranks).to eq([1, 2])
+    end
+
+    it "zeroes out confirmed_registrations_count on all items" do
+      item1.update_columns(confirmed_registrations_count: 5)
+      item2.update_columns(confirmed_registrations_count: 3)
```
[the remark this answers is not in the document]
fosterfarrell9: No,it is fine here.

app/models/rosters/maintenance_service.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      unless force || within_capacity?(rosterable)
+        raise(CapacityExceededError,
+              "Capacity exceeded for #{rosterable.class.name} #{rosterable.id}")
+      end
+
+      rosterable.add_user_to_roster!(user)
+      propagate_to_lecture!(user, rosterable)
+      update_registration_materialization(user, rosterable)
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

app/views/cohorts/_modal_form.html.erb
```diff
@@ -0,0 +1,72 @@
+<%= form_with model: cohort,
+              id: dom_id(cohort, "form"),
+              data: {
+                turbo: true,
+                controller: "capacity-guard",
+                action: "submit->capacity-guard#submit",
+                capacity_guard_current_count_value: cohort.members.count,
+                capacity_guard_original_capacity_value: cohort.capacity || 0,
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed.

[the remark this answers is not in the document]
fosterfarrell9: This is already fixed.

app/controllers/registration/campaigns_controller.rb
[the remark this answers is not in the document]
Splines: could you double-check @fosterfarrell9 ?

app/frontend/entrypoints/application.scss
```diff
@@ -46,6 +46,7 @@ $container-max-widths: (
 @import "~/styles/submissions.scss";
 @import "~/styles/vertices.scss";
 @import "~/styles/link.scss";
+@import "../styles/roster.scss";
```
Splines: This CSS shouldn't be included in every single view we have (which it is as this is `application.scss`)

app/frontend/registration/campaigns/_add_group_tile.html.erb
```diff
@@ -0,0 +1,38 @@
...
+                    registration_section: registration_section,
+                    registration_campaign_id: campaign&.id)
+end %>
+
+<% flexible_group_path =
+  new_cohort_path(lecture_id: lecture.id,
+                  group_type: :cohorts,
+                  registration_section: registration_section,
+                  registration_campaign_id: campaign&.id,
+                  cohort: { propagate_to_lecture: true }) %>
```
Splines, line 22: It seems this is the only place we set `propagate_to_lecture` to `true`. Where do we initialize cohorts that don't propagate to lecture as they are planning only?

On this note: we explicitly don't have the `purpose` column in the `cohorts` model anymore, right? (just noting this since the Architecture book talks about this "purpose-driven behavior" [here](http://localhost:3004/features/03-rosters.html?highlight=purpose#cohort-propagation)).

db/schema.rb
```diff
@@ -935,13 +997,29 @@
     t.index ["tutorial_id"], name: "index_tutor_tutorial_joins_on_tutorial_id"
   end
```
Splines, line 1000: (maybe Muesli-off-topic, but still relevant and feel free to fix it here):
We don't have a unique constraint here on `[:tutorial_id, :tutor_id]`. I think we should have for data integrity, such that a tutor cannot get assigned to the same tutorial twice at DB level.

db/schema.rb
```diff
@@ -311,6 +354,18 @@
...
+    t.uuid "source_campaign_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["lecture_id"], name: "index_lecture_memberships_on_lecture_id"
+    t.index ["source_campaign_id"], name: "index_lecture_memberships_on_source_campaign_id"
+    t.index ["user_id", "lecture_id"], name: "index_lecture_memberships_on_user_id_and_lecture_id", unique: true
+    t.index ["user_id"], name: "index_lecture_memberships_on_user_id"
+  end
+
   create_table "lecture_user_joins", force: :cascade do |t|
```
Splines, line 369: For the `lecture_memberships` we have a unique constraint on `[:lecture_id, :user_id]`, but not on the lecture itself, i.e. a user could (on the DB-level) subscribe to the same lecture multiple times. I think we don't want that.

app/models/lecture.rb
```diff
@@ -86,6 +90,8 @@ class Lecture < ApplicationRecord
   # users to tutors, editors or teachers
   has_many :vouchers, dependent: :destroy
 
+  has_many :cohorts, as: :context, dependent: :destroy
```
Splines, line 93: I'm not entirely sure how deletion works for our polymorphic `context` of the Cohorts (see columns `context_type` and `context_id`). Unfortunately, it's [not possible](https://stackoverflow.com/q/922184/) to have a foreign key constraint in a polymorphic association.

What do we want to happen with a Cohort if the lecture is deleted? I think it should be deleted as well, right? Just like the code here is proposing via `dependent: :destroy`.

Could you write some tests for scenarios where main entities like lectures is deleted (skipping users for now).

app/controllers/talks_controller.rb
```diff
@@ -19,13 +21,38 @@ def new
     authorize! :new, @talk
     I18n.locale = @talk.lecture.locale_with_inheritance ||
                   current_user.locale || I18n.default_locale
+
+    respond_to do |format|
+      format.js do
+        Rails.logger.warn("[MUESLI-DEPRECATION] Legacy JS format accessed in " \
+                          "TalksController#new")
```
Splines, line 28: We have multiple of those legacy JS format warnings. Not sure when this particular path here (JS over TurboStream) could ever happen and why we want to catch it. I mean if the user has deactivated JS, they won't get neither JS nor Turbo_Stream, and if they have activated JS, why shouldn't they get the Turbo_Stream?

For future PRs: I think in general, we use so much UI-critical JS now with Hotwire that we should just introduce an explicit warning banner whenever somebody has deactivated JS. This way, we wouldn't even have to bother about `format.html` anymore and just render whatever JS we prefer, e.g. turbo streams directly.

config/locales/registration/de.yml
```diff
@@ -1,7 +1,337 @@
 de:
+  registration:
+    campaign:
+      errors:
+        already_finalized: "Anmeldeverfahren ist bereits abgeschlossen."
+      created: "Anmeldeverfahren erfolgreich erstellt."
+      create_failed: "Anmeldeverfahren konnte nicht erstellt werden."
+      destroyed: "Anmeldeverfahren erfolgreich gelöscht."
+      cannot_delete: "Anmeldeverfahren kann nicht gelöscht werden."
+      confirm_delete: "Sind Sie sicher, dass Sie dieses Anmeldeverfahren löschen wollen?"
```
Splines: we should use "Du" everywhere instead of "Sie"

app/frontend/registration/campaigns/_card_header.html.erb
```diff
@@ -0,0 +1,168 @@
...
+                            data: { turbo_frame: campaign_policy_form_frame_id(campaign) },
+                            title: t("basics.edit") do %>
+                  <i class="bi bi-pencil-fill"></i>
+                <% end %>
+
+                <%= button_to registration_campaign_policy_path(campaign, policy),
+                              method: :delete,
+                              class: "policy-badge-action text-danger",
+                              form: { class: "d-inline" },
+                              data: { turbo_confirm: t("confirmation.generic"), turbo_stream: true },
```
Splines: Don't use the "confirmation.generic" since a phrase like "Are you sure?" does not clearly state what will hapen if the user confirms the dialog.

Splines: same for other places in this PR

app/frontend/registration/campaign_confirm_base.controller.js
```diff
@@ -0,0 +1,43 @@
...
+  connect() {
+    this.confirmed = false;
+  }
+
+  async buildMessage() {
+    let message = this.confirmMessageValue;
+
+    try {
+      const response = await fetch(
+        `/campaigns/${this.campaignIdValue}/check_unlimited_items`,
```
Splines: Is there a specific reason why we need to exchange JSON here and therefore introduce a "custom API" beyond Rails-standard functionality?

Splines (changes requested):
Nice, this PR looks very mature, happy that we are close to merging!

Some first review (posting this early while I'm still reviewing to allow for a quicker feedback cycle with you)

app/frontend/registration/campaigns/_add_group_tile.html.erb
fosterfarrell9: This `{ propagate_to_lecture: true }` contains the default setting for the "Enable autoamtic lecture enrollment" switch in the "Create Cohort" modal (which should better be called "Create flexible group", I guess). Together with the switch, there is the text "If enabled, members are automatically enrolled in the lecture. Disable this for waitlists or planning groups.". The modal pops up when you click the "Add flexible group" button in the "add tile". Having the default set this way and having the switch to turn it off makes sense to me.

And yes, the `purpose` column for cohorts does no longer exist (maybe that is not yet reflected correctly in the architecture book).

app/controllers/talks_controller.rb
fosterfarrell9: As discussed on basecamp, this legacy js still comes into action when the feature flags are disabled. In this special case: in the "old" system, talks were created within the content tab, while in the new system (which uses hotwire) they are created in the groups tab.

db/schema.rb
fosterfarrell9: Good point, I added a uniqueness index in [b4df379](https://github.com/MaMpf-HD/mampf/pull/1081/commits/b4df379aae304ae747f8c6ce74c9f3a23b4ab5af).

db/schema.rb
fosterfarrell9: I added a uniqueness index in [b4df379](https://github.com/MaMpf-HD/mampf/pull/1081/commits/b4df379aae304ae747f8c6ce74c9f3a23b4ab5af).

app/frontend/entrypoints/application.scss
fosterfarrell9: Indeed. I fixed this in [7374c88](https://github.com/MaMpf-HD/mampf/pull/1081/commits/7374c88f7274dd5a81e3de54b6990c07b4421cbb)[7374c88](https://github.com/MaMpf-HD/mampf/pull/1081/commits/7374c88f7274dd5a81e3de54b6990c07b4421cbb). I played around in the GUI, still seems to work. But better check as well.

app/models/lecture.rb
fosterfarrell9: Yes, when a lecture is deleted, so should be all of its cohorts. While we cannot enforce this natively with a foreign key, Rails' `dependent: :destroy` also works for polymorphic associations, see [here](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#module-ActiveRecord::Associations::ClassMethods-label-Polymorphic+Associations) in the second example. We have many instances of the same pattern in our code base that have been working fine for a long time, e.g.:

```ruby
class Tutorial
  has_many :claims, as: :claimable, dependent: :destroy
````

(and there are many more of this).
If you insist that I write specs for this I will, but that should actually covered by the specs for Rails itself:
They test that dependent destroy works (apparently uniformly on several associations) in [here](https://github.com/rails/rails/blob/main/activerecord/test/cases/associations/has_many_associations_test.rb),
also on [this](https://github.com/rails/rails/blob/main/activerecord/test/models/car.rb) model which "has many wheels as a wheelable".

File: app/frontend/registration/campaign_confirm_base.controller.js (hunk 30 above)
Comment (fosterfarrell9) [reply app/frontend/registration/campaign_confirm_base.controller.js]: No, this is just bad code. I rewrote it in [6a38bc2](https://github.com/MaMpf-HD/mampf/pull/1081/commits/6a38bc2758a38305a594339e21c1a6d071eeb25e) to just make use of Hotwire (and also removed this capacity warning from the finalization text; it has no place there).

Comment (fosterfarrell9) [reply app/frontend/registration/campaign_confirm_base.controller.js]: I just saw that I messed up in [6a38bc2](https://github.com/MaMpf-HD/mampf/pull/1081/commits/6a38bc2758a38305a594339e21c1a6d071eeb25e), so [5cee649](https://github.com/MaMpf-HD/mampf/pull/1081/commits/5cee6490836f32c309ed9086e72d5a97d94a4f6e) fixes it and is also part of the fix for the comment.

File: app/controllers/registration/campaigns_controller.rb (hunk 15 above)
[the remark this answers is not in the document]
Comment (fosterfarrell9) [reply app/controllers/registration/campaigns_controller.rb]: @coderabbitai Can you recheck? The code for the `reopen` method already looks like this:

```ruby
    def reopen
      if @campaign.completed?
        respond_with_error(t("registration.campaign.cannot_reopen_completed"))
        return
      end

      was_processing = @campaign.processing?

      attributes = { status: :open }
      if params[:registration_deadline].present?
        attributes[:registration_deadline] = params[:registration_deadline]
      end

      @campaign.transaction do
        @campaign.update!(attributes)
        @campaign.reset_allocation_results! if was_processing
      end

      respond_with_success(t("registration.campaign.reopened"))
    rescue ActiveRecord::RecordInvalid
      respond_with_error(@campaign.errors.full_messages.join(", "))
    end
```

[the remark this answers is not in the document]
fosterfarrell9: @coderabbitai hop out of your burrow

config/locales/registration/en.yml
```diff
@@ -1,8 +1,338 @@
...
+      cannot_delete: "Campaign cannot be deleted."
+      confirm_delete: "Are you sure you want to delete this campaign?"
+      opened: "Campaign opened."
+      closed: "Campaign closed."
+      reopened: "Campaign reopened."
+      finalized: "Campaign finalized successfully."
+      cannot_reopen_completed: "Cannot reopen campaign because it has already been finalized."
+      not_found: "Registration campaign not found."
+      completed_title: "Campaign Completed"
+      completed_description: "The registration period has ended and allocations have been finalized. Please use the Roster to manage student groups."
```
Splines: no user knows what a "Roster" is, I think we should avoid using that word altogether in the UI

config/locales/roster/en.yml
```diff
@@ -0,0 +1,294 @@
...
+      campaign_processing_short: "Processing"
+      campaign_completed: "Campaign: Completed"
+      campaign_completed_short: "Completed"
+      post_campaign: "Post-campaign"
+      post_campaign_short: "Post-camp."
+      self_enrollment: "Self-enrollment"
+      no_policy_enforcement: "No policy enforcement"
+      gated_by_policies: "Gated by policies: %{policies}"
+      enforced_policies: "Campaign enforced: %{policies}"
+    participants_hint: "This list contains all people enrolled in this lecture. Use the filters:<ul><li><strong>All:</strong> Shows everyone enrolled in the lecture</li><li><strong>Without group:</strong> Shows people who are enrolled but not member of any tutorial or cohort with lecture enrollment</li></ul>"
```
Splines: "Cohort" is quite a technical word. I'd like to avoid it in any user-facing dialog.

app/frontend/roster/components/roster_participants_component.html.erb
```diff
@@ -0,0 +1,141 @@
...
+          <div class="input-group input-group-sm">
+            <span class="input-group-text bg-white border-end-0 pe-1">
+              <i class="bi bi-search text-muted"></i>
+            </span>
+
+            <%= f.text_field :search,
+                             value: @search_string,
+                             placeholder: t("basics.search"),
+                             class: "form-control border-start-0 ps-1 focus-ring-0 shadow-none",
+                             data: { action: "input->auto-submit-form#debouncedSubmit" } %>
```
Splines: Some remarks for the search input field:

- When you click inside it, the border is thicker on one side.
- When you write something and are slow, the form already autosubmits and one looses text cursor focus and has to reclick inside the search bar to continue typing. This is quite annoying.
- It would be nice if the input field had an "X" symbol at the right side to delete the entire search string.
- The search bar could also be a tiny bit longer.

config/locales/registration/de.yml
Splines: fixed in c77507c5528 (replacing "Sie" by "Du")

app/controllers/registration/campaigns_controller.rb
```diff
@@ -0,0 +1,288 @@
...
+      end
+
+      if @campaign.update(attributes)
+        respond_with_success(t("registration.campaign.closed"))
+      else
+        respond_with_error(@campaign.errors.full_messages.join(", "))
+      end
+    end
+
+    def reopen
```
Splines, line 143: A question regarding completed campaigns. For `reopen`, you explicitly block the completed campaign from reopening and respond with an error here.

However, it seems that this is not checked in the `open`, `update` and `update_status` methods, so one could still change the campagn status through open or close, even though the reopen method here explicitly forbids it.

The respective test named `allows updates if status is changing (re-opening)` even assures this behavior. I was just wondering if that is really what we want.

app/controllers/registration/campaigns_controller.rb
```diff
@@ -0,0 +1,288 @@
...
+              "new_campaign_form",
+              partial: "registration/campaigns/form",
+              locals: { campaign: @campaign,
+                        lecture: @campaign.campaignable }
+            ), status: :unprocessable_content
+          end
+        end
+      end
+
+      def respond_with_error(message, redirect_path: nil)
```
Splines: It seems methods like `respond_with_error`, `respond_with_success` and the like are reinvented in many different controllers, although they basically do the same thing.

Please also take a look at the `flash.rb` module I once made and which may help here (or could maybe be extended should it not suffice).

app/frontend/registration/policies/_form.html.erb
```diff
@@ -0,0 +1,99 @@
...
+    <!-- Institutional Email -->
+    <div
+      data-registration-policy-form-target="configSection"
+      data-kind="institutional_email"
+      class="d-none"
+    >
+      <div class="mb-3">
+        <%= f.label :allowed_domains, t("registration.policy.configs.allowed_domains"), class: "form-label" %>
+        <%= helpdesk(t("registration.policy.helpdesk.allowed_domains"), false) %>
+        <%= f.text_field :allowed_domains, class: "form-control" %>
```
Splines: Idea: add a new env variable for a default allowed domain such that people can use a default like `uni-heidelberg.de` here. This also reduces the risk of typos and thus annoyed students and teachers.

db/migrate/20260404000010_add_skip_campaigns_to_registerables.rb
```diff
@@ -0,0 +1,7 @@
+class AddSkipCampaignsToRegisterables < ActiveRecord::Migration[8.0]
+  def change
+    add_column :tutorials, :skip_campaigns, :boolean, default: false, null: false
+    add_column :talks, :skip_campaigns, :boolean, default: false, null: false
+    add_column :cohorts, :skip_campaigns, :boolean, default: false, null: false
```
Splines: Modify this migration to account for old tutorials. They should get a `skip_campaign: true`, even though the new default will be `false`.

Splines (changes requested):

app/controllers/registration/campaigns_controller.rb
```diff
@@ -0,0 +1,288 @@
...
+      unless params[:source] == "panel"
+        redirect_to edit_lecture_path(@campaign.campaignable, tab: "groups")
+        return
+      end
+
+      # Exclude any users already on the lecture roster
+      lecture_roster_ids = @campaign.campaignable.allocated_user_ids
+
+      @unassigned_users = @campaign.unassigned_users
+                                   .where.not(id: lecture_roster_ids)
```
Splines: In 2bd0ecf8ad, I've added a failing test that essentially would be fixed by removig this `where` condition here. Let me know what you think about the test and what the desired behavior should be according to your opinion.

Splines: As a summary, here is what the test is doing:

1. Student previously assigned and is on lecture roster.
2. Student is no longer in any tutorial/talk/cohort for the campaign.
3. Campaign correctly consdiers them as unassigned, see the `unassigned_users` in the Campaign model.
4. But the controller doesn't show them as unassigned, because they are still on the lecture roster. This is what is done here in the `unassigned` method by the `where` condition.

This mismatch is what the test points out and thus the panel for the completed campaign (see `dissolved_footnote`) wouldn't show the student as "unplaced registration" even though they are unassigned.

app/frontend/registration/allocations/_item_capacities_table.html.erb
```diff
@@ -0,0 +1,111 @@
...
+
+            <% if data[:item].registerable_type == "Tutorial" %>
+              <details
+                class="allocation-capacity-editor d-inline-block ms-1"
+                data-controller="details-editor"
+                data-action="toggle->details-editor#toggle"
+              >
+                <summary class="btn btn-sm btn-link text-secondary p-0 align-baseline">
+                  <i class="bi bi-pencil"></i>
+                </summary>
```
Splines: Why do you use the `summary` tag here for a pencil button? This is not accessible.

Furthermore, when you currently click on this pencil button here, the sidepanel opens and closes all the time as the whole row is triggered. This shifts the layout. Maybe don't trigger the whole toggle if only the pencil button is clicked.

spec/requests/roster/maintenance_spec.rb
```diff
@@ -0,0 +1,554 @@
...
+
+      it "does not move the user" do
+        expect do
+          patch(move_member_lecture_path(lecture, user_id: member.id),
+                params: { target_id: target_tutorial.id, target_type: "Tutorial" })
+        end.not_to(change { tutorial.members.count })
+      end
+    end
+  end
+
```
Splines, line 445: I've added this failing test, let me know what you think. The problem is that the `bulk_update_self_materialization` method uses `update_all` which skips validations and callbacks. Thus, the rosterable validation is not run and `skip_campaigns` is not correctly updated, leaving the group in an inconsistent state.

Here in the test, the `free_tutorial` should be bulk-updated and should end up with both `add_only` self materialization mode and `skip_campaigns = true`, but it ends up with `skip_campaigns = false`.

The problem can be fixed by replacing part of the `bulk_update_self_materialziation` method with this:

```rb
ActiveRecord::Base.transaction do
        query.scopes_by_type.each do |scope|
          scope.find_each do |registerable|
            registerable.update!(self_materialization_mode: mode)
          end
        end
      end
```

Splines (changes requested):

config/locales/registration/en.yml
fosterfarrell9: Fixed in [3da7790](https://github.com/MaMpf-HD/mampf/pull/1081/commits/3da77901d790703ac634e4855c7668905c50582a).

config/locales/roster/en.yml
fosterfarrell9: I replaced it by "flexible group" in [a5c0640](https://github.com/MaMpf-HD/mampf/pull/1081/commits/a5c06402b47ea1a79a64659d9b6af63b356dc5d3). Note that there are other appearances of the word cohort in the locale file texts (not just as keys, where it is okay), but most of them seem to be unused. How do we want to proceed?

app/controllers/registration/campaigns_controller.rb
fosterfarrell9: I agree that this needs to be DRYed up. I did this in [66f0a76](https://github.com/MaMpf-HD/mampf/pull/1081/commits/66f0a764b3bd692e632dc10433100b8051997836), [9a4be8c](https://github.com/MaMpf-HD/mampf/pull/1081/commits/9a4be8c615d36c2ea71d46631ce6927aededaf88) Please check that everything still works as expected. I added dedicated specs for the new `respond_with_flash` method that is now used throughout several controllers and cut down some request specs that were testing things now covered by the tests for `respond_with_flash`.

app/controllers/registration/campaigns_controller.rb
fosterfarrell9: The `Campaign` class actually has a built-in protection against updating completed campaigns:

```ruby
validate :ensure_editable, on: :update

....

      def ensure_editable
        return unless status_was == "completed"
        return unless changed?

        errors.add(:base, :already_finalized) unless status_changed?
      end
```

Unfortunately, this is too weak, as the `unless` would indeed allow changing the status away from `completed`, which is not what we want. I would suggest that we remove the unless clause here. This would also disallow doing such updates via the methods you mentioned.

In the `reopen` action, the existing check for completedness gives a slightly more precise error message than the generic one  produced by the validation.

What do you think on this?

app/controllers/registration/campaigns_controller.rb
fosterfarrell9: I think I agree with you here. The question is what the source of truth is for the unassigned panel:

- Lecture roster = "is this person placed in the lecture?"
- Campaign item assignment = "did this specific campaign produce a valid placement?"

I guess the second one should be the source of truth (which means we should indeed remove the `where` condition here). The consequence is then of course that when another student sucessfully assigned this student, the student will still be shown as unassigned here (this was possibly the motivation for the additional clause here). I remember that I spent some time thinking about this in the past already.

spec/requests/roster/maintenance_spec.rb
fosterfarrell9: I agree, your fix is the correct approach for this.

Splines: Ok, fixed in 41bcdc59c4cac

app/controllers/registration/campaigns_controller.rb
Splines: I went with this option now:

> Campaign item assignment = "did this specific campaign produce a valid placement?"

or in other words: even if a student is already on the lecture roster (e.g. via an old campaign), but looses their group assignment, they should show up in the "unassigned panel" for a newly opened campaign.

Please carefully review my respective commit 04ce2022d5ab8 (see also the commit message, where I describe what changed).

---

> The consequence is then of course that when another student sucessfully assigned this student, the student will still be shown as unassigned here

Sorry, I don't get this sentence. How can a student assign another student? Did you mean teacher/editor?

fosterfarrell9: It should read "another campaign successfully assigned this student".

app/controllers/registration/campaigns_controller.rb
Splines: The new `respond_with_flash` method is such an elegant solution with the `block_given?` approach! I went carefully over every change and it looks fine.

In 578a163478, I've introduced a new `assert_flash_error` test assertion.

Furthermore, you changed `flash.keep[...]` to just `flash[...]` here. For the case where no message is passed, this is correctly handled by the `if message` conditional, so in this regards the behavior stays the same. However, existing previous flash messages are no longer kept. I don't know if this is really a problem and to be honest, this is just the fallback to html, so let's not worry too much about this. But for the case we encounter some issues, let's keep this in the back of our minds.

Splines: feel free to resolve

app/controllers/registration/campaigns_controller.rb
Splines: feel free to resolve

app/controllers/registration/campaigns_controller.rb
Splines: I think the way you suggested is fine by removing the last `unless status_changed?` statement.

Fixed in d233a1523e, feel free to resolve.

app/frontend/registration/policies/_form.html.erb
Splines: I've implemented this in 6d3b91c9. To quickly test this out without having to rebuild the entire DevContainer setup, simply put `MUESLI_CAMPAIGN_REGISTRATION_DEFAULT_ALLOWED_DOMAIN=uni-heidelberg.de` in front of `bundle exec` in `run.sh` and restart the MaMpf server. Then head to the policies. You should see this:

image

feel free to resolve

config/locales/roster/en.yml
Splines: I went over the other occurrences of "Cohort" and fixed them in 87494ce2d2f. We don't have a fixed wording, e.g. "Sondergruppe", "flexible group", "group" can all be found in the locales meaning basically the same thing. But I can live with that right now ^^

I wouldn't bother about unused i18n keys at the moment, let's just do one PR that removes all of them automatically afterwards instead of wasting time by doing this manually.

Splines: feel free to resolve

db/migrate/20260404000010_add_skip_campaigns_to_registerables.rb
Splines: I've done this in 6e4e95e37e4, feel free to resolve.

app/frontend/roster/components/roster_participants_component.html.erb
fosterfarrell9: I have worked on these points in [41e5505](https://github.com/MaMpf-HD/mampf/pull/1081/commits/41e5505267d50f9dff1823d412fff3b51f769568). Check if this is okay with you.

app/frontend/registration/allocations/_item_capacities_table.html.erb
fosterfarrell9: I addressed this in [30b3837](https://github.com/MaMpf-HD/mampf/pull/1081/commits/30b3837a551b858a01c8b376662d0dad97b3912c). I also changed the info text to include a mention of the possibility to edit the capacities directly via these buttons.

db/schema.rb
fosterfarrell9: And I changed the duplicate removal from raw SQL to ActiveRecord. Still, we might just check whether there are any duplicates at all in production, and if not, throw the removal lines out.

app/frontend/registration/campaigns/_card_header.html.erb
fosterfarrell9: Fixed in [4c66396](https://github.com/MaMpf-HD/mampf/pull/1081/commits/4c6639670935bd0d3b0168f8458d75d055cadce6).

app/frontend/roster/components/roster_participants_component.html.erb
Splines: Thanks! So the "X" clear button is working fine, the other features not yet:

- The border has two different colors and is composed of two separate elements that have a different animation length. Can't we use just one input box with one coherent border?
- The autosubmit feature is unreliable. Sometimes, I randomly lose focus of the input box, cannot delete characters (because a request is still in process and then comes back when I currently type), or the search doesn't work at all.
- Furthermore, the layout shifts are quite annoying. The horizontal layout shift due to the scroll bar appearing/disappearing is fine, but the vertical one is more difficult to look at.

https://github.com/user-attachments/assets/2cad32d9-c5e7-41fc-a599-fd24a784e34f

I will tackle those problems soon.

app/frontend/registration/allocations/_item_capacities_table.html.erb
Splines: this works like a charm

app/frontend/roster/components/roster_participants_component.html.erb
Splines: Fixed in 5040611b0 and 9a70b0481b3. Feel free to resolve.

app/frontend/registration/campaigns/_add_group_tile.html.erb
Splines: Thanks, this makes sense.

db/schema.rb
Splines: thanks! this was later refined and then brought to next via #1089

db/schema.rb
Splines: thanks! this was later refined and then brought to next via #1089

app/models/lecture.rb
Splines: Nice, thanks a lot for all those references, this helped!

app/models/registration/policy.rb
```diff
@@ -27,68 +34,76 @@ class Policy < ApplicationRecord
...
+
       case kind.to_sym
       when :institutional_email
-        evaluate_institutional_email(user)
+        Registration::Policy::InstitutionalEmailHandler.new(self)
       when :prerequisite_campaign
-        evaluate_prerequisite_campaign(user)
+        Registration::Policy::PrerequisiteCampaignHandler.new(self)
+      when :student_performance
+        Registration::Policy::Handler.new(self)
```
Splines, line 84: Note that this will raise, as the respective evaluate method of the policy will call `NotImplementedError`. This should be addressed in a follow-up PR.

Splines:
@coderabbitai full review

app/controllers/registration/allocations_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+    end
+
+    def create
+      authorize! :allocate, @campaign
+
+      unless @campaign.preference_based?
+        respond_with_flash(
+          :alert,
+          "Allocation can only be triggered for preference-based campaigns. " \
+          "As a user, you should never see this error, please contact the MaMpf team.",
```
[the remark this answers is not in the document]
Splines: This message should be fine. I didn't translate it since it should never occur in the UI.

app/frontend/registration/allocations/_policy_violations_alert.html.erb
```diff
@@ -0,0 +1,87 @@
...
+        <% end %>
+        <%= button_to destroy_for_user_registration_campaign_registrations_path(campaign, user_id: user_id, source: allocation_source),
+                      method: :delete,
+                      class: "btn btn-sm btn-link text-danger p-0 ms-2 align-baseline",
+                      data: {
+                        confirm:
+                        t("registration.campaign.confirmations.user_registration_conflict_remove",
+                        email: first[:email])
+                      },
+                      form: { class: "d-inline-block" },
```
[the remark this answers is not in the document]
Splines: fixed in d0a4b3f06

app/frontend/registration/policies/_form.html.erb
```diff
@@ -0,0 +1,104 @@
...
+    >
+      <div class="mb-3">
+        <label class="form-label">
+          <%= t("registration.policy.configs.lecture_id") %>
+        </label>
+
+        <%= select_tag "registration_policy[config][lecture_id]",
+                         options_from_collection_for_select(Lecture.all, :id, :title, policy.config&.fetch("lecture_id", nil)),
+                         include_blank: true,
+                         class: "form-select" %>
```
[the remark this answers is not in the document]
Splines: thanks, but will address this in a future PR, should be fine for now

app/models/tutorial.rb
```diff
@@ -46,10 +59,40 @@ def add_tutor(tutor)
...
+                        .where(user_id: user_ids)
+                        .delete_all
+    end
+
     def check_destructibility
-      throw(:abort) unless destructible?
-      true
+      return unless Submission.where(tutorial: self).proper.any?
+
+      errors.add(:base, I18n.t("controllers.tutorials.errors.cannot_delete_with_submissions"))
```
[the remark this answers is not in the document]
Splines: fixed in d0a4b3f06

db/migrate/20260404000002_create_flipper_tables.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+    add_index :flipper_features, :key, unique: true
+
+    create_table :flipper_gates do |t|
+      t.string :feature_key, null: false
+      t.string :key, null: false
+      t.text :value
+      t.timestamps null: false
+    end
+    add_index :flipper_gates, [:feature_key, :key, :value], unique: true, length: { value: 255 }
```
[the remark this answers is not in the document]
Splines: fixed in 407add196360

db/migrate/20260404000012_add_location_to_tutorials.rb
```diff
@@ -0,0 +1,5 @@
+class AddLocationToTutorials < ActiveRecord::Migration[8.0]
```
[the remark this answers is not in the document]
Splines: This is not true, or to be more precise, it was wrongly specified in our instructions file. Fixed in 59ddcece7f0

lib/tasks/demo_setup.rake
```diff
@@ -0,0 +1,33 @@
...
+    abort "Cannot run in production!" if Rails.env.production?
+
+    puts "=== Demo Setup ==="
+    puts ""
+
+    puts "Enabling feature flags..."
+    Flipper.enable(:roster_maintenance)
+    Flipper.enable(:registration_campaigns)
+    puts "  ✓ roster_maintenance"
+    puts "  ✓ registration_campaigns"
```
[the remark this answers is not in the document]
Splines: fixed in the PR description

spec/factories/registration_policies.rb
```diff
@@ -14,6 +15,13 @@
 
     trait :prerequisite_campaign do
       kind { :prerequisite_campaign }
+      after(:build) do |policy|
+        unless policy.config && policy.config["prerequisite_campaign_id"]
+          prereq = create(:registration_campaign, :completed)
+          policy.config ||= {}
+          policy.config["prerequisite_campaign_id"] = prereq.id
```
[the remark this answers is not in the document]
Splines: fixed in b30acc055

.herb.yml
```diff
@@ -23,7 +23,6 @@ linter:
     - "app/frontend/login/**/*.html.erb"
     - "app/frontend/media/**/*.html.erb"
     - "app/frontend/modal/**/*.html.erb"
-    - "app/frontend/roster/**/*.html.erb"
     - "app/frontend/search/**/*.html.erb"
     - "app/frontend/spinner/**/*.html.erb"
     - "app/frontend/vignettes/**/*.html.erb"
```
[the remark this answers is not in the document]
Splines: This comment does not make sense as there is nothing related to roster templates in this exclude list anymore.

app/controllers/cypress/feature_flags_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  class FeatureFlagsController < CypressController
+    def enable
+      Flipper.enable(params[:name].to_sym)
+      render json: { name: params[:name], enabled: true }, status: :created
+    end
+
+    def disable
+      Flipper.disable(params[:name].to_sym)
+      render json: { name: params[:name], enabled: false }, status: :created
+    end
```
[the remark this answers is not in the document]
Splines: This is fine, it's already guarded in the routes.rb

app/controllers/lectures_controller.rb
```diff
@@ -73,12 +73,13 @@ def new
...
-      eager_load_stuff
-      render template: "lectures/edit/edit"
+    eager_load_stuff
+    if params[:campaign_id]
+      @campaign = @lecture.registration_campaigns.find_by(id: params[:campaign_id])
+    elsif params[:new_campaign]
+      @new_campaign = @lecture.registration_campaigns.build
     end
+    render template: "lectures/edit/edit"
   end
```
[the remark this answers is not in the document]
Splines: Yes, this was a correctness concern. We might tackle this in a future PR should we experience too high load times (we should also track those)

app/helpers/application_helper.rb
```diff
@@ -48,6 +48,47 @@ def hide(value)
     value ? "none;" : "block;"
   end
 
+  # rubocop:disable Metrics/ParameterLists
+  def progress_bar(value, max, classification: :neutral, label: nil,
+                   height: "1.5rem", show_label: true,
+                   container_class: "progress mb-2", style: nil)
+    # rubocop:enable Metrics/ParameterLists
+    clamped_value = value.to_i.clamp(0, max.to_i)
+    percentage = max.to_i.positive? ? (value.to_f / max * 100).clamp(0, 100) : 0
```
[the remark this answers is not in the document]
Splines: fixed in be7b08f

app/models/rosters/maintenance_service.rb
```diff
@@ -0,0 +1,109 @@
...
+        unless force || within_capacity?(rosterable)
+          raise(CapacityExceededError,
+                "Capacity exceeded for #{rosterable.class.name} #{rosterable.id}")
+        end
+
+        rosterable.add_user_to_roster!(user)
+        propagate_to_lecture!(user, rosterable)
+        update_registration_materialization(user, rosterable)
+      end
+    end
```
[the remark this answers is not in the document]
Splines: fixed in acefb9c3c524

app/models/rosters/maintenance_service.rb
```diff
@@ -0,0 +1,109 @@
...
+        cascade_removal_from_subgroups!(user, rosterable)
+      end
+    end
+
+    def move_user!(user, from_rosterable, to_rosterable, force: false)
+      ActiveRecord::Base.transaction do
+        remove_user!(user, from_rosterable)
+        add_user!(user, to_rosterable, force: force)
+      end
+    end
```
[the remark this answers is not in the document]
Splines: also fixed in acefb9c3c524

app/views/talks/_roster_modal_form.html.erb
```diff
@@ -0,0 +1,98 @@
+<%= form_with model: talk,
+              id: dom_id(talk, "form"),
+              data: {
+                turbo: true,
+                controller: "capacity-guard",
+                action: "submit->capacity-guard#submit",
+                capacity_guard_current_count_value: talk.speakers.count,                capacity_guard_original_capacity_value: talk.capacity || "",                capacity_guard_confirm_message_value: t("roster.warnings.capacity_reduction")
```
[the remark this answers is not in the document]
Splines: fixed in 929b05df51

db/migrate/20260404000005_add_unique_index_to_registration_items.rb
```diff
@@ -0,0 +1,15 @@
...
+                 [:registerable_type, :registerable_id],
+                 name: "index_registration_items_on_registerable"
+    remove_index :registration_items,
+                 [:registration_campaign_id, :registerable_type, :registerable_id],
+                 unique: true,
+                 name: "index_registration_items_uniqueness"
+
+    add_index :registration_items, [:registerable_type, :registerable_id],
+              unique: true,
+              name: "index_registration_items_on_unique_registerable"
```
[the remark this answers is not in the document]
Splines: This is a valid point, but right now in the UI, we don't even allow dragging around existing tutorials to new campaigns. So won't do this in this PR.

app/models/registration/policy.rb
Splines: This is intended behavior as only future PRs will tackle this. Right now, there is no way to access this via the UI, so it should be fine.

Splines: See commit de9292c

fosterfarrell9: It was referenced in the policy dropdown. I removed this possibility in [de9292c](https://github.com/MaMpf-HD/mampf/pull/1081/commits/de9292cb79f51e4c4abea95783ed9d921807a9af).

.herb.yml
```diff
@@ -23,7 +23,6 @@ linter:
     - "app/frontend/login/**/*.html.erb"
     - "app/frontend/media/**/*.html.erb"
     - "app/frontend/modal/**/*.html.erb"
-    - "app/frontend/roster/**/*.html.erb"
     - "app/frontend/search/**/*.html.erb"
```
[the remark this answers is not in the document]
Splines: already handled

app/controllers/cypress/feature_flags_controller.rb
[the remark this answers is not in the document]
Splines: already handled

app/controllers/flash.rb
```diff
@@ -31,15 +31,26 @@ def render_flash
...
+        flash.now[flash_type] = message if message
+
+        streams = [stream_flash]
+        if block_given?
+          custom_streams = yield
+          streams += Array(custom_streams)
+        end
+
+        render turbo_stream: streams.compact
       end
```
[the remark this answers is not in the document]
Splines: this is fine

app/controllers/registration/allocations_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+          redirect_path: registration_campaign_path(@campaign)
+        )
+        return
+      end
```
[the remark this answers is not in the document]
Splines: fine, already discussed previously

app/frontend/lectures/edit/_participants.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="col-12">
+  <%= turbo_frame_tag "roster_participants_panel",
+                      src: lecture_roster_participants_path(@lecture),
```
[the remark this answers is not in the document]
Splines: fixed in f5b2ad209

app/frontend/lectures/edit/_roster.html.erb
```diff
@@ -0,0 +1,12 @@
+<% group_type ||= :all %>
+<div class="col-12">
+  <%= turbo_frame_tag roster_maintenance_frame_id(group_type),
+                      src: lecture_roster_path(@lecture, group_type: group_type),
```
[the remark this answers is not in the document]
Splines: fixed in f5b2ad209

lib/tasks/demo_setup.rake
```diff
@@ -0,0 +1,33 @@
+namespace :demo do
+  desc "Enable feature flags and create all demo campaigns with registrations"
+  task setup: :environment do
+    abort "Cannot run in production!" if Rails.env.production?
```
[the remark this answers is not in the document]
Splines: no, staging environment also use the production env, so this is fine

spec/models/rosters/registerable_ordering_spec.rb
```diff
@@ -0,0 +1,116 @@
+require "rails_helper"
+
+# Missing top-level docstring, please formulate one yourself 😁
```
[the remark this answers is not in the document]
Splines: haha fixed in 487695ef19

app/models/registration/policy.rb
```diff
@@ -27,68 +34,76 @@ class Policy < ApplicationRecord
...
       else
         raise(ArgumentError, "Unknown policy kind: #{kind}")
       end
     end
```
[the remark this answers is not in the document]
Splines: already discussed previously, this is fine for now

app/frontend/roster/components/roster_participants_component.html.erb
fosterfarrell9: Works like a charm!

app/models/tutorial_membership.rb
```diff
@@ -0,0 +1,24 @@
...
+# to distinguish between automated allocations and manual enrollments.
+class TutorialMembership < ApplicationRecord
+  belongs_to :user
+  belongs_to :tutorial
+  belongs_to :source_campaign, class_name: "Registration::Campaign", optional: true
+
+  validate :unique_membership_per_lecture
+
+  private
+
```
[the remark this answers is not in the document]
Splines: good point, this is something we might tackle in a follow-up PR

app/models/tutorial_membership.rb
```diff
@@ -0,0 +1,24 @@
...
+      return unless tutorial
+
+      scope = TutorialMembership.joins(:tutorial)
+                                .where(tutorials: { lecture_id: tutorial.lecture_id })
+                                .where(user_id: user_id)
+      scope = scope.where.not(id: id) if persisted?
+
+      return unless scope.exists?
+
+      errors.add(:base, :already_in_lecture_tutorial)
```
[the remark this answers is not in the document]
Splines: duplicate

Splines (approved):
Wohoo, amazing work ✨

PR #1082, opened 2026-04-07
Muesli: Exams auto campaigns

fosterfarrell9:
In this PR, we rewrite the user interface for exams in a way that users no longer have to deal with the underlying campaign; instead it is directly integrated into the exam setup.


fosterfarrell9:
Closed in favor of #1109

PR #1086, opened 2026-04-07
PR-8.6 — Submission GUI simplification (roster-based tutorials)

plienta:


app/views/submissions/_form.html.erb
plienta, line 30: for simplicity

PR #1096, opened 2026-04-14
Enforce uniqueness for tutorial memberships on a database level

fosterfarrell9:
A user should only be in one tutorial per lecture. This was previously enforced by an application-level validation only, which means it could be bypassed by any write path that skips validations — most notably the bulk `insert_all` used during campaign materialization.

This PR moves the constraint to the database by adding a `lecture_id` column to `tutorial_memberships` with a unique index on `(user_id, lecture_id)`. The migration backfills the column and deduplicates any historical violations before creating the index.

Other changes that follow from this:
- `materialize_allocation!` uses `upsert_all` against the new index, so a user already in a sibling tutorial gets moved atomically rather than via delete-then-insert.
- `Tutorial#lecture_id` is now validated as immutable after creation, making explicit an invariant that already held in practice.
- Concurrent manual adds that lose the DB race now raise `UserAlreadyInBundleError` instead of an unhandled `RecordNotUnique`.

Note for the reviewer: I cannot reproduce the failing playwright test locally. Instead different playwright test fails locally for me: the `expect(row2.timeOnSlide).toBe(String(6 + 21));` test in the vignettes spec file (getting `21` there).


db/migrate/20260414000000_add_lecture_id_to_tutorial_memberships.rb
```diff
@@ -0,0 +1,35 @@
+class AddLectureIdToTutorialMemberships < ActiveRecord::Migration[7.2]
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [5f31903](https://github.com/MaMpf-HD/mampf/pull/1096/commits/5f31903da4e21fbc66ffe8459c336497504a0f84).

app/models/tutorial.rb
```diff
@@ -63,24 +64,52 @@ def roster_entries
...
+  # into a UserAlreadyInBundleError, maintaining the same error contract
+  # that callers of the concern expect.
+  def add_user_to_roster!(user, source_campaign = nil)
+    super
+  rescue ActiveRecord::RecordNotUnique
+    conflicting = TutorialMembership
+                  .where(lecture_id: lecture_id, user_id: user.id)
+                  .where.not(tutorial_id: id)
+                  .first
+    raise(Rosters::UserAlreadyInBundleError, conflicting&.tutorial)
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [1ab3ffe](https://github.com/MaMpf-HD/mampf/pull/1096/commits/1ab3ffe6172ca5aa0ad99a35b8f41448f46d357b).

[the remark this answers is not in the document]
fosterfarrell9: Fixed in [1ab3ffe](https://github.com/MaMpf-HD/mampf/pull/1096/commits/1ab3ffe6172ca5aa0ad99a35b8f41448f46d357b).

app/models/tutorial.rb
```diff
@@ -63,24 +64,52 @@ def roster_entries
...
+          source_campaign_id: campaign.id,
+          created_at: now,
+          updated_at: now
+        }.merge(scope_attrs)
+      end
+
+      roster_entries.upsert_all( # rubocop:disable Rails/SkipsModelValidations
+        attributes,
+        unique_by: :index_tutorial_memberships_on_user_id_and_lecture_id,
+        update_only: [:tutorial_id, :source_campaign_id]
```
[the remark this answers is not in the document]
fosterfarrell9: This is incorrect. I added a spec in [212c849](https://github.com/MaMpf-HD/mampf/pull/1096/commits/212c8499f03428ee8d8ea08bee2f21919da0fc39)that proves that an update of the `updated_at` timestamp indeed happens. Adding it here in fact even triggers an error.

db/migrate/20260613000000_add_lecture_id_to_tutorial_memberships.rb
```diff
@@ -0,0 +1,35 @@
+class AddLectureIdToTutorialMemberships < ActiveRecord::Migration[8.0]
+  def up
```
Splines, line 2: The up method does this de-duplication logic. Do we really need this? It's not yet in production, right? I thought we don't have any TutorialMembership entries yet.

db/migrate/20260613000000_add_lecture_id_to_tutorial_memberships.rb
```diff
@@ -0,0 +1,35 @@
...
+    add_column :tutorial_memberships, :lecture_id, :bigint
+
+    # rubocop:disable Rails/SkipsModelValidations
+    TutorialMembership.update_all(
+      "lecture_id = " \
+      "(SELECT lecture_id FROM tutorials WHERE tutorials.id = tutorial_memberships.tutorial_id)"
+    )
```
Splines: I think speed is not a concern here (since we don't have any TutorialMemberships in production yet, right?). So I'd prefer a statement with ActiveRecord instead of manual sql.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2026_04_16_000000) do
+ActiveRecord::Schema[8.0].define(version: 2026_05_23_000000) do
```
Splines: sorry for the late review, but would be great to update the date once more to today ^^

app/models/tutorial_membership.rb
```diff
@@ -3,22 +3,33 @@
...
 
   private
 
+    def set_lecture_from_tutorial
+      self.lecture_id = tutorial&.lecture_id if lecture_id.nil?
+    end
+
     def unique_membership_per_lecture
-      return unless tutorial
+      return unless lecture_id && user_id
```
Splines, line 21: Is there a case when `lecture_id` or `user_id` can be `nil`?

app/models/rosters/maintenance_service.rb
```diff
@@ -96,8 +96,10 @@ def update_registration_materialization(user, rosterable)
       def ensure_uniqueness!(user, rosterable)
         return unless rosterable.is_a?(Tutorial)
 
-        siblings = rosterable.lecture.tutorials.where.not(id: rosterable.id)
-        membership = TutorialMembership.where(tutorial: siblings, user: user).first
+        membership = TutorialMembership
+                     .where(lecture_id: rosterable.lecture_id, user_id: user.id)
```
Splines: we could still use `user: user` instead of `user_id: user.id`, right? maybe same for `tutorial_id` and `lecture_id`?

app/models/tutorial.rb
```diff
@@ -63,24 +64,49 @@ def roster_entries
...
+      )
     end
```
Splines: This method seems brittle in my eyes. We duplicate the logic from the parent `add_missing_users!` and already some things have changed, e.g. `Time.current` is now stored as variable in `now`. Over time, this method may drift even more.

Instead, can you again validate if there is no better approach to this that only overwrites and handles what the difference, i.e. `upsert_all` instead of `insert_all`?

Splines (changes requested):
Nice! Here some first review comments.

Splines (changes requested):
Furthermore, upon executing the pending migration here, I see this diff in my worktree:

```diff
diff --git a/db/schema.rb b/db/schema.rb
index 626c3864e..cbacbafa4 100644
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -631,8 +631,10 @@ ActiveRecord::Schema[8.0].define(version: 2026_05_23_000000) do
     t.uuid "registration_campaign_id", null: false
     t.uuid "registration_item_id", null: false
     t.datetime "materialized_at"
+    t.boolean "exclusive_assignment", default: false, null: false
     t.index ["registration_campaign_id", "user_id", "preference_rank"], name: "index_reg_user_regs_unique_ranked", unique: true, where: "(preference_r
ank IS NOT NULL)"
-    t.index ["registration_campaign_id", "user_id"], name: "index_reg_user_regs_unique_unranked", unique: true, where: "(preference_rank IS NULL)"

+    t.index ["registration_campaign_id", "user_id", "registration_item_id"], name: "index_reg_user_regs_unique_item_user", unique: true
+    t.index ["registration_campaign_id", "user_id"], name: "index_reg_user_regs_unique_exclusive_assignment_unranked", unique: true, where: 
"((exclusive_assignment = true) AND (preference_rank IS NULL))"
     t.index ["registration_campaign_id"], name: "index_reg_user_regs_on_campaign_id"
     t.index ["registration_item_id"], name: "index_registration_user_registrations_on_registration_item_id"
     t.index ["status"], name: "index_registration_user_registrations_on_status"
```

Edit: This is not an issue anymore as you've merged `next` into this branch.

db/schema.rb
fosterfarrell9: Fixed in [491eb41](https://github.com/MaMpf-HD/mampf/pull/1096/commits/491eb412e16d4e4a00f60042a2246699aedf09de).

app/models/rosters/maintenance_service.rb
fosterfarrell9: Indeed. Fixed in [3d8c792](https://github.com/MaMpf-HD/mampf/pull/1096/commits/3d8c7929463bf83281e504a483199665e5f05783).

app/models/tutorial_membership.rb
fosterfarrell9: This method is connected to this validation

```ruby
  validate :unique_membership_per_lecture
```

The validation might (at some point) run against incomplete in-memory records that have not yet gone through the validation that comes with `belongs_to` (which is only enforced upon saving). So, it cannot hurt to have this i, I think. Of course it might be that it is never called with in-memory objects.

app/models/tutorial.rb
fosterfarrell9: I refactored this in [b848803](https://github.com/MaMpf-HD/mampf/pull/1096/commits/b8488039e6ba62e02e1089849bfeedc8bda81fcc). During the implementation I stmbled upon the `UserAlreadyInBundle` error and didnt like its positioning inside maintenance service module, so I extracted it.

db/migrate/20260613000000_add_lecture_id_to_tutorial_memberships.rb
fosterfarrell9: Indeed (I guess this was not clear to me at the time of writing this code). Removed in [8415eef](https://github.com/MaMpf-HD/mampf/pull/1096/commits/8415eef06ff01e3cee69bb2f925e1addbc980199).

db/migrate/20260613000000_add_lecture_id_to_tutorial_memberships.rb
fosterfarrell9: Removed in [8415eef](https://github.com/MaMpf-HD/mampf/pull/1096/commits/8415eef06ff01e3cee69bb2f925e1addbc980199) sicne we don't have data in production for this (and development we can just reseed).

app/models/tutorial_membership.rb
Splines: ok, so let's keep the guard (early-return), but only for the sake of avoiding unnecessary duplicate-check queries.

For your information, note from the [Validation Guide](https://guides.rubyonrails.org/active_record_validations.html):

> **1.5 Checking Validity**
> [...]
> An object instantiated with `new` will not report errors even if it's technically invalid, because validations are automatically run only when the object is saved, such as with the `create` or `save` methods.

> **7.2 Custom methods**
> [...]
> By default, such validations will run every time you call `valid?` or save the object.

app/models/tutorial.rb
Splines: Nice, this feels way better as abstraction.

Splines (approved):
Nice!

PR #1084, opened 2026-04-07
🪣 Müsli: Assessments, grading, exams...

fosterfarrell9:
### TODO

- [x] Split this PR into smaller ones with at most 6,000 lines of code each.

PR #1104, opened 2026-04-30
📦 Müsli bucket, slice 0: Rework finalization process

fosterfarrell9:
This PR closes a larger architectural gap that was left in #1052 and whose extent became only clear to me when I implemented student performance policies in #1084. The problem is that the old workflow mixed separate things like eligibility checking and allocation/finalization. In practice this meant that students who clearly failed a policy and shouldn't even be considered for free seats were actually allowed to get one, with the "violator dashboard" being the one step that prevented them from getting a seat - and where the teacher had to choose between just giving them a seat or completely hard-deleting their registration (with no trace left for either side what happened). The PR introduces a dedicated screening step before allocation/finalization that classifies outcomes as `pass`, `auto_reject`, or `blocker`. The last one is necessary for situations where data are missing for policy evaluation like missing certificates for student performances. In the new solution, only blockers are being brought to the teachers attention in the finalization process, registrations with `auto_reject` are automatically rejected, but they are stored, together with their rejection reason. In post-finalization, teachers can clearly distinguish rejected students from merely unassigned ones, inspect the reason, and still manually override by adding them if needed.

In the process of implementing this PR, I got more and more annoyed at the very colorful design of buttons, alerts etc in the allocation and finalization process, so that I finally decided to make the design are little more muted (see the screenshots below). I think this now fits a lot better with the general design in the groups tab.

Since this is slice 0 of the old muesli-bowl and in there the rake tasks providing stuff to play around in the GUI with became a little chaotic, I restructured the whole demo task business. The demo workflow now uses four clearly named tasks:

- `demo:flags` enables the feature flags needed for the current slice.
- `demo:campaigns` prepares the reusable campaign examples.
- `demo:rosters` prepares roster-related demo data (without the need to manually finalize campaigns)
- `demo:setup` runs the full setup for the current slice.

This gives the demo setup a clearer structure and makes it easier to run exactly the part you need, while keeping the overall workflow consistent as more Muesli features are added in later slices.

**Note for the reviewer:** In order to get a good feeling of how it works, just run the `demo:campaigns` task and finalize all campaigns that this provides (two for Linear Algebra 2 and three for the Campaign Test Seminar).
<details>
<summary>Implementation notes</summary>

- Added policy screening via `Registration::ScreeningService` and integrated it into allocation/finalization (block on blockers, auto-reject deterministic violators).
- Persisted rejection decisions with additional columns on on `Registration::UserRegistration` (reason fields + timestamps) and track allocation decisions with the new column  `allocation_decided_at` on `Registration::Campaign`.
</details>
<details>
<summary>Below are a few screenshots</summary>

- Revised color scheme:
Screenshot 2026-05-02 134242

- Finalization auto-reject preview:
Screenshot 2026-05-02 161311

- flash message after finalization contains info on the number of rejected/unassigned students (if existent)
Screenshot 2026-05-02 134315

- new color scheme again:
Screenshot 2026-05-02 134329

- new color scheme, also note that the presentation logic for the counts (wording, given numbers) is now more coherent
Screenshot 2026-05-02 134339

- campaign dissolved pill now also contains info on rejected students (due to policy vioplations)
Screenshot 2026-05-02 134351

- rejected students get their own sidepanel
Screenshot 2026-05-02 134405

- rejected students are now also given (as a number) and in a table with rejection reason
Screenshot 2026-05-02 134510
</details>


app/models/registration/allocation_dashboard.rb
```diff
@@ -12,30 +12,56 @@ def stats
...
+        rejected_user_ids = @campaign.user_registrations
+                                     .where(status: :rejected)
+                                     .distinct
+                                     .pluck(:user_id)
+
+        Registration::AllocationStats.new(
+          @campaign,
+          assignment,
+          rejected_user_ids: rejected_user_ids
+        )
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [2b68525](https://github.com/MaMpf-HD/mampf/pull/1104/commits/2b68525b90b695327d51df442ecabf2ec415a635).

app/models/registration/user_registration.rb
```diff
@@ -57,6 +69,47 @@ class UserRegistration < ApplicationRecord
...
+      )
+    end
+
+    def clear_rejection_decision!
+      update!(
+        rejection_reason_type: nil,
+        rejection_reason_code: nil,
+        rejection_reason_label: nil,
+        rejected_at: nil
+      )
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [c88219b](https://github.com/MaMpf-HD/mampf/pull/1104/commits/c88219bcfca87d9a73799af7330cb55b3902ca71).

app/models/registration/policy/handler.rb
```diff
@@ -23,16 +23,24 @@ def summary
...
-        def pass_result(code = :ok, details = {})
+        def pass_result(code = :ok, details = {}, **metadata)
           { pass: true, code: code, details: details }
+            .merge(sanitize_result_metadata(metadata))
         end
 
-        def fail_result(code, message, details = {})
+        def fail_result(code, message, details = {}, **metadata)
           { pass: false, code: code, message: message, details: details }
+            .merge(sanitize_result_metadata(metadata))
```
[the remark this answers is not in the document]
fosterfarrell9: The signature change in `pass_result` / `fail_result` is real, but the intended result shape in this code path is top-level metadata, not nested `details`. `ScreeningService` currently consumes `classification`, `blocker_kind`, `reason_type`, `reason_code`, and `reason_label` from the top level, so rewriting the existing callers to pass an explicit third `details` hash would break the current screening/allocation flow rather than preserve it. I added focused handler specs in [a586bae](https://github.com/MaMpf-HD/mampf/pull/1104/commits/a586bae6d5ac1227216097726a90ed30bcbf11db)to make that contract explicit and prevent accidental regressions in either direction.

app/models/registration/campaign.rb
```diff
@@ -121,40 +125,79 @@ def finalize!
       with_lock do
         return if completed?
 
+        if first_come_first_served?
+          screening = Registration::ScreeningService.new(
+            self,
+            registrations: user_registrations.where.not(status: :rejected)
+          ).call
+
+          return false if screening.blocked?
```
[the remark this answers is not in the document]
fosterfarrell9: @coderabbitai What is your take on this?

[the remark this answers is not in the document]
fosterfarrell9: I adressed this in [021b0d3](https://github.com/MaMpf-HD/mampf/pull/1104/commits/021b0d36f378d3379b12b64e49d97be3ef4a2338).

app/models/registration/campaign.rb
```diff
@@ -201,6 +262,56 @@ def roster_group_type
...
+      end
+
+      def reject_pending_registrations!
+        user_registrations.pending.find_each do |registration|
+          registration.reject!(
+            reason_type: Registration::UserRegistration::REJECTION_REASON_TYPE_CAPACITY,
+            reason_code: Registration::UserRegistration::REJECTION_REASON_CODE_SOLVER_UNASSIGNED,
+            reason_label: I18n.t("registration.user_registration.reason_labels.solver_unassigned")
+          )
+        end
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [785af19](https://github.com/MaMpf-HD/mampf/pull/1104/commits/785af194bb139515cd4bc79e779eacd0250f7bf2).

app/models/registration/campaign.rb
```diff
@@ -121,40 +134,79 @@ def finalize!
...
+        if first_come_first_served?
+          screening = Registration::ScreeningService.new(
+            self,
+            registrations: user_registrations.where.not(status: :rejected)
+          ).call
+
+          raise(FinalizationBlockedError, screening) if screening.blocked?
+
+          apply_rejections!(screening.auto_reject_violations)
+        end
```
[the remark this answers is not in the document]
fosterfarrell9: Preference-based campaigns are already covered by the new screening workflow, but at allocation time rather than at finalization time.

For preference-based mode, `Registration::AllocationService#allocate!` runs `Registration::ScreeningService`, blocks on blocker violations, and applies deterministic auto-rejections before the solver runs. After `allocation_decided_at` is set, finalization intentionally materializes that already-reviewed allocation result without re-screening live user data again. The specs currently lock this behavior in as well.

So the asymmetry with FCFS is deliberate: FCFS has no separate allocation phase, so finalization is where screening must happen there. Re-screening preference-based campaigns again at finalization would be a semantic change, because it would allow post-allocation user-data changes to invalidate an allocation that was already computed and reviewed.

app/frontend/registration/allocations/_policy_violations_alert.html.erb
```diff
@@ -11,8 +21,9 @@
   — <%= t("registration.allocation.errors.policy_violation_desc") %>
 
 
-  <% grouped = dashboard.violations_by_user %>
-  <% policy_counts = dashboard.violation_counts_by_policy %>
+  <% grouped = blocker_violations.group_by { |violation| violation[:user_id] } %>
+  <% policy_counts = blocker_violations.group_by { |violation| violation[:policy] }
+                                     .transform_values(&:size) %>
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [fe07252](https://github.com/MaMpf-HD/mampf/pull/1104/commits/fe072521c8330ea71935fddb20a799793abe9920).

app/frontend/registration/allocations/_stats_summary.html.erb
```diff
@@ -1,27 +1,79 @@
...
+     } %>
+  <% if dashboard.stats.rejected_users.positive? %>
+    <% summary_items << {
+         class: "text-danger fw-medium",
+         value: [
+           dashboard.stats.rejected_users,
+           t("registration.allocation.stats.currently_rejected_inline")
+         ].join(" ")
+       } %>
+  <% end %>
```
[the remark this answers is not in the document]
fosterfarrell9: This is deliberate. FCFS has no waiting-state bucket analogous to preference-based unassigned users; those cases are modeled differently (using cohorts that don't propagate to the lecture roster). I removed the leftover FCFS ‘currently without place’ wording that made the omission look accidental in [0ccf44e](https://github.com/MaMpf-HD/mampf/pull/1104/commits/0ccf44e999386985ceec622c9fc0287dbd2defcc).

app/models/registration/allocation_dashboard.rb
```diff
@@ -114,5 +166,15 @@ def calculate_demand_per_item
...
       end
+
+      def rejected_registrations_by_user
+        @rejected_registrations_by_user ||= @campaign.user_registrations
+                                                     .where(
+                                                       status: :rejected,
+                                                       user_id: stats.rejected_user_ids
+                                                     )
+                                                     .to_a
+                                                     .group_by(&:user_id)
```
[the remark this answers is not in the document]
fosterfarrell9: I address this in [0d4f114](https://github.com/MaMpf-HD/mampf/pull/1104/commits/0d4f114301236f2a452df26f5d68b043d73191e7).

config/locales/registration/de.yml
```diff
@@ -12,6 +12,14 @@ de:
...
       reopened: "Anmeldung wiedereröffnet."
       finalized: "Anmeldeverfahren erfolgreich abgeschlossen."
+      finalization_summary:
+        rejected:
+          one: "1 Person hat derzeit eine abgelehnte Anmeldung."
+          other: "%{count} Studierende haben derzeit eine abgelehnte Anmeldung."
+        unassigned:
+          one: "1 Person konnte nicht zugeteilt werden."
+          other: "%{count} Studierende konnten nicht zugeteilt werden."
+        manual_addition: "Nachträgliches manuelles Hinzufügen ist möglich."
```
Splines, line 22: for singular you've used "person", but "Studierende" for plural. Maybe stick to "student" ("Studi" in German)

app/models/registration/screening_service.rb
```diff
@@ -0,0 +1,120 @@
+module Registration
+  class ScreeningService
```
Splines, line 6: add a short docstring on this new class about _why_ we have it and what it does (summary). Same for other classes / service objects, e.g. the finalization guard.

config/locales/registration/de.yml
```diff
@@ -85,10 +92,10 @@ de:
...
-        force_finalize: "Bist Du sicher? Dies ignoriert alle Richtlinienverstöße und speichert die Zuteilung. Auch Studierende, die gegen Richtlinien verstoßen, werden übernommen."
         reallocate: "Möchtest Du die Zuteilung wirklich neu berechnen? Die aktuelle Verteilung wird dabei überschrieben und kann sich ändern."
         complete_planning: "Bist Du sicher? Dies markiert die Planung als abgeschlossen. Es werden keine Listen verändert."
-        user_registration_conflict_remove: "Are you sure to remove the registration for user with email %{email} from this campaign? You cannot undo this action."
+        user_registration_reject_for_user: "Möchtest Du die Anmeldungen von %{email} ablehnen? Die Anmeldungen bleiben erhalten und können später erneut bearbeitet werden."
+        user_registration_defer_due_to_blocker: "Möchtest Du die Anmeldung von %{email} für jetzt manuell ablehnen? Die Anmeldung bleibt erhalten, erscheint in der Liste der abgelehnten Anmeldungen und kann später erneut bearbeitet werden."
```
Splines: "Für jetzt" is not a good translation of "for now" in my opinion. What about "vorläufig" or "vorerst"?

lib/demo/campaign_setup_support.rb
```diff
@@ -0,0 +1,682 @@
...
+        setup_preference_campaign!
+        seed_preference_campaign_registrations!
+        setup_mixed_fcfs_campaign!
+        seed_mixed_fcfs_campaign_registrations!
+        setup_two_stage_campaign!
+      end
+    end
+
+    def setup_preference_campaign!
+      ensure_non_production!
```
Splines, line 25: It seems very important to ensure this for every method here that we don't accidentally run this in production. To further avoid any mistakes (e.g. when adding new methods here), can we add a hook that executes this check before every method automatically?

app/controllers/registration/user_registrations_controller.rb
```diff
@@ -89,19 +89,43 @@ def no_registrations_found?(registrations)
...
+        registrations = registrations.where.not(status: :rejected).to_a
+        count = registrations.size
+        reason_code, reason_label = rejection_reason
+
+        Registration::UserRegistration.transaction do
+          registrations.each do |registration|
+            registration.reject!(
+              reason_type: Registration::UserRegistration::REJECTION_REASON_TYPE_MANUAL,
+              reason_code: reason_code,
+              reason_label: reason_label
```
Splines: I'm not sure I understand why we pass the `reason_label` dynamically? The mapping `reason_code` to `reason_label` is bijective, isn't it? So it is known already at "compile-time". If this is true, I'd rather prefer to store this mapping in a map at `Registration::UserRegistration`. And then, dynamically use that stored label to call `t(<stored_label_from_map>)`.

app/models/registration/user_registration.rb
```diff
@@ -57,6 +69,49 @@ class UserRegistration < ApplicationRecord
...
+
+      policy_key = "registration.policy.errors.#{translated_code}"
+      return I18n.t(policy_key) if I18n.exists?(policy_key)
+
+      reason_key = "registration.user_registration.reason_labels.#{translated_code}"
+      return I18n.t(reason_key) if I18n.exists?(reason_key)
+
+      reason_label
+    end
+
```
Splines: related to my other comment about the bijection between campaign code and reason label

app/models/registration/finalization_guard.rb
```diff
@@ -1,89 +1,76 @@
...
+        auto_reject_violations = policy_violations.select do |violation|
+          violation[:classification] == ScreeningService::CLASSIFICATION_AUTO_REJECT &&
+            blocked_registration_ids.exclude?(violation[:registration_id])
+        end
+
+        auto_reject_violations.group_by { |violation| violation[:registration_id] }
+                              .values
+                              .map(&:first)
+      end
+    end
```
Splines: This is duplicating logic from the ScreeningService.

app/models/registration/campaign.rb
```diff
@@ -121,40 +134,79 @@ def finalize!
...
 
+        if first_come_first_served?
+          screening = Registration::ScreeningService.new(
+            self,
+            registrations: user_registrations.where.not(status: :rejected)
+          ).call
+
+          raise(FinalizationBlockedError, screening) if screening.blocked?
+
+          apply_rejections!(screening.auto_reject_violations)
```
Splines, line 159: Can you explain why we don't want to apply auto-reject-violations if screening is blocked (thus raising here)?

app/models/registration/finalization_guard.rb
```diff
@@ -1,89 +1,76 @@
 module Registration
   class FinalizationGuard
```
Splines, line 4: I'm not 100% sure I get the role of FinalizationGuard vs. ScreeningService (especially with the duplicated code mentioned below).

Splines (changes requested):
I've looked at major parts of this PR, just not the specs, nor the frontend yet. This feels like a really nice change to make the finalization process and results from it more comprehensible for users.

In general, I like the proposed abstraction but still have some remarks regarding general design choices as well as more scoped things in the code. This is a first review round.

config/locales/registration/de.yml
fosterfarrell9: I changed the wording in [d175231](https://github.com/MaMpf-HD/mampf/pull/1104/commits/d175231718004437b34c0ecb10490cd738168717).

config/locales/registration/de.yml
fosterfarrell9: Indeed. I went with "vorläufig" in [91f458a](https://github.com/MaMpf-HD/mampf/pull/1104/commits/91f458af19f5a7c186e9cc6f8f2861a5d050c0a6).

app/models/registration/screening_service.rb
fosterfarrell9: Added a docstring in [993e207](https://github.com/MaMpf-HD/mampf/pull/1104/commits/993e20749f8f2f59b0b7156f92eaaae0d8e75321).

app/models/registration/finalization_guard.rb
fosterfarrell9: Indeed.- I removed the duplication in [993e207](https://github.com/MaMpf-HD/mampf/pull/1104/commits/993e20749f8f2f59b0b7156f92eaaae0d8e75321). See also the other comment on the difference between the two classes.

app/models/registration/finalization_guard.rb
fosterfarrell9: The duplicated code indeeed made the distinction murky. I removed the duplicateion in [993e207](https://github.com/MaMpf-HD/mampf/pull/1104/commits/993e20749f8f2f59b0b7156f92eaaae0d8e75321), and I would define their responsibilities as:

- `ScreeningService` is the bulk finalization screener that collects and classifies violations across registrations. It answers:  "Across this set of registrations, which violations exist, and how should they be classified?"
- `FinalizationGuard` is pre check for the finalization action, combining campaign-state checks with screening where needed. It answers: "May finalization be attempted right now?".

I hope this makes it clearer. I think that these two responsibilities are different enough that they each deserve a class.

app/models/registration/campaign.rb
fosterfarrell9: `FinalizationGuard` only does a pre check for the UI. It can say “finalization looks possible right now”, but it does not lock the campaign state. That is why `finalize!` runs the screening again during the actual operation. If something has changed in the meantime, finalization may still need to stop.

In that situation, it seems safer to me not to apply automatic rejections first. Otherwise you could end up in a half-finished state: for example, 8 students are already auto-rejected, but then finalization still fails because 2 other students now have blocker issues that need manual clarification. The teacher would then see a failed finalization, but with some registrations already changed in the background. The current behavior avoids that: if blockers are present at the moment of the real finalization run, nothing is changed yet, and the teacher can first resolve or defer the blocker cases and then retry cleanly.

app/models/registration/user_registration.rb
fosterfarrell9: See my other comment o the bijection above.

app/controllers/registration/user_registrations_controller.rb
fosterfarrell9: I agree for the fixed built-in reasons, and I cleaned that part up: the manual reasons and `solver_unassigned` now derive their label centrally in `Registration::UserRegistration` instead of translating it separately at each call site.

I kept the broader combined `reason_code` / `reason_label` model where `reason_label` serves as a fallback though, because it is still useful for robustness: stored rejection reasons remain displayable even if a later PR renames a code, changes a translation key, removes an alias, or introduces a new policy reason before centralizing it.

lib/demo/campaign_setup_support.rb
fosterfarrell9: Shouldnt it be suffcient that we make these methods private?

Splines: I mean we could still accidentally run it in production then, even if the methods are private, I don't see how that can change it.

But I agree that this might be overkill as I don't see a simple solution to automatically add this to every statement. Could you maybe still add the check on the respective methods of `demo_setup.rake`? And make sure that all methods individually have that as well. E.g. what about `setup_rosters!`, it is not protected right now I guess.

app/models/registration/finalization_guard.rb
Splines: Nice, thanks!

app/models/registration/finalization_guard.rb
Splines: yes thanks, that makes it clearer

app/models/registration/finalization_guard.rb
```diff
@@ -1,89 +1,87 @@
+# Checks whether the registration campaign can be finalized and performs the
+# necessary checks and screenings.
 module Registration
   class FinalizationGuard
-    Result = Struct.new(:success?, :error_code, :error_message, :data, keyword_init: true)
+    Result = Struct.new(:success?,
+                        :error_code,
+                        :error_message,
+                        :data,
```
Splines: Can we also call this `violations` here? Everything is "data" somehow, it's a very generic word.

app/models/registration/campaign.rb
Splines: ah ok, thanks that makes it clear now. Maybe add 1-2 sentences here as a comment as I don't think this reasoning is trivial when looking at the code

app/frontend/registration/allocations/_policy_violations_alert.html.erb
```diff
@@ -55,33 +94,33 @@
...
-                        email: first[:email])
-                      },
-                      form: { class: "d-inline-block" },
-                      title: t("registration.user_registration.actions.remove_user") do %>
-          <i class="bi bi-trash"></i>
+        <% if violations.none? { |violation| violation[:blocker_kind] == Registration::ScreeningService::BLOCKER_KIND_CONFIGURATION } %>
+          <%= button_to reject_for_user_registration_campaign_registrations_path(campaign,
+                                                                                user_id: user_id,
+                                                                                source: allocation_source,
+                                                                                reason: Registration::UserRegistration::REJECTION_REASON_CODE_DEFERRED_DUE_TO_BLOCKER),
```
Splines: these lines are too long

app/models/registration/campaign.rb
```diff
@@ -201,6 +255,35 @@ def unassigned_users(preload_registrations: false)
...
+        ]
+      ).order(:name, :email)
+    end
+
+    def open_rejected_registrations
+      user_registrations.rejected
+                        .where(
+                          "rejection_reason_code IS NULL OR rejection_reason_code != ?",
+                          Registration::UserRegistration::REJECTION_REASON_CODE_SOLVER_UNASSIGNED
+                        )
```
Splines: is there a more Rails-like way to write this `where` statement?

app/controllers/registration/user_registrations_controller.rb
Splines: I get the point about robustness, but I still don't like the current implementation.

In my mental modal, it would be easier if the `reject!` method of `user_registration`, could do the lookup "code -> label". Then, we don't scatter this around to individual handlers and the actual lookup would happen at the place where the respective keys are defined.

Then, for your robustness concern: we could still store the label in the DB (if you insist), but only use that code as fall-back to the stored label if the code is no longer resolvable. <sub>In any case: yes, those translation keys shouldn't be renamed in the best case, and that's why we should always check for occurrences in the whole code base for those keys when removing/updating them. This is valid for any i18n key, so not sure if we really need this special handling here.</sub>

Furthermore, we now have the methods `built_in_rejection_reason_label` and `localized_rejection_reason_label`, which do similar things. When moving the lookup to the `user_registration`, we could get rid of those methods as well. In addition, I don't get why you have to re-expose `def localized_rejection_reason_label` and just call `self.class.the_very_same_method`?

If you agree on moving this lookup to the model itself (`user_registration`), please also add specs for these cases: known code without a stored label, known code with renamed translation key, unknown legacy code with stored fallback label.

app/frontend/registration/allocations/allocation.css
```diff
@@ -44,3 +44,103 @@
...
+    background: rgba(var(--bs-danger-rgb), 0.12);
+    border: 1px solid var(--bs-danger-border-subtle);
+}
+
+.allocation-action-primary:hover,
+.allocation-action-primary:focus {
+    color: var(--bs-danger-text-emphasis);
+    background: rgba(var(--bs-danger-rgb), 0.2);
+    border-color: rgba(var(--bs-danger-rgb), 0.28);
+}
```
Splines: please use the CSS nesting selector `&` to group those

app/frontend/registration/allocations/_stats_summary.html.erb
```diff
@@ -1,27 +1,79 @@
...
+       class: [
+         dashboard.stats.unassigned_users > 0 ? "text-danger" : "text-muted",
+         "fw-medium"
+       ].join(" "),
+       value: [
+         dashboard.stats.unassigned_users,
+         t("registration.allocation.stats.unassigned_inline")
+       ].join(" ")
+     } %>
+<% end %>
```
Splines: This is too much Ruby logic for a html.erb file.

app/frontend/registration/campaigns/_dissolved_footnote.html.erb
```diff
@@ -49,9 +75,17 @@
           data-roster-panel-path="<%= unassigned_registration_campaign_path(campaign, source: :panel, format: :turbo_stream) %>"
         >
           <i class="bi bi-exclamation-triangle-fill text-warning"></i>
-          <span><%= unassigned %> <%= t('roster.candidates.title') %></span>
+
+          <span>
+            <%= unassigned %> <%= t('roster.candidates.short_title', default: 'unplaced') %>
```
Splines: get rid of the default value for the i18n string

Splines: same for other occurences in this PR

app/frontend/roster/roster_drag.controller.js
```diff
@@ -119,7 +119,7 @@ export default class extends Controller {
     const targetTitle = tile.dataset.rosterTitle;
     const targetAddPath = tile.dataset.rosterAddMemberPath;
 
-    if (this.sourceTypeValue === "unassigned") {
+    if (["unassigned", "rejected"].includes(this.sourceTypeValue)) {
```
Splines: this includes check is used twice in the file -> outsource to a method

app/helpers/application_helper.rb
```diff
@@ -65,6 +65,14 @@ def progress_bar(value, max, classification: :neutral, label: nil,
...
                   when :neutral
                     "bg-primary"
+                  when :allocation_first
+                    "allocation-progress-bar allocation-progress-bar--first"
+                  when :allocation_second
+                    "allocation-progress-bar allocation-progress-bar--second"
+                  when :allocation_other
+                    "allocation-progress-bar allocation-progress-bar--other"
+                  when :allocation_forced
+                    "allocation-progress-bar allocation-progress-bar--forced"
```
Splines: Don't like having this in the application_helper too much. Especially since this crates an implicit dependency: whenever you use the `progress_bar`, you may need to make sure that you include the stylesheets that have these classes like `allocation-progress-bar` defined.

I'd prefer adding a method to the application helper that can clamp a percentage for us, then use that in a custom progress bar. If we need some global styling for all progress bars, that could go into a global CSS class then.

Splines: same for `utilization_color`

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2026_05_01_000001) do
+ActiveRecord::Schema[8.0].define(version: 2026_05_24_000001) do
```
Splines: please upgrade to a date after the 13th of June to better reflect the commit history on next

Splines (changes requested):
some more comments

app/helpers/application_helper.rb
fosterfarrell9: I refactored this in [0441835](https://github.com/MaMpf-HD/mampf/pull/1104/commits/0441835790ff870e9ca2f0c1579c5075f640263c) for `progress_bar`.

app/controllers/registration/user_registrations_controller.rb
fosterfarrell9: I moved this to the `user_registration` model in [be622d5](https://github.com/MaMpf-HD/mampf/pull/1104/commits/be622d53035ff65b1ffce79df9dadf27e3a2295f) and I hid the duplicate comment on this further below.

lib/demo/campaign_setup_support.rb
fosterfarrell9: Fair enough. I just added the non-production check at the top of each of the affected methods in [1e7c411](https://github.com/MaMpf-HD/mampf/pull/1104/commits/1e7c41148d2397b48d0bbeda2362e6b57c262c2d).

db/schema.rb
fosterfarrell9: Fixed in [f00dcb9](https://github.com/MaMpf-HD/mampf/pull/1104/commits/f00dcb92729e639e322dd4fef80d02f9691828c7).

app/models/registration/finalization_guard.rb
fosterfarrell9: Fixed in [6e35ee7](https://github.com/MaMpf-HD/mampf/pull/1104/commits/6e35ee756c982c5454148fcd41fa1122729dcec2).

app/models/registration/campaign.rb
fosterfarrell9: I rewrote it in [b52448d](https://github.com/MaMpf-HD/mampf/pull/1104/commits/b52448d62dbeba46e4a4b71632f3a9f3c395da28).

app/frontend/registration/allocations/_stats_summary.html.erb
fosterfarrell9: Indeed. I extracted the logic to `AllocationDashboard`.

app/frontend/registration/allocations/_policy_violations_alert.html.erb
fosterfarrell9: Fixed in [541a8c3](https://github.com/MaMpf-HD/mampf/pull/1104/commits/541a8c3e7b21145a381eec9f9fd36ffa63a7fee6).

app/frontend/registration/allocations/allocation.css
fosterfarrell9: Fixed in [844395c](https://github.com/MaMpf-HD/mampf/pull/1104/commits/844395c975030487097d8894f855e6dfd37557e4).

app/frontend/registration/campaigns/_dissolved_footnote.html.erb
fosterfarrell9: Fxied in [731d3aa](https://github.com/MaMpf-HD/mampf/pull/1104/commits/731d3aa635715556f74906d43f0303c39fd0b376).

app/frontend/roster/roster_drag.controller.js
fosterfarrell9: Fixed in [cb96e55](https://github.com/MaMpf-HD/mampf/pull/1104/commits/cb96e55e62e4f705ea6ca74cd1389ab0b6b2ee3f).

app/controllers/registration/user_registrations_controller.rb
Splines: thanks! (by the way, this is not a duplicate comment, it's just that GitHub displays this a bit weird down below, since I didn't reply directly to the comment, but instead via the "Request changes" review)

app/frontend/registration/allocations/_stats_summary.html.erb
Splines: Thanks, however you outsourced this to a model, and not to some ViewComponent, helper (or controller).

According to the [MVC-patterns](https://guides.rubyonrails.org/getting_started.html#model-view-controller-basics) we use so far in our code base, models should never be responsible to set a CSS class, which is by design frontend-functionality. Please move this out of the model to preferably some view component or a helper.

app/helpers/application_helper.rb
```diff
@@ -48,31 +48,32 @@ def hide(value)
     value ? "none;" : "block;"
   end
 
+  def clamped_percentage(value, max)
+    return 0 if max.to_i <= 0
+
+    # rubocop:disable Style/FloatDivision
+    (value.to_f / max.to_f * 100).clamp(0, 100)
+    # rubocop:enable Style/FloatDivision
```
Splines: Is there a specific reason to not follow the RuboCop advice here?

Splines (changes requested):
Thanks, here is the probably last review round. Please also note the last open item from the last round [here](https://github.com/MaMpf-HD/mampf/pull/1104#discussion_r3438896865). Also, there are some linting errors.

app/frontend/registration/allocations/_stats_summary.html.erb
fosterfarrell9: I moved the css stuff to a helper in [4bd0627](https://github.com/MaMpf-HD/mampf/pull/1104/commits/4bd0627707436713042bd5e2948ff33ee053df93).

app/helpers/application_helper.rb
fosterfarrell9: Fixed in [6dd5762](https://github.com/MaMpf-HD/mampf/pull/1104/commits/6dd5762b12780b1a785272cb6f4e83a5ffb88a2a).

Splines (approved):
Nice, great new touches for the finalization process!

PR #1115, opened 2026-05-06
Upgrade ruby (Docker hardened image)

Splines:
not ready for review yet even though flagged like this just to be able to run through some actions and hardened docker images


fosterfarrell9:
I would say we should close this: Ruby update happens in #1151, and I am now less convinced of the docker hardened images than before - the new multistage build with slim-trixie as container in the end in #1151 already reduces the surface significantly.

fosterfarrell9:
Closed in favor of #1151 (but docker hardened images may make a comeback at a later point in time).

PR #1116, opened 2026-05-10
PR-8.6 — Submission GUI simplification

plienta:
This pull request 
* removes the ability for tutors to manually move submissions between tutorials
* when roster maintenance is enabled + lecture potentially rosterizable (any tutorials of it has rosters || belongs to a campaign || enable self-roster): 
  * if student is in roster -> automatically assign tutorials
  * else -> require student to enroll

**Removal of manual submission movement:**
- Eliminated the ability for users to manually select or move a submission to a different tutorial, including the removal of related controller actions (`select_tutorial`, `move`), routes, views, permissions, and UI buttons. 

**Automatic tutorial assignment with roster maintenance:**
- Updated `submission_create_params` and `submission_update_params` to automatically assign the user's rosterized tutorial when the roster maintenance feature is enabled and the lecture has rosterized tutorials, overriding manual selection.

**Submission form and UI improvements:**
- Updated the submission form (`_form.html.erb`) to display the user's assigned tutorial or prompt them to enroll if necessary, instead of allowing manual selection when roster maintenance is active.


app/models/lecture.rb
```diff
@@ -867,6 +867,10 @@ def roster_entries
     lecture_memberships
   end
 
+  def has_rosterized_tutorials?
+    tutorials.any? { |t| t.rosterized? }
```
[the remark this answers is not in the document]
plienta: changed into
```
tutorials.joins(:tutorial_memberships).exists? ||
    tutorials.joins(:registration_items).exists? ||
    tutorials.where.not(self_materialization_mode: "disabled").exists?
```
in e20b3c65ae41f3f678f0ea733d3b3c9024a7eada

app/models/tutorial.rb
```diff
@@ -70,6 +70,12 @@ def materialize_allocation!(user_ids:, campaign:)
     end
   end
 
+  def rosterized?
+    tutorial_memberships.exists? ||
+      in_campaign? ||
+      (!self_materialization_mode.nil? && self_materialization_mode != "disabled")
```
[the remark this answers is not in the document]
plienta: applied same suggestion, in a0a4849f57ac05c9f0fb02876cc4dce5e76bd837

app/views/submissions/_form.html.erb
```diff
@@ -5,14 +5,29 @@
...
+          <% if current_user.tutorial_rosterized(lecture) %>
+            <p>
+              <%= current_user.tutorial_rosterized(lecture)&.title_with_tutors %>
+            </p>
+          <% else %>
+          <!-- if not, require student to enroll first-->
+            <p>
+              <%= t('submission.enroll_first') %>
+            </p>
+          <% end %>
```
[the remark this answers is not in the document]
plienta: [hiding the submit action when roster eligible for current lecture but user has not enrolled](https://github.com/MaMpf-HD/mampf/pull/1116/changes/144c590ef6ac713ef59dbff1d1f682e65bdee502)

spec/controllers/submissions_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:tutorial)    { create(:tutorial, lecture: lecture) }
+  let(:submission)  { create(:submission, assignment: assignment, tutorial: tutorial) }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+    allow(Flipper).to receive(:enabled?).with(:roster_maintenance).and_return(true)
+    allow_any_instance_of(Lecture).to receive(:has_rosterized_tutorials?).and_return(true)
+    allow(user).to receive(:tutorial_rosterized).and_return(tutorial)
+    controller.params = ActionController::Parameters.new(submission: { tutorial_id: tutorial.id,
+                                                                       assignment_id: assignment.id })
```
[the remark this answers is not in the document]
plienta: fixed in 922768976197aa13bfedf101d9031c9df427b62c

app/views/submissions/update.coffee
```diff
@@ -3,10 +3,6 @@
 # clean up from previous error messages
 $('#submission-tutorial-error').empty()
 # display error message
-<% if @errors[:tutorial].present? %>
-$('#submission-tutorial-error')
-  .append('<%= @errors[:tutorial].join(" ") %>').show()
-<% end %>
 <% if @errors[:manuscript].present? %>
 alert('<%= @errors[:manuscript].join(" ") %>')
 <% end %>
```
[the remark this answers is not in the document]
plienta: re-added in bf26987b1dd54e5f1840ad672883f94d70c279f5

app/controllers/submissions_controller.rb
```diff
@@ -258,12 +247,26 @@ def set_submission
...
     # disallow modification of assignment
     def submission_update_params
-      params.expect(submission: [:tutorial_id])
+      permitted = params.expect(submission: [:tutorial_id])
+      lecture = @submission.assignment.lecture
+      if Flipper.enabled?(:roster_maintenance) && lecture&.has_rosterized_tutorials?
+        permitted[:tutorial_id] = current_user.tutorial_rosterized(lecture)&.id
+      else
+        params.expect(submission: [:tutorial_id])
+      end
```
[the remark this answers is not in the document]
plienta: apply suggested change in 806e592bb7eecc1f9946b2462909ab6568465c66

spec/controllers/submissions_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+describe SubmissionsController do
```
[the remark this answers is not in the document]
plienta: added cmt in 59a21b5db5db8b8fe49df955e6644ff956e5fd76

spec/controllers/submissions_controller_spec.rb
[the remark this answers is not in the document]
plienta: use `other_tutorial` in 59a21b5db5db8b8fe49df955e6644ff956e5fd76

spec/models/lecture_spec.rb
```diff
@@ -134,6 +134,23 @@
...
+      tut1 = create(:tutorial, lecture: lecture, self_materialization_mode: "add_and_remove")
+      tut2 = create(:tutorial, lecture: lecture, self_materialization_mode: "disabled")
+      allow(tut1).to receive(:rosterized?).and_return(true)
+      allow(tut2).to receive(:rosterized?).and_return(false)
+      expect(lecture.has_rosterized_tutorials?).to eq(true)
+    end
+
+    it "returns false if no tutorials are rosterized" do
+      allow_any_instance_of(Tutorial).to receive(:rosterized?).and_return(false)
+      expect(lecture.has_rosterized_tutorials?).to eq(false)
```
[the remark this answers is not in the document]
plienta: refactored this test in 61112f3cea93de2831febed876afbc5b61fa2dea and 922768976197aa13bfedf101d9031c9df427b62c

app/views/submissions/update.coffee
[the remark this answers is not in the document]
plienta: in fact, all changes relating to update.coffee has been reverted

app/views/submissions/_other_actions.html.erb
```diff
@@ -1,7 +1,3 @@
-<%= link_to t('buttons.move'),
```
fosterfarrell9, line 1: The "Move" button (with the `select_tutorial`/`move` actions, routes, `_select_tutorial` partial and `select_tutorial.js.erb`) is a **tutor** tool — gated on `user.in?(submission.tutorial.tutors)` — to reassign a submission to another tutorial (e.g. a student filed under the wrong one). Deleting it removes it for **non-roster** lectures too, where nothing replaces it: the edit form's tutorial select belongs to the *submitter* (`submission.users`), not the tutor. So staff can no longer move a mis-filed submission.

Keep the machinery and gate it on roster instead of deleting:
- wrap this button in `<% unless enabled_roster_for_lecture?(lecture) %>`;
- restore the `select_tutorial`/`move` actions + routes + `_select_tutorial.html.erb` + `select_tutorial.js.erb` (`move_params`/`move.coffee` become used again, and `mampf_routes.js` no longer needs regenerating);
- for defense in depth, gate the ability so a roster lecture can't be moved via a direct request: `user.in?(submission.tutorial.tutors) && !enabled_roster_for_lecture?(submission.assignment.lecture)` — otherwise a tutor `move` sets a `tutorial_id` inconsistent with the roster (the `submission_update_params` override only fires on student edits).

Also check `new.coffee` (−8): if those lines were the TomSelect init for the tutorial dropdown, they're still needed for the non-roster dropdown — condition rather than delete.

app/controllers/submissions_controller.rb
```diff
@@ -265,12 +254,25 @@ def set_submission
...
 
     def submission_create_params
-      params.expect(submission: [:tutorial_id, :assignment_id])
+      permitted = params.expect(submission: [:tutorial_id, :assignment_id])
+      assignment_id = params[:submission][:assignment_id]
+      assignment = Assignment.find_by(id: assignment_id)
+      lecture = assignment&.lecture
+
+      if Flipper.enabled?(:roster_maintenance) && lecture&.roster_eligible_tutorials?
+        permitted[:tutorial_id] = current_user.tutorial_rosterized(lecture)&.id
```
fosterfarrell9: When the lecture is roster-eligible but `tutorial_rosterized` returns nil (student not on any tutorial roster), this sets `permitted[:tutorial_id] = nil`, and since `submissions.tutorial_id` is `null: false` / `belongs_to :tutorial`, create/update fails with a generic "tutorial must exist" error. The UI handles it (save hidden, "enroll first"), so it only bites a direct POST — no data loss or security hole — but a clear server-side guard/message (bail early in `create` with an "enroll first" flash) would beat relying on the UI + a confusing validation error.

app/models/lecture.rb
```diff
@@ -916,6 +916,14 @@ def roster_entries
     lecture_memberships
   end
 
+  # Mirrors Tutorial#roster_eligible? logic at the SQL level for performance.
+  # If eligibility conditions change, update both methods.
+  def roster_eligible_tutorials?
```
fosterfarrell9, line 923: `Lecture#roster_eligible_tutorials?` re-implements `Tutorial#roster_eligible?` at the SQL level, synced only by the "update both methods" comment. The mirror is correct today (`in_campaign?` = `registration_items.exists?`; the `prefix: true` enum makes `self_materialization_mode_disabled?` match `where.not(self_materialization_mode: "disabled")`), but fragile. A shared `Tutorial` scope (e.g. `Tutorial.roster_eligible`) that both the instance predicate and the lecture-level `exists?` build on would keep the condition in one place.

app/views/submissions/_card_main.html.erb
```diff
@@ -132,6 +132,9 @@
       <%= t('submission.nothing_here') %>
     <% elsif assignment.active? %>
       <%= t('submission.nothing_here') %>
+    <% elsif enabled_roster_for_lecture?(assignment.lecture) &&
```
fosterfarrell9: `enabled_roster_for_lecture?` runs up to three `exists?` queries (`roster_eligible_tutorials?`) plus one for `tutorial_rosterized`, called inline per assignment. `_form.html.erb` computes both once at the top (lines 1–2); do the same here (or once per lecture) so a page with many assignments doesn't fire ~4×N queries.

app/controllers/submissions_controller.rb
```diff
@@ -265,12 +254,25 @@ def set_submission
     end
 
     def submission_create_params
-      params.expect(submission: [:tutorial_id, :assignment_id])
+      permitted = params.expect(submission: [:tutorial_id, :assignment_id])
+      assignment_id = params[:submission][:assignment_id]
```
fosterfarrell9: `params.expect(...)` already returned `permitted` with `:assignment_id`; re-reading `params[:submission][:assignment_id]` raw is redundant — use `permitted[:assignment_id]`.

spec/controllers/submissions_controller_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+  let(:user)        { create(:confirmed_user) }
+  let(:lecture)     { create(:lecture) }
+  let(:assignment)  { create(:assignment, lecture: lecture) }
+  let(:tutorial)    { create(:tutorial, lecture: lecture) }
+  let(:other_tutorial) { create(:tutorial, lecture: lecture) }
+  let(:submission) { create(:submission, assignment: assignment, tutorial: tutorial) }
+
+  before do
+    sign_in user
+    allow_any_instance_of(User).to receive(:tutorial_rosterized).and_return(tutorial)
```
fosterfarrell9: Good coverage of `submission_create_params`/`submission_update_params` across flag/eligibility combos, but it leans entirely on `allow_any_instance_of` (`roster_eligible_tutorials?` / `tutorial_rosterized`) — testing the param methods in isolation, never a real create/update. Two gaps: (1) the nil case (roster enabled, `tutorial_rosterized` → nil) isn't exercised — that's the `tutorial_id = nil` edge; (2) no request/integration test that an enrolled student's submission lands on their roster tutorial and a non-enrolled one is blocked — the security-relevant behavior.

fosterfarrell9 (changes requested):
The core design is sound but the inline comments need to be resolved before this merges; The biggest: removing `move`/`select_tutorial` drops a tutor's ability to reassign a mis-filed submission on **non-roster** lectures with nothing to replace it, which is a regression — keep the machinery and just hide the button for roster lectures (that also makes the "dead code" and route-regeneration concerns moot).

app/controllers/submissions_controller.rb
plienta: Good catch. I have cleaned this in 1af8ff24a6ba4b0063ab52b29da382667bafdfe2

app/controllers/submissions_controller.rb
plienta: Thanks! Added an explicit guard now. Both submission_create_params and submission_update_params raise a TutorialNotRosteredError when the lecture is roster-eligible but the current user has no roster tutorial entry. 1af8ff24a6ba4b0063ab52b29da382667bafdfe2

app/models/lecture.rb
plienta: Agreed, that's fragile as-is. Added a Tutorial.roster_eligible scope built to mirror roster_eligible?, and switched Lecture#roster_eligible_tutorials? to use this scope. 687257ba0eb2dbbfaccf96a2306c5ec89cd93041

app/views/submissions/_card_main.html.erb
plienta: Good catch, thanks. For this, I have added per-request cache `roster_cache` on the controller and exposed via `helper_method` so repeated lookups for the same lecture should hit the cache. 6e64d5365ea7a57329fe98c3a82b06e0291ba430  and  d4968c1635d1c22e593225bc5e3adfdee4fc26a6

spec/controllers/submissions_controller_spec.rb
plienta: Good points, thanks. Removed stubs and added a request spec covering both cases: an enrolled student's submission lands on their roster tutorial, and a non-enrolled student's submission in a rosterized lecture is blocked. 1af8ff24a6ba4b0063ab52b29da382667bafdfe2

app/views/submissions/_other_actions.html.erb
plienta: For this, all the `move`/`select_tutorial` changes have been reverted. Added checks for: displaying the "Move" button, the `move` ability itself, and the TomSelect init for the tutorial dropdown — each now conditioned on whether the lecture is roster-eligible. 4c0bbe9c66316f786ca1da67d621a288d4ca05ca ea90fce6cb295abc862a959563660505fec7b525 3b67d5a915b20f1a9eb9404c254283a4404682a7 4063ff64fb5ea5795aa39aa12f91b045c7d78347

app/models/lecture.rb
fosterfarrell9: Nice! Just one thing:  It appears that the duplication only moved: `Tutorial#roster_eligible?` (tutorial.rb:95) still uses same three conditions, and repo-wide it has no callers outside `tutorial_spec.rb`. I would suggest we drop it.

app/views/submissions/_card_main.html.erb
fosterfarrell9: Nice solution. Still N+1 on the tutorials page though: `tutorials/index` → `_rows` → `_submission_row:50` →`_other_actions:1` calls `enabled_roster_for_lecture?` once per submission row. `helper_method :roster_cache` only reaches views rendered by SubmissionsController (`TutorialsController._helpers` doesn't have it), so `respond_to?(:roster_cache)` is false there and every row falls into the uncached branch.

Memoizing on an instance variable in the helper fixes it everywhere at once: the view object has the same per-request lifetime as the controller, but every partial in the response shares it, whichever controller rendered it.

```ruby
def enabled_roster_for_lecture?(lecture)
  @roster_enabled ||= {}
  @roster_enabled.fetch(lecture.id) do
    @roster_enabled[lecture.id] =
      Flipper.enabled?(:roster_maintenance) && lecture.roster_eligible_tutorials?
  end
end

fosterfarrell9 (changes requested):

app/models/lecture.rb
fosterfarrell9: I see this has been fixed now.

app/views/submissions/_card_main.html.erb
fosterfarrell9: Nice fix on your side. I refactored a little and close this now as resolved.

PR #1123, opened 2026-05-14
Rework Dependabot settings & enable it

Splines:
I came up with some new config for Dependabot, let me know what you think. I've separated into these PRs:

- Every Monday, we will get the following PRs:
  - Regular dependency updates concerning Gems and npm packages (patch, minor and security)
  - Regular infrastructure dependency updates concerning Docker and GitHub Actions (patch, minor and security)
- Every 20 days:
  - Dependency updates concerning Germs and npm packages (MAJOR versions)
  - Infrastructure Dependency updates concerning Docker and GitHub Actions (MAJOR versions)

The Dependabot PRs target the `security` branch.

A future PR could tackle merging specific Dependabot PRs automatically to the `security` branch.

---

Note that I don't know if this config will actually work, we'll have to try it out ;) At least, the config is syntactically correct (see the dependabot action that reported this)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -150,7 +150,7 @@ updates:
     directory: "/"
     target-branch: "security"
     multi-ecosystem-group: "major-infrastructure"
-    ppatterns: ["*"]
+    patterns: ["*"]
     commit-message:
       prefix: "deps (infra, major): "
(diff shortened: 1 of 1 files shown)


fosterfarrell9 (approved):
Looks good! Let's try it.

PR #1129, opened 2026-05-16
Let users post identical comments

fosterfarrell9:
A teacher complained about the impossibility to post comments on the same medium with the same content, which does not really make sense: E.g. if you have a manuscript and students report errors, then you would usually write something like "Yes, thank you, I fixed it" or so - currently you cannot do this and have to write a different text each time. I have been affected by this also and I really dislike the current way.

The reason for the problem is in a validation within the commontator gem:

```ruby
class Commontator::Comment < ActiveRecord::Base
  validates :body, presence: true, uniqueness: {
    scope: [ :creator_type, :creator_id, :thread_id, :deleted_at ], message: :double_posted
  }
```

This PR monkeypatches commontator. It removes the validation and replaces it by a narrower one: Duplicates are now only rejected if they are by the same poster to the same parent comment (this is a gate against accidental double posts). Specs are included. See the two screenshots for before/after:

Screenshot 2026-05-16 125715
Screenshot 2026-05-16 125525


config/initializers/commontator.rb
```diff
@@ -325,6 +325,29 @@
...
+      Commontator::Comment._validators[:body].delete(validator)
+      Commontator::Comment.skip_callback(
+        :validate,
+        :before,
+        validator,
+        raise: false
+      )
+    end
+
     Commontator::Comment.include(Extensions::Commontator::Comment)
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [3f87161](https://github.com/MaMpf-HD/mampf/pull/1129/commits/3f871612ac96f4776e369f3c1b6c5a4ec224f46a).

config/initializers/commontator.rb
```diff
@@ -325,6 +325,29 @@
   end
 
   if ActiveRecord::Base.connection.table_exists?(:thredded_topics)
+    validators = Commontator::Comment._validators[:body].select do |candidate|
+      candidate.attributes == [:body] &&
+        (
+          candidate.is_a?(ActiveModel::Validations::PresenceValidator) ||
+          (
+            candidate.is_a?(ActiveRecord::Validations::UniquenessValidator) &&
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [3f87161](https://github.com/MaMpf-HD/mampf/pull/1129/commits/3f871612ac96f4776e369f3c1b6c5a4ec224f46a).

lib/extensions/commontator/comment.rb
```diff
@@ -4,6 +4,14 @@ module Comment
       extend ActiveSupport::Concern
 
       included do
+        validates :body, presence: true
+        validates :body, uniqueness: {
+          scope: [
+            :creator_type, :creator_id, :thread_id, :parent_id, :deleted_at
+          ],
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [876cd24](https://github.com/MaMpf-HD/mampf/pull/1129/commits/876cd24198eb05f5b8f589a5ffc12b38d0795ae5).

config/initializers/commontator.rb
```diff
@@ -315,6 +315,58 @@
...
+
+    Commontator::Comment.validates(:body, presence: true)
+    Commontator::Comment.validates(:body, uniqueness: {
+                                     scope: CUSTOM_UNIQUENESS_SCOPE,
+                                     message: :double_posted
+                                   })
+
+    return if Commontator::Comment.reflect_on_association(:annotation) &&
+              Commontator::Comment.method_defined?(:medium)
+
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [6781d52](https://github.com/MaMpf-HD/mampf/pull/1129/commits/6781d52b51631ce75fd9bf445097d0dc505fb682).

config/initializers/commontator.rb
```diff
@@ -324,7 +376,5 @@
     next
   end
 
-  if ActiveRecord::Base.connection.table_exists?(:thredded_topics)
-    Commontator::Comment.include(Extensions::Commontator::Comment)
-  end
+  CommontatorCommentPatch.apply!
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [2aa6308](https://github.com/MaMpf-HD/mampf/pull/1129/commits/2aa63084914f77ba52817b8a0fb39cf9faa1d584).

config/initializers/commontator.rb
```diff
@@ -315,16 +315,75 @@
...
+        validator,
+        raise: false
+      )
+    end
+
+    Commontator::Comment.validates(:body, presence: true)
+    Commontator::Comment.validates(:body, uniqueness: {
+                                     scope: CUSTOM_UNIQUENESS_SCOPE,
+                                     message: :double_posted
+                                   })
```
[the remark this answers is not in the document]
fosterfarrell9: This is a patch for a gem and it relaxes the even stricter model based validation that was there before, I will definitely not add a database backed constraint here.

config/initializers/commontator.rb
```diff
@@ -315,16 +315,75 @@
...
+    Rails.logger.info(
+      "DATABASE_ADAPTER env var is #{db_adapter}. " \
+      "Skipping Commontator comment patch."
+    )
     next
   end
 
-  if ActiveRecord::Base.connection.table_exists?(:thredded_topics)
-    Commontator::Comment.include(Extensions::Commontator::Comment)
-  end
```
Splines, line 329: This was removed, but I cannot find a `Comment.include()` anymore.  Which part of `CommontatorCommentPatch` handles this?

config/initializers/commontator.rb
```diff
@@ -315,16 +315,75 @@
...
+  ].freeze
+  CUSTOM_UNIQUENESS_SCOPE = [
+    :creator_type, :creator_id, :thread_id, :parent_id, :deleted_at
+  ].freeze
+
+  def apply!
+    return unless ActiveRecord::Base.connection.table_exists?(:thredded_topics)
+
+    validators.each do |validator|
+      Commontator::Comment._validators[:body].delete(validator)
```
Splines: With this line, we remove all existing validations for the body. Don't we want to keep some of Commontator itself?

config/initializers/commontator.rb
```diff
@@ -315,16 +315,75 @@
   config.mentions_enabled = false
 end
 
+module CommontatorCommentPatch
```
Splines: Let's move this patch to some new file (if possible) since this initializer is more of a "set default values for the gem" file in my eyes.

Splines (changes requested):
Some first comments.

config/initializers/commontator.rb
fosterfarrell9: Fair enough, I extracted this in [daf62c5](https://github.com/MaMpf-HD/mampf/pull/1129/commits/daf62c5639a620c7199f36e0783085dbe8ea43af).

config/initializers/commontator.rb
fosterfarrell9: `Comment.include(...)` was removed intentionally because the patch no longer uses a concern-based extension module. `Extensions::Commontator::CommentPatch.apply!` now applies the whole monkeypatch directly to `Commontator::Comment` by re-registering the validators, adding the `annotation` association, and defining `medium`. The old `table_exists?` guard was moved into `apply!`. This avoids the reload/idempotence problems we had with including a reloadable concern repeatedly (see the whole refactoring process in the genesis of this PR through the earlier reviews by copilot).

config/initializers/commontator.rb
fosterfarrell9: This refers to validation on the Commontator::Comment model. In [commontator's source](https://github.com/lml/commontator/blob/main/app/models/commontator/comment.rb) you can see that these are the following:

```ruby
class Commontator::Comment < ActiveRecord::Base

  validates :editor, presence: true, on: :update
  validates :body, presence: true, uniqueness: {
    scope: [ :creator_type, :creator_id, :thread_id, :deleted_at ], message: :double_posted
  }
  validate :parent_is_not_self, :parent_belongs_to_the_same_thread, if: :parent
```

So, there is only this validation

```ruby
  validates :body, presence: true, uniqueness: {
    scope: [ :creator_type, :creator_id, :thread_id, :deleted_at ], message: :double_posted
  }
```

that gets deleted and then replaced by 
```rb
  validates :body, presence: true, uniqueness: {
    scope: [:creator_type, :creator_id, :thread_id, :parent_id, :deleted_at], message: :double_posted
```

Splines: nice, thanks a lot, that explains it

Splines (approved):
Nice, works like a charm!

PR #1130, opened 2026-05-18
Remove PDF compressor dependency

Splines:
We have long used the [pdfcomprezzor](https://github.com/henrixapp/pdfcomprezzor) implemented in Go using Web Workers. The package is not maintained anymore, so we remove the dependency here from MaMpf. This means that users cannot compress or merge their PDF documents within MaMpf anymore.

Maybe in the future we want to use a modern library to replace this functionality, especially since `media.coffee` needs a complete rewrite. But for now, and for the sake of security, let's remove this dependency.

My setup is too slow, so I didn't test rebuilding the Docker image. As we will merge this to `next`, maybe it's worth not rebuilding right now, but only once it is merged to `next` and then into the many feature branches.


app/frontend/js/upload.coffee
```diff
@@ -710,11 +601,8 @@ window.userManuscriptUpload = (fileInput) ->
     if this.files.length > 1
       $('#userManuscript-status').show(400)
       $('#multiple-files-selected').show()
-      $('#files-merge').show()
-      $('#file-size-correct').hide()
-      $('#file-size-way-too-big').hide()
-      $('#file-size-too-big').hide()
-      $('#file-optimize').hide()
+      $('#file-permission-field').hide()
+      $('#submission-final-upload-dialogue').hide()
```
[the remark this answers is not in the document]
Splines: good point, fixed in 2f17497e036

.devcontainer/devcontainer.json
```diff
@@ -97,7 +97,6 @@
             "onAutoForward": "openBrowser"
         }
     },
-    "initializeCommand": "git submodule update --init --recursive",
```
Splines, line 100: this might be problematic

fosterfarrell9 (approved):
I built this locally, the build went through and everything works. I played around in the GUI and uploaded submission pdfs with smaller/larger size than the 20 MB boundary and everything worked as expected.

.devcontainer/devcontainer.json
Splines: not a problem apparently

PR #1132, opened 2026-05-18
Add Brakeman to CI/CD as static code analysis tool

Splines:
We add [Brakeman](https://brakemanscanner.org/) to our GitHub Actions. Brakeman is a free vulnerability scanner to statically analyze Rails application code.

Note that I set up brakeman in the `tests.yml` workflow to avoid having to approve the runs twice (that would be necessary if I create another workflow files just for Brakeman).


fosterfarrell9 (commented):
LGTM!

fosterfarrell9 (approved):

PR #1139, opened 2026-05-23
Harden pdfupload

fosterfarrell9:
This PR improves the robustness of PDF uploads by hardening the pdftk-based metadata extraction path. It removes shell invocation, adds execution and size bounds, and only accepts the expected structure.mampf attachment during unpacking.


fosterfarrell9:
Ah, this got stacked on top of the password policy PR.  I will close it in favor of a clean slate starting form main.

PR #1146, opened 2026-05-25
Sanitize submission response filenames for manuscripts and corrections

fosterfarrell9:
**Note:** This PR has been superseded by #1151.

In writing #1144 I realized that our submission downloads deserve a hardening, that's why I wrote this fix. Note that this is not a path traversal problem we fix here: the file path served by Rails already comes from Shrine-managed storage, not from the uploaded filename. Still, uploaded metadata was flowing into the response filename, so we now strip path segments and control characters before handing it to Rails. Adds request coverage for hostile metadata on both endpoints.


fosterfarrell9:
Closed in favor of #1151

PR #1149, opened 2026-05-27
Integrate uppy for uploads

fosterfarrell9:
In the PR we remove the infamous `upload.coffee` file and replace it by the use of the uppy library (which is the library recommended by shrine). Note that the R is much smaller than LOC indicate because I built it on top of #1139 and #1143 which harden the file uploaders on the server side and are not yet in `next`.


fosterfarrell9:
Closed in favor of #1195.

PR #1147, opened 2026-05-25
Set up Thruster integration and update Docker configurations

fosterfarrell9:
**Note:** This PR has been superseded by #1151 

In this PR, we set up thruster and remove nginx. As a consequence, we also remove the master/worker load balancing. This is working for me in development, for production there is most probably still some work to do.

PR #1151, opened 2026-06-17
💎 Harden the app and modernize the runtime

fosterfarrell9:
This is the app-repo side of moving MaMpf onto its new deployment setup (the companion changes live in `mampf-infra` - the changes in this PR only make sense in together with `mampf-infra`). Two main thrusts:

## Security hardening
- **Uploads:** per-uploader allow-lists (extensions + MIME via marcel), size limits, auth required on upload routes; dropped the unused zip uploader.
- **New upload post-processing pipeline** — replaces the riskiest legacy tooling: **pdftk → qpdf** and **ImageMagick/MiniMagick → libvips**. This is what supersedes parts of the PRs listed below.
- **Malware scanning (ClamAV)** via `MalwareScanGate`: bounded-prefix scan for large videos, scan metrics, log-injection-safe logging.
- **Upload authorization:** internal authz endpoint + nginx `auth_request` gate; ActiveStorage direct-upload gate.
- **App-level:** CSV formula-injection neutralization (`CsvSafe`), removal of the dead `secrets.yml` + `.env` hygiene, baseline **CSP** (report-only for now).

## Runtime / base-image modernization
- **Debian 13 (trixie-slim)**, **Ruby 3.4.10**, libvips, OR-Tools, Node/yarn v4, Postgres/Redis bumps, `/up` healthcheck, `strong_migrations`.
- **Slimmer, hardened images:** new production `docker-compose` + **multistage Dockerfile** — build toolchain stripped from the final image, slim base ⇒ smaller images and a reduced attack surface.

## Supporting changes
- Media served **through Rails** instead of the external media server (stream_video, VTT, screenshots, images, inline manuscript/geogebra; removed the media-server config + Apache vhost).
- Signed + scanned image CI/CD (cosign, Trivy, ghcr `mampf-deploy`).
- Thruster app server (gem + cleanup); devcontainer tooling (SOPS, age, Ruby LSP).

## Included / superseded PRs
This branch bundles the PRs below (which inflates the diff/LOC). It also **supersedes several of them**, because the new qpdf/libvips upload-postprocessing pipeline replaces their pdftk/ImageMagick-based approach:

- #1134
- #1140
- #1143
- #1144
- #1147
- #1148


app/uploaders/correction_uploader.rb
```diff
@@ -1,16 +1,79 @@
-require "image_processing/mini_magick"
-
 # UserPdfUploader Class
 class CorrectionUploader < Shrine
+  MAX_SIZE = 30 * 1024 * 1024
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [12cba44](https://github.com/MaMpf-HD/mampf/pull/1151/commits/12cba44cb35aa22b01673a2eb9ecf3932b2ae52a).

app/models/submission.rb
```diff
@@ -167,10 +167,20 @@ def check_file_properties_any(metadata, sort)
...
+                           CorrectionUploader.accepted_extension_list))
+    end
+    if errors.blank? &&
+       !CorrectionUploader.allowed_mime_type?(filename: file_name,
+                                              mime_type: metadata["mime_type"])
+      errors.push(I18n.t("submission.wrong_mime_type",
+                         mime_type: metadata["mime_type"],
+                         accepted_mime_types:
+                           CorrectionUploader.accepted_mime_types_for(file_name)
+                                             .join(", ")))
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [12cba44](https://github.com/MaMpf-HD/mampf/pull/1151/commits/12cba44cb35aa22b01673a2eb9ecf3932b2ae52a).

app/services/upload_endpoint_authorization.rb
```diff
@@ -0,0 +1,50 @@
...
+        # upload intent: a historic editor of an existing non-talk medium
+        # must still be able to replace the video on that medium.
+        content_editor?(user) || user&.speaker?
+      when "PdfUploader", "GeogebraUploader", "ScreenshotUploader"
+        content_editor?(user)
+      when "CorrectionUploader"
+        content_editor?(user) || user&.tutor?
+      else
+        true
+      end
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [12cba44](https://github.com/MaMpf-HD/mampf/pull/1151/commits/12cba44cb35aa22b01673a2eb9ecf3932b2ae52a).

app/services/csv_safe.rb
```diff
@@ -0,0 +1,30 @@
...
+  FORMULA_TRIGGERS = ["=", "+", "-", "@", "\t", "\r"].freeze
+
+  class << self
+    # Sanitizes a single cell value. Non-string values (numbers, dates, nil)
+    # cannot carry a formula payload and are returned unchanged.
+    def cell(value)
+      return value unless value.is_a?(String)
+      return value unless value.start_with?(*FORMULA_TRIGGERS)
+
+      "'#{value}"
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [12cba44](https://github.com/MaMpf-HD/mampf/pull/1151/commits/12cba44cb35aa22b01673a2eb9ecf3932b2ae52a).

app/abilities/user_ability.rb
```diff
@@ -4,7 +4,7 @@ class UserAbility
   def initialize(user)
     clear_aliased_actions
 
-    can [:delete_account, :teacher], User
+    can [:delete_account, :teacher, :image], User
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [9a27555](https://github.com/MaMpf-HD/mampf/pull/1151/commits/9a27555ee6c6ad6871616d76834057ede0e0e304).

.yarnrc.yml
```diff
@@ -0,0 +1,5 @@
+enableScripts: true
+
+nodeLinker: node-modules
+
+npmMinimalAgeGate: 0
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [5b5855a](https://github.com/MaMpf-HD/mampf/pull/1151/commits/5b5855a26b7fb547ff6fb82f93ccaaa14e8d44aa).

app/abilities/user_ability.rb
```diff
@@ -6,6 +6,14 @@ def initialize(user)
 
     can [:delete_account, :teacher], User
 
+    # Profile images are a teacher-profile feature: teacher images are shown to
+    # everyone on the teacher info page, but a user's image must not be
+    # enumerable across accounts. (Upload is gated separately in
+    # UploadEndpointAuthorization.)
+    can :image, User do |given_user|
+      user.admin? || user == given_user || given_user.teacher?
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Addressed — `:image` now uses `user&.admin?` (safe navigation), so the public teacher-image case no longer raises for an anonymous (nil) user. Added a `UserAbility.new(nil)` spec to lock the contract (teacher allowed, non-teacher denied). Fixed in [d564e7c](https://github.com/MaMpf-HD/mampf/pull/1151/commits/d564e7cde72cb158d20d4df79382d3c169d9cfe7).

app/services/upload_endpoint_authorization.rb
```diff
@@ -0,0 +1,55 @@
...
+        content_editor?(user) || user&.speaker?
+      when "PdfUploader", "GeogebraUploader", "ScreenshotUploader"
+        content_editor?(user)
+      when "CorrectionUploader"
+        content_editor?(user) || user&.tutor?
+      when "SubmissionUploader", "ProfileimageUploader"
+        # Intentionally open to any authenticated user; these endpoints sit
+        # behind `authenticate :user`. Listed explicitly so the gate fails
+        # closed for any future uploader not covered above.
+        true
```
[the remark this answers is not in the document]
fosterfarrell9: Addressed — `SubmissionUploader` now returns `user.present?` instead of `true`, so a nil caller fails closed. The `authenticate :user` route filter stays as defence in depth.

app/controllers/users_controller.rb
```diff
@@ -92,6 +104,24 @@ def delete_account
...
   private
 
+    # Only the owner and admins may fetch the original upload (full resolution
+    # and EXIF). Everyone else — e.g. a student viewing a teacher profile — is
+    # served the resized, metadata-stripped derivative.
+    def image_variant_for(user)
+      return params[:variant] if current_user.admin? || current_user == user
+
+      "normalized"
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Addressed — `image_variant_for` now uses `current_user&.admin?` (safe navigation), so anonymous viewers no longer raise after the public teacher-image authorization.

spec/abilities/user_ability_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:user) { create(:confirmed_user) }
+  let(:teacher) do
+    create(:confirmed_user).tap { |u| create(:lecture, teacher: u) }
+  end
+  let(:other) { create(:confirmed_user) }
+
+  describe ":image" do
+    it "lets anyone view a teacher's profile image" do
+      expect(ability.can?(:image, teacher)).to be(true)
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Done — added a `described_class.new(nil)` example asserting the public teacher-image contract (teacher allowed, non-teacher denied) without crashing. Fixed in [d564e7c](https://github.com/MaMpf-HD/mampf/pull/1151/commits/d564e7cde72cb158d20d4df79382d3c169d9cfe7)

docker/production/docker-compose.production.yml
```diff
@@ -13,65 +13,33 @@ x-mampf:
...
+    image: redis:8.8-alpine
     networks:
       - mampf
 
   cache:
     restart: always
-    image: memcached:alpine
+    image: memcached:1.6-alpine
     networks:
       - default
```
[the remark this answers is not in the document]
fosterfarrell9: False positive for the actual deployment. This in-repo compose file is **not** the production deployment — the live stack is rendered by our `mampf-infra` (infrastructure-as-code) repo, where `app`, `sidekiq`, `redis`, `cache` and `clamav` all share one internal network, so `cache` resolves fine. This file is an unused standalone example (placeholders like `yourfileserverdomain`); I have added an "EXAMPLE ONLY" header to make that explicit.

docker/production/docker.env
```diff
@@ -1,11 +1,24 @@
+# =============================================================================
+# BUILD-TIME DUMMY ENV FILE — NOT USED AT RUNTIME.
+#
+# None of the values below are real secrets, and none of them reach a running
+# production/staging container. The actual runtime configuration is injected by
+# the deployment automation, which requires every secret to be set explicitly
+# (a deploy aborts otherwise).
```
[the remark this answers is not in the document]
fosterfarrell9: False positive for the actual deployment. In production the runtime env is provided by vaulted variables in our `mampf-infra` repo, not by `docker.env` — which is a build-time dummy only (asset precompile), as documented there. The `env_file: docker.env` line lives in the unused example compose, now marked "EXAMPLE ONLY".

docker/production/Dockerfile
```diff
@@ -107,4 +96,58 @@ RUN cp -r $(bundle info --path sidekiq)/web/assets /usr/src/app/public/sidekiq &
...
+            /usr/local/bin/yarn /usr/local/bin/yarnpkg \
+            /usr/src/app/bin/yarn && \
+        rm -rf /usr/local/lib/node_modules/npm \
+            /usr/local/lib/node_modules/corepack && \
+    rm -rf /usr/local/bundle/cache
+
+USER app
+
+HEALTHCHECK --interval=30s --timeout=10s --start-period=20s --retries=3 \
+    CMD ruby -rnet/http -e 'port = ENV.fetch("APP_HEALTHCHECK_PORT", "3000"); uri = URI("http://127.0.0.1:#{port}/up"); response = Net::HTTP.start(uri.host, uri.port) { |http| http.get(uri.request_uri, {"X-Forwarded-Proto" => "https", "Host" => "localhost"}) }; exit(response.is_a?(Net::HTTPSuccess) ? 0 : 1)'
```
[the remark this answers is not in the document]
fosterfarrell9: Not an issue in the actual deployment: the live `mampf-infra` compose gives the `sidekiq` service its own healthcheck, which overrides this image-level HEALTHCHECK; only the HTTP `app` service inherits the `/up` check. So Sidekiq is never marked unhealthy. Keeping the image-level `/up` check since `app` is the primary consumer of this shared image.

app/controllers/users_controller.rb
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [d564e7c](https://github.com/MaMpf-HD/mampf/pull/1151/commits/d564e7cde72cb158d20d4df79382d3c169d9cfe7).

.devcontainer/boot.sh
```diff
@@ -10,8 +32,23 @@ echo "🚀 Install zsh"
 echo "🚀 Install MdBook"
 ./.devcontainer/install-mdbook.sh
 
+if [ -d /workspaces/mampf-infra ]; then
+  echo "🚀 Install sops and age for mampf-infra"
+  sudo apt-get update && sudo apt-get install -y age
+  wget -O /tmp/sops https://github.com/getsops/sops/releases/download/v3.13.1/sops-v3.13.1.linux.amd64
+  chmod +x /tmp/sops
+  sudo mv /tmp/sops /usr/local/bin/sops
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/368bb23670bedf78a34f1a956ca4c2bd7710d45d — `boot.sh` now downloads `sops` into a `mktemp` directory and verifies it against the upstream-published `sops-v3.13.1.checksums.txt` (`sha256sum -c`) before installing.

.devcontainer/install-ansible.sh
```diff
@@ -0,0 +1,36 @@
...
+uv_bin="$HOME/.local/bin/uv"
+
+sudo apt-get update
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y ssh-askpass
+
+ensure_uv() {
+  if [ ! -x "$uv_bin" ]; then
+    curl -LsSf https://astral.sh/uv/install.sh | sh
+  fi
+}
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/368bb23670bedf78a34f1a956ca4c2bd7710d45d — replaced the `curl … | sh` install with a pinned uv release (`0.11.25`): download the tarball plus its published `.sha256`, verify with `sha256sum -c`, then install the binary. No more piping a remote script to a shell.

bin/yarn
```diff
@@ -6,12 +6,20 @@ Dir.chdir(APP_ROOT) do
...
+  corepack = ENV["PATH"].split(File::PATH_SEPARATOR).
+    select { |dir| File.expand_path(dir) != __dir__ }.
+    product(["corepack", "corepack.cmd", "corepack.ps1"]).
+    map { |dir, file| File.expand_path(file, dir) }.
+    find { |file| File.executable?(file) }
 
   if yarn
     exec yarn, *ARGV
+  elsif corepack
+    exec corepack, "yarn", *ARGV
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/368bb23670bedf78a34f1a956ca4c2bd7710d45d — `bin/yarn` now prefers Corepack (which honours the repo-pinned `packageManager`) and only falls back to a direct `yarn` on PATH when Corepack is unavailable.

app/services/malware_scan_gate.rb
```diff
@@ -0,0 +1,217 @@
...
+      end
+
+      def merge_metadata(metadata)
+        merged_metadata = metadata.is_a?(Hash) ? metadata.deep_dup : {}
+        merged_metadata[METADATA_KEY] = {
+          STATUS_KEY => CLEAN_STATUS,
+          "scanner" => SCANNER_NAME,
+          "scanned_at" => Time.current.utc.iso8601
+        }
+        merged_metadata
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/8c1518aefee2103af89e9588898f48cd559db0e8 — the scan `scope` (`prefix`/`full`) is now stored in the signed `malware_scan` metadata and included in the HMAC signature, so a bounded-prefix verdict is distinguishable from a full-file one and tamper-proof. `clean_scanned?` / `validate_cached_file!` / `ensure_promotable!` additionally verify the scope matches the uploader's expectation (`expected_scope_for`), so a prefix-scanned blob can't be accepted where a full-file scan is required.

app/uploaders/screenshot_uploader.rb
```diff
@@ -1,23 +1,36 @@
...
   plugin :pretty_location
   plugin :derivatives
 
+  Attacher.prepend(MalwareScannableAttacher)
+
   Attacher.validate do
+    MalwareScanGate.validate_cached_file!(self)
+
     validate_mime_type_inclusion ["image/jpeg", "image/png", "image/gif"],
                                  message: "falscher MIME-Typ"
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/abcf509e709136ec4e4b92f305f16df0918da3fc — the hardcoded `"falscher MIME-Typ"` is replaced with the existing i18n key `submission.wrong_mime_type`, interpolating the detected MIME type (`file&.mime_type`) and the accepted types (`ACCEPTED_MIME_TYPES.join(", ")`). The error is now localized (de/en) and informative. Applied consistently across all four `validate_mime_type_inclusion` uploaders (screenshot, profileimage, video, pdf — pdf carried the same hardcoded message but wasn't flagged here).

app/uploaders/profileimage_uploader.rb
```diff
@@ -1,23 +1,36 @@
...
   plugin :pretty_location
   plugin :derivatives
 
+  Attacher.prepend(MalwareScannableAttacher)
+
   Attacher.validate do
+    MalwareScanGate.validate_cached_file!(self)
+
     validate_mime_type_inclusion ["image/jpeg", "image/png", "image/gif"],
                                  message: "falscher MIME-Typ"
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/abcf509e709136ec4e4b92f305f16df0918da3fc — the hardcoded `"falscher MIME-Typ"` is replaced with the existing i18n key `submission.wrong_mime_type`, interpolating the detected MIME type (`file&.mime_type`) and the accepted types (`ACCEPTED_MIME_TYPES.join(", ")`). The error is now localized (de/en) and informative. Applied consistently across all four `validate_mime_type_inclusion` uploaders (screenshot, profileimage, video, pdf — pdf carried the same hardcoded message but wasn't flagged here).

app/uploaders/video_uploader.rb
```diff
@@ -2,17 +2,34 @@
 
 # VideoUploader class
 class VideoUploader < Shrine
+  MAX_SIZE = 4 * 1024 * 1024 * 1024
+  # Upper bound (bytes) on what is streamed to clamd. Video is scanned over a
+  # bounded prefix only: this is defense-in-depth against known malware near the
+  # container start, NOT a full-file clean verdict (see MalwareScanGate and
+  # docs/CLAMAV_UPLOAD_SCANNING.md in mampf-infra).
+  SCAN_MAX_BYTES = 32 * 1024 * 1024
+  WRONG_TYPE_MESSAGE = "wrong type".freeze
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in https://github.com/MaMpf-HD/mampf/commit/abcf509e709136ec4e4b92f305f16df0918da3fc — the hardcoded `WRONG_TYPE_MESSAGE = "wrong type"` constant is removed; the validation now uses the existing i18n key `submission.wrong_mime_type`, interpolating the detected MIME type (`file&.mime_type`) and the accepted types (`ACCEPTED_MIME_TYPES.join(", ")`), so the error is localized (de/en) and informative. The same change was applied to the other three `validate_mime_type_inclusion` uploaders (screenshot, profileimage, pdf) for consistency.

.github/workflows/linter.yml
```diff
@@ -57,11 +57,29 @@ jobs:
...
+          if: ${{ steps.js-changed.outputs.changed-files != ''}}
+          uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
+          with:
+            path: |
+              ~/.yarn/berry/cache
+              .yarn/install-state.gz
+              node_modules
+            key: ${{ runner.os }}-node24-yarn4-${{ hashFiles('package.json', 'yarn.lock', '.yarnrc.yml') }}
+            restore-keys: |
+              ${{ runner.os }}-node24-yarn4-
```
Splines, line 83: Why did you disable the yarn cache and instead implement a more "custom" solution via Corepack? What was the problem with the yarn cache of setup-node?

.github/workflows/schedule-deploy-image-rebuild.yml
```diff
@@ -0,0 +1,32 @@
+name: Schedule Deploy Image Rebuild
+
+on:
+  workflow_dispatch:
+    inputs:
+      build_ref:
+        description: Branch to rebuild
+        required: false
+        default: deploy/app-baseline-thruster
```
Splines, line 14: This should probably be changed, right?

.github/workflows/schedule-deploy-image-rebuild.yml
```diff
@@ -0,0 +1,32 @@
...
+        type: string
+  schedule:
+    - cron: "17 3 * * 1"
```
Splines, line 17: Why do we schedule an automatic rebuild here? And why this oddly specific time?

.github/workflows/schedule-deploy-image-rebuild.yml
```diff
@@ -0,0 +1,32 @@
...
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Dispatch build-deploy-image workflow on target branch
+        run: |
+          curl --fail --silent --show-error \
+            -X POST \
+            -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/actions/workflows/build-deploy-image.yml/dispatches" \
+            -d "{\"ref\":\"${BUILD_REF}\"}"
```
Splines: I'd prefer a solution using the GitHub CLI here as this would probably be more stable. E.g.

```
# The -f flag sends string values, so nested objects must use the
# key[subkey]=value syntax (or pass a full JSON body via --input).
gh api repos/owner/repo/dispatches \
  -f event_type=deploy \
  -f 'client_payload[environment]=staging' \
  -f 'client_payload[version]=1.2.3'
```

[from here](https://oneuptime.com/blog/post/2025-12-20-repository-dispatch-github-actions/view). See also the [GitHub Docs](https://cli.github.com/manual/gh_api).

docker/production/Dockerfile
```diff
@@ -107,4 +96,63 @@ RUN cp -r $(bundle info --path sidekiq)/web/assets /usr/src/app/public/sidekiq &
...
+    bundle clean --force && \
+    rm -rf /usr/local/bundle/bundler/gems/*/.git && \
+    rm -rf /usr/local/bundle/cache /usr/local/bundle/ruby/*/cache && \
+    find /usr/local/bundle -name "*.gem" -delete
+
+FROM ruby:3.4.9-slim-trixie
+
+ENV RAILS_ENV=production \
+    BUNDLE_DEPLOYMENT=1 \
+    BUNDLE_WITHOUT="development:test" \
```
Splines: Some of these env variables repeat in this file. Can you put the repeating variables in the global scope and use them in the different build stages? [The docs on Scoping](https://docs.docker.com/build/building/variables/#scoping) explain this pretty well.

docker/production/Dockerfile
```diff
@@ -107,4 +96,63 @@ RUN cp -r $(bundle info --path sidekiq)/web/assets /usr/src/app/public/sidekiq &
...
+    BUNDLE_PATH=/usr/local/bundle \
+    APP_HEALTHCHECK_PORT=3000 \
+    APP_CACHE_ROOT=/caches
+
+ARG BUNDLER_VERSION=4.0.12
+
+EXPOSE 3000
+EXPOSE 9394
```
Splines: We don't have `EXPOSE` lines in the development setup neither. And I don't see the advantage of declaring those ports here. This is really just a documentation, the statement itself [does nothing](https://docs.docker.com/reference/dockerfile/#expose).

> The EXPOSE instruction doesn't actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published. 

I think we should rather "document" this as actual code where the ports are specified (docker compose and our Ansible Infrastructure repo).

Splines (changes requested):
Nice, this in conjunction with the new private infrastructure repo (Ansible) is such a major step in MaMpf's maturity and security! Awesome work, here just some more comments.

docker/production/Dockerfile
```diff
@@ -107,4 +96,63 @@ RUN cp -r $(bundle info --path sidekiq)/web/assets /usr/src/app/public/sidekiq &
...
+
+ENTRYPOINT ["/usr/src/app/docker/production/entrypoint-app.sh"]
+
+ARG OS_SECURITY_REFRESH=
+# Empty in normal (reproducible) builds; `just patch-image` passes a unique value
+# to bust this layer and apply all available Debian security updates on top of the
+# (force-pulled) base image.
```
Splines: Put the comment above the variable, not below

docker/production/Dockerfile
```diff
@@ -1,77 +1,66 @@
...
-RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
-ENV NVM_DIR=/root/.nvm
-RUN source "${NVM_DIR}/nvm.sh" && nvm install "${NODE_VERSION}" && \
-    nvm use "${NODE_VERSION}" && nvm alias default "${NODE_VERSION}"
-
-ENV NODE_PATH="${NVM_DIR}/versions/node/v${NODE_VERSION}/bin/"
-ENV PATH="${NODE_PATH}:${PATH}"
-
-RUN nvm current; node --version
+COPY --from=node /usr/local/ /usr/local/
```
Splines, line 28: Nice, that's very clever to "install" Node.js by just copying over the binaries from the already-built Docker image.

.github/workflows/build-deploy-image.yml
```diff
@@ -0,0 +1,163 @@
...
+        required: false
+        type: string
+      os_security_refresh:
+        description: Force-pull base and apply all available OS security updates (non-reproducible)
+        required: false
+        type: boolean
+        default: false
+  push:
+    branches:
+      - deploy/app-baseline-thruster
```
Splines, line 26: deploy/app-baseline-thruster should probably be replaced by something like "main", see also other occurrences in the codebase

.gitignore
```diff
@@ -24,6 +24,12 @@ app/frontend/js/mampf_routes.js
 
 /node_modules
 /yarn-error.log
+/.yarn/*
+!/.yarn/patches
+!/.yarn/plugins
+!/.yarn/releases
+!/.yarn/sdks
+!/.yarn/versions
```
Splines, line 32: why do we need this exclusion list from the .yarn folder? e.g. why do we want to git-commit .yarn/sdks?

But fine with me if we keep it

.yarnrc.yml
```diff
@@ -0,0 +1,7 @@
+approvedGitRepositories: []
+
+enableScripts: true
+
+nodeLinker: node-modules
+
+npmMinimalAgeGate: "1w"
```
Splines, line 7: That's really nice

docker/development/init.sh
```diff
@@ -1,31 +1,32 @@
...
   # Database preseed
-  if [[ "${DB_SQL_PRESEED_URL}" ]]; then
+  if [[ -n "${DB_SQL_PRESEED_URL:-}" ]]; then
     if [[ -f "${DB_SQL_PRESEED_URL}" ]]; then
       echo "💾  Found DB preseed file: $DB_SQL_PRESEED_URL"
       latest=$DB_SQL_PRESEED_URL
     else
       echo "💾  Found DB preseed at URL: $DB_SQL_PRESEED_URL"
       mkdir -pv db/backups/development
       wget --content-disposition --directory-prefix=db/backups/development/ --timestamping $DB_SQL_PRESEED_URL
```
Splines: Could you install the extension "ShellCheck" in VSCode? It highlights "Double quote to prevent globbing and word splitting." here.

docker/development/init.sh
```diff
@@ -1,31 +1,32 @@
...
-set -e
+set -Eeuo pipefail
 cd /workspaces/mampf/
 
 check_for_preseeds() {
   echo "💾  Checking for preseeds (in development env)"
 
   # Database preseed
-  if [[ "${DB_SQL_PRESEED_URL}" ]]; then
+  if [[ -n "${DB_SQL_PRESEED_URL:-}" ]]; then
```
Splines, line 9: Beforehand, this was without the fallback and without the "-n". Just for my understanding, in which cases was that not enough?

.devcontainer/install-ansible.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -e
+
+infra_root=/workspaces/mampf-infra
+uv_bin="$HOME/.local/bin/uv"
+uv_version="0.11.25"
```
Splines: The newest release is 0.11.26

app/controllers/media_controller.rb
```diff
@@ -1,6 +1,12 @@
 # MediaController
 class MediaController < ApplicationController
-  skip_before_action :authenticate_user!, only: [:play, :display]
+  skip_before_action :authenticate_user!, only: [:play, :screenshot,
+                                                 :chapters_vtt,
+                                                 :references_vtt, :display,
+                                                 :stream_video,
+                                                 :inline_manuscript,
+                                                 :geogebra, :inline_geogebra,
+                                                 :download]
```
Splines, line 9: Why do we skip authentication for these methods?

config/environments/development.rb
```diff
@@ -81,6 +81,9 @@
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
 
+  config.action_dispatch.x_sendfile_header =
+    ENV.fetch("ACTION_DISPATCH_X_SENDFILE_HEADER", nil)
```
Splines, line 85: Don't we rather want to throw if this key is not set? I.e use ENV.fetch("...") without the `nil` fallback. This would make it evident that we might have forgotten to set this key somewhere instead of relying (unknowlingy) on a fallback value.

Same for other Env variable changes throughout this PR.

docker/development/.env
```diff
@@ -1,10 +1,18 @@
...
+# (the vignette images) that are HMAC-signed from this key. A different value makes
+# those images silently disappear on render. To rotate it, the seed data in the
+# mampf-init-data repo must be regenerated/re-signed with the new key.
 SECRET_KEY_BASE=myNotSoSecretKeybaseForMaMpfDevelopment
 INSTANCE_PATH=mampf
 VITE_RUBY_HOST=0.0.0.0
 REDIS_URL=redis://redis:6379/1
 SPROCKETS_CACHE=/cache
+CLAMAV_HOST=clamav
+CLAMAV_PORT=3310
```
Splines, line 14: Do we somehow have to make this port available in VSCode as well? If yes, add it to the devcontainer.json

db/migrate/20260718000000_drop_vtt_containers.rb
Splines, line 1: Update date to newer date & regenerate schema

Splines: Could you maybe quickly summarize to me what these vtt_containers were and why we don't need them anymore?

.devcontainer/devcontainer.json
```diff
@@ -100,6 +95,9 @@
     "postCreateCommand": ".devcontainer/boot.sh",
     "customizations": {
         "vscode": {
+            "settings": {
+                "rubyLsp.customRubyCommand": "export PATH=/usr/local/bin:$PATH"
```
Splines: add a link to the issue on RubyLSP as comment

.devcontainer/devcontainer.json
```diff
@@ -19,13 +9,18 @@
     ],
     "service": "app",
     // https://github.com/orgs/community/discussions/15506
+    "mounts": [
+    "source=${localWorkspaceFolder}/../mampf-infra,target=/workspaces/mampf-infra,type=bind"
```
Splines: Will this still work if one hasn't cloned the mampf-infra repo yet? HiWis might have to work in MaMpf but never in mampf-infra, so I think this should only be optional and fallback gracely.

app/uploaders/pdf_uploader.rb
```diff
@@ -1,82 +1,167 @@
-require "image_processing/mini_magick"
+require "image_processing/vips"
```
Splines, line 1: According to the [Readme](https://github.com/janko/image_processing#installation) of `image_processing`, one should also include explicitly the `ruby-vips` dependency in the Gemfile.

Also note that we are still on image processing v1, there is v2 out there already, so you may want to do a major upgrade.

.devcontainer/boot.sh
```diff
@@ -1,6 +1,28 @@
 #!/bin/bash
 set -e
 
+OR_TOOLS_URL="https://github.com/MaMpf-HD/build-artifacts/releases/download/or-tools-9.15-trixie/or-tools-9.15-trixie-amd64.tar.gz"
+OR_TOOLS_SHA256="067e96ba57bbca8bf8e56c9b85010e223d49696cb5fe1a56ef94f5bd3647c0cf"
+
+echo "🚀 Fix bundle cache ownership"
+sudo mkdir -p /usr/local/bundle
+sudo chown -R "$(id -u):$(id -g)" /usr/local/bundle
```
Splines, line 9: Is this a problem you encountered? Could you add a small comment, maybe also link to a Bundle issue on GitHub.

.devcontainer/boot.sh
```diff
@@ -1,6 +1,28 @@
 #!/bin/bash
 set -e
 
+OR_TOOLS_URL="https://github.com/MaMpf-HD/build-artifacts/releases/download/or-tools-9.15-trixie/or-tools-9.15-trixie-amd64.tar.gz"
```
Splines, line 4: The build-artifacts repo only includes a Readme.md. How is this binary built?

docker/development/compose.yml
```diff
@@ -26,12 +32,22 @@ services:
     environment:
       - RAILS_ENV=development
       - NODE_ENV=development
-      - MAMPF_PORT=3000
+      - MAMPF_PORT=3001
```
Splines, line 35: The `MAMPF_PORT` was changed here, but not in the `devcontainer.json` (forwarded ports), is that intentional?

Splines (changes requested):

docker/development/init.sh
Splines: I just verified and ShellCheck is already installed in our DevContainer, so you should see that warning as well.

app/services/csv_safe.rb
```diff
@@ -0,0 +1,30 @@
...
+# untouched everywhere else in the application.
+class CsvSafe
+  FORMULA_TRIGGERS = ["=", "+", "-", "@", "\t", "\r", "\n"].freeze
+
+  class << self
+    # Sanitizes a single cell value. Non-string values (numbers, dates, nil)
+    # cannot carry a formula payload and are returned unchanged.
+    def cell(value)
+      return value unless value.is_a?(String)
+      return value unless value.start_with?(*FORMULA_TRIGGERS)
```
Splines: What if an attacker just prefixes with a whitespace? Shouldn't we check for the first _non-whitespace_ character instead?

app/services/csv_safe.rb
Splines, line 1: When using this class, one has to sanitize each cell manually by calling `CsvSafe.cell(...)` (e.g. see tutorial.rb). What about an API instead where we wrap Ruby CSV in a way like this:
```rb
SafeCsv.generate(col_sep: ";") do |csv|
  csv << ["=evil()", "ok", 42]
end
```
Here, `csv` would be a wrapper of Ruby's raw CSV object that always applies `CsvSafe.row`, such that every written row is always sanitized.

And then to ensure we also adopt this pattern in the codebase, add a custom RuboCop rule that forbits direct calls of `CSV.generate` or `CSV.open` (with a hint to use `SafeCsv.generate` instead).

Splines (dismissed):

.github/workflows/build-deploy-image.yml
fosterfarrell9: Added next and main as triggers in [8115ccf](https://github.com/MaMpf-HD/mampf/pull/1151/commits/8115ccf487c6cad3ebeb2e197b7c319cdcc983d7). Topology is `next → staging`, `main → production`. A `push.branches` entry fires from the workflow version on the branch pushed to, so next/main stay dormant until the baseline (including this workflow) actually reaches them — nothing builds from a branch that lacks the code, so it's safe to list both ahead of cutover. thruster drops once retired.

Note the full switch is cross-repo — in the `mampf-infra` repo `just deploy latest` hard-codes this branch too, and the per-env `mampf_app_baseline_branch` inventory var maps `branch → env`.

.github/workflows/schedule-deploy-image-rebuild.yml
fosterfarrell9: Left on `deploy/app-baseline-thruster` for now, with an inline note on why in [8115ccf](https://github.com/MaMpf-HD/mampf/pull/1151/commits/8115ccf487c6cad3ebeb2e197b7c319cdcc983d7): this is the weekly OS-patch rebuild and it fires on a schedule (not on code arrival), so pre-adding `next` would rebuild a branch that doesn't have the baseline yet. At cutover it becomes a matrix over both live baselines (`next = staging`, `main = production`) instead of a single ref. (The `BUILD_REF` fallback a few lines down moves together with this default.)

.github/workflows/schedule-deploy-image-rebuild.yml
fosterfarrell9: It's a weekly rebuild of the deploy image so its base/OS layers stay current on security updates instead of drifting as the image ages — it builds + signs a fresh image but does not deploy (promotion stays a manual `just deploy`). It complements the on-demand `just app-image patch` and the Trivy scan that runs on every build. (To make the weekly run genuinely re-pull the base instead of reusing cache, I'll have it force an OS refresh — small follow-up.)

The time isn't arbitrary: `17 3 * * 1` = Mondays 03:17 UTC. GitHub delays scheduled runs during the load spike at the top of every hour and recommends not using round times, so it's offset to :17; 03:xx UTC is low-traffic (night in EU/US), and weekly-on-Monday means a freshly-patched image is ready for the week.

db/migrate/20260718000000_drop_vtt_containers.rb
fosterfarrell9: Done in [03e4060](https://github.com/MaMpf-HD/mampf/pull/1151/commits/03e40604c01befc118697201e76a42dba3a0c9e7).

``vtt_containers` (added back in 2020) cached the two WebVTT tracks we attach to a video — the table-of-contents / chapters track (`table_of_contents_data`) and the cross-references track (`references_data`) — as stored text blobs.
We don't store them anymore: both tracks are now built on the fly from the canonical data (the medium's TOC items and its referrals) via Medium#toc_vtt_content / references_vtt_content, served by the chapters_vtt / references_vtt endpoints. That's strictly better — no stale cache, and the references track is scoped per user (referral_visible_for_vtt? hides references to media the viewer can't access), which a single stored blob couldn't do. The VttContainer model + wiring were already removed in 53cfa7e2; this migration just drops the now-orphaned table.

app/uploaders/pdf_uploader.rb
fosterfarrell9: The problem with image processing v2 is that that there are some problems with Rails Acive Storag for Rails 8.0 (see https://github.com/rails/rails/pull/57403). I would wait with image processing v2 until there is  Rails version out that natively supports it. And in v1, ruby-vips is a runtime dependency of image processing, so no need to take it in now.

.devcontainer/devcontainer.json
fosterfarrell9: Fixed in [ff6b62c](https://github.com/MaMpf-HD/mampf/pull/1151/commits/ff6b62c43b8427ff1fc82a31dc8abf8d16f5ea2f).

config/environments/development.rb
fosterfarrell9: Went through the env vars in this PR case-by-case. The required ones already fail loud — no default, so a missing key raises: URL_HOST, MAILSERVER, MAMPF_EMAIL_USERNAME/PASSWORD, MEMCACHED_SERVER, MAMPF_PORT. The ones with defaults are intentionally optional (CLAMAV_*, MAILSERVER_PORT 587, X-Accel-Redirect, CACHE_MAX_AGE_HOURS, CSP_REPORT_ONLY).

This specific line — ACTION_DISPATCH_X_SENDFILE_HEADER in development — is deliberately optional: x-sendfile is off in dev (Rails serves files itself), so it should be nil when unset, not raise. And ENV.fetch("X", nil) is actually the form our Style/FetchEnvVar RuboCop rule enforces (a bare ENV["X"] gets flagged), so I've left it as-is. Fail-fast intent is already covered for the keys that matter.

app/services/csv_safe.rb
fosterfarrell9: Also done in [fce64b6](https://github.com/MaMpf-HD/mampf/pull/1151/commits/fce64b6f72f1172a1a656bb0b9dc583ff663382e), see the other csv comment.

app/services/csv_safe.rb
fosterfarrell9: Done in [fce64b6](https://github.com/MaMpf-HD/mampf/pull/1151/commits/fce64b6f72f1172a1a656bb0b9dc583ff663382e). Added a SafeCsv.generate wrapper whose << runs every row through CsvSafe automatically, migrated the two CSV exporters to it (dropping the manual per-cell calls), and added a custom Mampf/NoRawCsv cop that forbids raw CSV.generate/CSV.open and points to SafeCsv (only safe_csv.rb itself is exempt). Also hardened CsvSafe.cell for the whitespace-prefix bypass you flagged — it now checks the first non-whitespace char, so "  =cmd()" is caught, while leading control chars stay direct triggers. Rubocop + specs green, zeitwerk:check clean.

.github/workflows/schedule-deploy-image-rebuild.yml
fosterfarrell9: Fixed in [3a02abd](https://github.com/MaMpf-HD/mampf/pull/1151/commits/3a02abd82a91a871cf7e3e9f5d962196e04fed7d).

docker/production/Dockerfile
fosterfarrell9: Removed in [df88d76](https://github.com/MaMpf-HD/mampf/pull/1151/commits/df88d7661bc013b270d794f12842ec5b2958ccaa).

docker/production/Dockerfile
fosterfarrell9: Fixed in [c9e6327](https://github.com/MaMpf-HD/mampf/pull/1151/commits/c9e6327b1a0dfd25a5aa1f13b19714f890000979).

.github/workflows/linter.yml
fosterfarrell9: The yarn cache isn't actually disabled — only setup-node's built-in one is (package-manager-cache: false), and we cache manually via actions/cache on ~/.yarn/berry/cache + .yarn/install-state.gz + node_modules instead. Reason: setup-node's built-in cache: yarn runs its cache logic before the Corepack prepare --activate step, so at that point yarn is still the runner's preinstalled classic 1.x — yarn config get cacheFolder then resolves the wrong (classic) cache dir, not the Yarn 4 Berry cache. Disabling the built-in cache + a manual actions/cache for the Berry cache is the standard pattern for Corepack-managed Yarn 4 in CI, which is why it's structured this way. I added a short comment in [205f1e8](https://github.com/MaMpf-HD/mampf/pull/1151/commits/205f1e82a5b60930c46d2d98c9df24db049320d1).

.devcontainer/boot.sh
fosterfarrell9: The OR-Tools binary is built from a documented Dockerfile (clones OR-Tools v9.15 from Google's repo, builds with CMake on debian:trixie-slim; manual build/extract/verify/upload). It was only in the private mampf-infra repo, so I moved it into the public build-artifacts repo where the release lives — the artifact is now self-documenting: MaMpf-HD/build-artifacts#1. Build stays manual (rare rebuild on an OR-Tools/Debian bump), pinned by SHA256 here.

.devcontainer/devcontainer.json
fosterfarrell9: This is actually a regression (#1153 already made the change in settings.json), I removed it in [c8f5a76](https://github.com/MaMpf-HD/mampf/pull/1151/commits/c8f5a76200641ffe2290caf95f2e147c71885777).

app/controllers/media_controller.rb
fosterfarrell9: Auth was already skipped here before this PR (`[:play, :display]`) — the diff just expands the list to the media-serving endpoints added in this PR (screenshot, chapters/references VTT, stream_video, inline manuscript/geogebra, download). These serve free/released media to guests (no login). Authorization is still enforced per-medium by MediumAbility (`visible_for_user? || free?`), so skipping authentication here doesn't skip authorization.

docker/development/.env
fosterfarrell9: Yes, ClamAV runs in dev (compose service, app depends_on: clamav), but `CLAMAV_PORT=3310` is an internal service-to-service port — the app reaches clamav:3310 over the compose network. forwardPorts is only for ports a human hits from the host (app, Vite, mailcatcher, pgAdmin…), so ClamAV doesn't need forwarding.

docker/development/compose.yml
fosterfarrell9: Intentional — it's the Thruster topology: Rails runs on `MAMPF_PORT=3001 (= THRUSTER_TARGET_PORT)`, and Thruster owns the public HTTP port 3000 and proxies to Rails on 3001. compose publishes 3000:3000 and the devcontainer forwards 3000 (the Thruster/public port). 3001 is internal behind Thruster — you never hit Rails directly, so it's deliberately not forwarded. (The e2e app service uses MAMPF_PORT=3145, which is forwarded.)

docker/development/init.sh
fosterfarrell9: The script runs under set -Eeuo pipefail (nounset). `DB_SQL_PRESEED_URL` is optional and unset for most people (no preseed), and referencing it directly then aborts the whole init with "unbound variable". The ${DB_SQL_PRESEED_URL:-} default makes the check safe when it's unset; -n is just the explicit non-empty test. So the case it didn't cover before was the common "no preseed configured" one.

.devcontainer/boot.sh
fosterfarrell9: Added a comment in [c014127](https://github.com/MaMpf-HD/mampf/pull/1151/commits/c014127e678de68d5f77de36e1d3281581f5c2c0). Bundler stores our git-sourced gems (thredded, thredded-markdown_katex) as bare repos in the bundle cache, and git refuses to use them when safe.bareRepository=explicit (fatal: cannot use bare repository), which breaks bundle install — so we set it to all (git's default). Ref: [rubygems/rubygems#7515](https://github.com/rubygems/rubygems/issues/7515). The chown fixes the bundle-volume ownership.

.gitignore
fosterfarrell9: That's Yarn 4's own recommended .gitignore for the node-modules linker: commit .yarn/patches, .yarn/releases, .yarn/plugins, .yarn/sdks (SDKs power editor TS/ESLint integration), ignore .yarn/cache, .yarn/install-state.gz, .yarn/unplugged, .yarn/build-state.yml. So the exclusion list is intentional — fine to keep.

.devcontainer/install-ansible.sh
fosterfarrell9: Fixed in [fc7ef11](https://github.com/MaMpf-HD/mampf/pull/1151/commits/fc7ef11bf318d8ca0943673f8a5bcd5c9bb74696).

docker/development/init.sh
fosterfarrell9: Fixed in [fc7ef11](https://github.com/MaMpf-HD/mampf/pull/1151/commits/fc7ef11bf318d8ca0943673f8a5bcd5c9bb74696).

docker/production/Dockerfile
fosterfarrell9: Fixed in [fc7ef11](https://github.com/MaMpf-HD/mampf/pull/1151/commits/fc7ef11bf318d8ca0943673f8a5bcd5c9bb74696).

PR #1155, opened 2026-06-21
deps: Upgrade Puma to v8

Splines:
This is a major version upgrade. Please review the [upgrade guide](https://github.com/puma/puma/blob/main/docs/8.0-Upgrade.md). I went through the points and didn't find anything that affects us (please double-check though).


fosterfarrell9 (dismissed):
Works as it should.

PR #1163, opened 2026-06-21
Fix bugs in groups tab

fosterfarrell9:
In this PR we fix two bugs in the groups tab. To reproduce these in `next`, successfully finish a campaign and the create a tutorial in a new campaign. Then the two things occur:

- the tutorial is actually created in the "no campaign" area
- both areas - campaign and no campaign - are collapsed


Splines (approved):
Nice, thanks for this fix, I didn't account for completed campaigns 😅

PR #1164, opened 2026-06-22
Add database constraint for policy config

fosterfarrell9:
We add a non-nil constraint for policy config.

### DIFF
--- a/db/migrate/20260626000000_set_registration_policy_config_not_null.rb
+++ b/db/migrate/20260626000000_set_registration_policy_config_not_null.rb
@@ -3,11 +3,9 @@ class SetRegistrationPolicyConfigNotNull < ActiveRecord::Migration[8.0]
   end
 
   def up
-    RegistrationPolicy.where(config: nil).find_each do |policy|
-      # rubocop:disable Rails/SkipsModelValidations
-      policy.update_columns(config: {})
-      # rubocop:enable Rails/SkipsModelValidations
-    end
+    # rubocop:disable Rails/SkipsModelValidations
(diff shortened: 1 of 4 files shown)


db/migrate/20260626000000_set_registration_policy_config_not_null.rb
```diff
@@ -0,0 +1,19 @@
+class SetRegistrationPolicyConfigNotNull < ActiveRecord::Migration[8.0]
+  class RegistrationPolicy < ApplicationRecord
+    self.table_name = "registration_policies"
+  end
```
Splines, line 3: Why does one need this? Can't Rails figure out the table name on its own?

Splines (changes requested):
LGTM, just one small question.

db/migrate/20260626000000_set_registration_policy_config_not_null.rb
fosterfarrell9: I think this is because we don't namespace below: the model is `Registration::RegistrationPolicy` but the table is called `registration_policies`. I am not sure how autoloading works in migrations. You might try to leave this out and write `Registration:RegistrationPolicy" below.

Splines: Apparently, you only need an empty class, with this it works as well:

```rb
class RegistrationPolicy < ApplicationRecord
end
```

Tested with
```sh
bin/rails db:migrate:down:primary VERSION=20260626000000
bin/rails db:migrate:up:primary VERSION=20260626000000
```

Splines (approved):
LGTM!

PR #1165, opened 2026-06-27
Fix broken playwright test

fosterfarrell9:
#1163 fixed a bug in the groups tab, but it didn't update the corresponding playwright test. So the playwright test fails currently, because it expects the faulty outcome. This PR fixes it.


Splines (approved):
LGTM!

PR #1169, opened 2026-07-04
Reorganize lecture home

Splines:
This PR adds a dedicated lecture Home page at `/lectures/:id/home` and keeps `/lectures/:id` as content route. The previous lecture overview is now labeled `Outline`, while `Home` becomes the lecture-level landing page for announcements/news and, when relevant, registration/roster workflow content.

Main changes:
- add a global lecture Home page with lecture header, news, and optional workflow modules

Screenshot 2026-07-08 171719

- keep the canonical lecture member route on `/lectures/:id`; `/lectures/:id/outline` now redirects there
- rename the former overview tab to `Outline`
- route non-subscribers from lecture content to Home instead of the legacy subscribe page
- allow staff and non-subscribed students to access Home, while keeping actual registration restricted to eligible students
- send lecture announcement notifications to Home
- add a small Home-tab update marker when Home has lecture news/forum updates
- show workflow content on Home only when relevant; otherwise Home falls back to a minimal lecture landing state

Design decisions:
- Home is now a universal lecture page
- Outline remains the canonical content page
- users without content access are redirected from content to Home
- registration/roster-specific sections on Home are conditional


config/routes.rb
```diff
@@ -307,12 +311,10 @@
                 as: :registration_campaigns
     end
 
-    constraints ->(_req) { Flipper.enabled?(:registration_campaigns) } do
-      resources :registration, only: [:index],
-                               controller: "registration/user_registrations",
-                               as: :user_registrations,
-                               defaults: { project: "registration" }
-    end
+    get "",
```
fosterfarrell9: `lecture_path` hijacks the resource route name, breaking lecture update & delete entirely.
`get "lectures/:id/outline", as: "lecture"` claims the lecture route name before `resources :lectures`, so the PATCH/PUT/DELETE routes at `/lectures/:id` are generated nameless. Every `form_with model: lecture` (all edit tabs: _preferences, _organizational_concept, _comments, _assignments, _people) now submits `PATCH /lectures/:id/outline`, and the delete link (lectures/edit/_header.html.erb:38) sends `DELETE /lectures/:id/outline` — both fall through to the `match "*path" → main#error` catch-all. Saving any lecture setting or deleting a lecture is broken. (Verified by reproducing the route set in a standalone Rails 8 RouteSet.)

Also: The old registration index route sat inside the Flipper constraints block; the replacement get "" → lectures/home#show does not, the deleted sidebar helper was the only other flag check, and nothing in the controller/templates checks the flag. Flag off ⇒ the bare lecture URL still runs the campaign services and renders the registration UI. The gate was removed, not moved.

app/controllers/lectures/home_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    before_action :set_lecture
+    before_action :set_user_locale
+
+    def current_ability
+      @current_ability ||= RegistrationUserRegistrationAbility.new(current_user)
+    end
+
+    def show
+      authorize! :index, @lecture
```
fosterfarrell9, line 15: the home page is unreachable for staff and non-subscribed students, while the sidebar advertises it to everyone. authorize! :index via RegistrationUserRegistrationAbility requires student_registration_participant? (subscribed, not teacher/editor/tutor). Teachers clicking their own lecture's Home link get CanCan::AccessDenied → root + alert; the template's privileged edit button is dead code (its audience can never render it); the old registration_sidebar_visible? guard that hid the link from exactly these users was deleted. 

Already fixed by #1173 (lecture-home-access), which opens :index to published lectures + staff.

app/controllers/lectures/home_controller.rb
```diff
@@ -0,0 +1,42 @@
+module Lectures
```
fosterfarrell9, line 1: Outside of the actual code changes: app/models/notification.rb:35 
announcement notifications become dead-ends.
Notification#path → Lecture#path → lecture_path → /outline, and the PR removed the news card from show.html.erb (it now lives only on the home page). Clicking an announcement notification lands on a page that no longer displays announcements.

Also:
app/frontend/lectures/search/_lecture.html.erb:1 + profile_controller.rb:91 + application_helper.rb:252 — every entrance except the sidebar bypasses the new home.
Search cards, the post-subscribe redirect, edit_or_show_lecture_path, and Lecture#path all still resolve to the outline; for non-subscribers the outline immediately bounces to the subscribe/passphrase wall — so the audience the home page was built for never reaches it. 

Largely addressed by #1173, which redirects non-subscribers from content URLs to the home page; a deliberate decision about the canonical URL (finding 1's rename) would resolve it at the root.

app/controllers/lectures/home_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      @current_ability ||= RegistrationUserRegistrationAbility.new(current_user)
+    end
+
+    def show
+      authorize! :index, @lecture
+
+      @campaigns_details = ::UserRegistrations::LectureCampaignsService
+                           .new(@lecture, current_user)
+                           .call
+      @rosterized_entries = Rosters::StudentMaterializedResultResolver
```
fosterfarrell9: ~100–130 SQL queries on every bare-lecture visit.
The full campaign/roster pipeline now runs unconditionally on the default landing path: SelfRosterOptionsQuery re-evaluates allow_self_add?/allow_self_remove? per rosterable in both select and sort_by with no memoization or eager-loading (~4–6 queries × per tutorial/talk/cohort — dominant term), the resolver and the query each load tutorials/cohorts/talks separately, CampaignDetailsService eagerly traces both policy phases per campaign, plus the Thredded unread aggregation. The services pre-date this PR, but the PR moves them from a deliberately-visited, flag-gated page onto every lecture click. Cheapest wins: memoize/eager-load in SelfRosterOptionsQuery? Maybe there are better options.

config/routes.rb
```diff
@@ -307,12 +311,10 @@
...
+        to: "lectures/home#show",
+        as: :user_registrations,
+        defaults: { project: "registration" }
```
fosterfarrell9: The home route is named as: :user_registrations with defaults: { project: "registration" } although it renders the home page — maybe rename to lecture_home / project: "home" (call sites: sidebar, two controller redirects, specs — mechanical); set_lecture's || params[:id]&.to_i fallback is dead (the only route always supplies :lecture_id); 

Or leave it as is, since it is temporary

fosterfarrell9 (changes requested):

config/routes.rb
fosterfarrell9: I fixed the route collision in [ea1ae25](https://github.com/MaMpf-HD/mampf/pull/1169/commits/ea1ae257171bcfae054baaa88ee886bd2425667d). The canonical lecture member route is back on /lectures/:id, so lecture_path once again refers to the normal Lecture member route and Rails can generate the correct update/delete/form targets. Lecture Home moved to /lectures/:id/home, and /lectures/:id/outline now only redirects to the canonical content route. The earlier concern about the removed flag gate is now intentional by design: Home is global for all lectures, while workflow-specific content on Home is conditional.

app/controllers/lectures/home_controller.rb
fosterfarrell9: I fixed this in [ea1ae25](https://github.com/MaMpf-HD/mampf/pull/1169/commits/ea1ae257171bcfae054baaa88ee886bd2425667d): Lecture announcement notifications no longer route through Lecture#path to the content/outline page. Notification#path now sends lecture announcements directly to lecture Home, which is where the news card lives. So clicking a lecture-announcement notification now lands on the page that actually displays the lecture news/announcement surface.

app/controllers/lectures/home_controller.rb
fosterfarrell9: Fixed via merge of #1173.

app/controllers/lectures/home_controller.rb
fosterfarrell9: The worst regression path is gone because Home is no longer the default lecture route, but no dedicated query optimization has been done yet. We defer that to a later PR.

config/routes.rb
fosterfarrell9: I did a home route naming cleanup in [58a53fc](https://github.com/MaMpf-HD/mampf/pull/1169/commits/58a53fc08787774a2fe0d9e04a7829b583f1894c).

app/controllers/profile_controller.rb
```diff
@@ -80,11 +80,24 @@ def subscribe_lecture
     if !@lecture.published? && !current_user.admin &&
        !@lecture.edited_by?(current_user)
       @unpublished = true
+      if html_redirect_flow?
+        return redirect_to lecture_home_path(@lecture),
+                           alert: t("admin.lecture.no_rights")
+      end
+      return
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [df7b90c](https://github.com/MaMpf-HD/mampf/pull/1169/commits/df7b90c360db2531727119dbcbb7bc9f44379444). The unpublished-lecture redirect no longer chains through `lecture_home_path`, which could immediately raise `CanCan::AccessDenied` for the same user and replace the intended alert with the generic rescue message. It now redirects directly to `root_path` with `admin.lecture.no_rights`, so the user always lands on a reachable page and sees the correct message

app/models/lecture.rb
```diff
@@ -811,6 +811,14 @@ def ensure_roster_membership!(user_ids)
...
       unique_by: [:user_id, :lecture_id]
     )
+
+    # Roster membership implies a subscription: members need access to the
+    # lecture's content, even when subscribing is gated by a passphrase
+    # (a roster seat is a stronger credential than a shared passphrase).
+    LectureUserJoin.insert_all(
+      attributes,
+      unique_by: [:lecture_id, :user_id]
+    )
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed. Lecture#ensure_roster_membership! now wraps the `LectureMembership.insert_all`and `LectureUserJoin.insert_all` calls in a single transaction, so roster membership and subscription are created atomically. This prevents the inconsistent partial state where a user could be added to the lecture roster without receiving lecture access.

app/frontend/lectures/show/_content.html.erb
```diff
@@ -1,35 +1,44 @@
...
+        <%= t("admin.lecture.#{lecture.sort}_content") %>
+      </h5>
+      <% if current_user.admin || lecture.course.edited_by?(current_user) %>
+        <%= link_to "",
+                    edit_lecture_path(lecture),
+                    class: "far fa-edit text-dark",
+                    data: { toggle: "tooltip",
+                            placement: "bottom" },
+                    title: t("buttons.edit"),
+                    style: "text-decoration: none;" %>
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [418c90b](https://github.com/MaMpf-HD/mampf/pull/1169/commits/418c90bdc591494ab42d0bf87fb3e309e7ecaf0a).

Splines (commented):
LGTM except for some subscribe-related inconsistencies discussed on Basecamp. Those will be fixed after the next time-critical release.

PR #1177, opened 2026-07-10
Fix datetimepicker init in the publish modal via Stimulus

fosterfarrell9:
Teachers reported that creation of assignments in the media publish modal is not always working

The release-date and assignment-deadline pickers are initialized imperatively (`$(document).ready` + a `turbo:load` call), which loses the race with the per-page module on Turbo navigations and sometimes left the pickers uninitialized, blocking assignment creation. In this PR we change this: We manage them with the existing datetimepicker Stimulus controller, guard the imperative initializer against double-initializing Stimulus-managed elements, and fix the assignment button targeting the wrong picker.

**Note:** Even better would be to migrate all existing datetimepickers to the stimulus controller, but I don't have the time to do this now.

### DIFF
--- a/app/frontend/js/datetimepicker.js
+++ b/app/frontend/js/datetimepicker.js
@@ -11,7 +11,7 @@ export function initializeDatetimePickers(container = document) {
     element = $(element);
     // Skip elements handled by the datetimepicker Stimulus controller; it inits
     // them on connect(), so initializing here too would create a second picker.
-    if ((element.attr("data-controller") || "").split(/\s+/).includes("datetimepicker")) {
+    if (element.attr("data-controller")?.split(/\s+/).includes("datetimepicker")) {
       return;
     }
     if (element.data("td-initialized")) {
--- a/app/frontend/js/datetimepicker.js
+++ b/app/frontend/js/datetimepicker.js
@@ -9,6 +9,11 @@ export function initializeDatetimePickers(container = document) {
 
   pickerElements.each((i, element) => {
     element = $(element);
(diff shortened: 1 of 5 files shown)


app/frontend/js/datetimepicker.js
```diff
@@ -9,6 +9,11 @@ export function initializeDatetimePickers(container = document) {
 
   pickerElements.each((i, element) => {
     element = $(element);
+    // Skip elements handled by the datetimepicker Stimulus controller; it inits
+    // them on connect(), so initializing here too would create a second picker.
+    if ((element.attr("data-controller") || "").split(/\s+/).includes("datetimepicker")) {
```
Splines: It's better here to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in JS, this way the subsequent methods are not even called if data-controller is not present. See my change in 6f4dd27e1b0.

app/views/media/_publish_modal.html.erb
```diff
@@ -57,6 +57,7 @@
           <div
             class="mb-3 input-group td-picker"
             id="release-date-picker"
+            data-controller="datetimepicker"
```
Splines, line 60: I additionally wanted to have the datetimepicker focus automatically when the "zum folgenden Zeitpunkt" is selected on the page. Changed in 3363d330bf025 (and also fixed a bug there due to an early return).

Splines (approved):
LGTM. I also improved the related Playwright test in 5806db6bd1c. Please also see my other comments that you are free to resolve.

PR #1178, opened 2026-07-10
Fix JS event handling on page init of lecture & profile

fosterfarrell9:
Pretty often it happens that some js events dont fire: Most often this happens to me on the profile page locally, e.g, when I want to change the language setting of a user by clicking the corresponding radio button and the "save your changes" banner does not appear. Similar with the lecture edit page: Hovering over a lesson date should highlight the associated tags. That also doesn't work reliably. This PR fixes these.

**Note:** The pattern that is replaced here can be found in many other places, but I currently don't have the time to fix it all.


app/frontend/js/profile.js
```diff
@@ -1,34 +1,32 @@
...
+  $("#profileChange").removeClass("d-none");
+  $("#profileChangeBottom").removeClass("d-none");
+});
 
-  $('input:checkbox[id^="user_lecture"]').on("change", function () {
-    const courseId = this.dataset.course;
-    const lectureId = parseInt(this.dataset.lecture);
-    const checkedCount = $(`input:checked[data-course="${courseId}"]`).length;
-    const authRequiredLectureIds = $(`#lectures-for-course-${courseId}`).data("authorize");
+$(document).on("change", 'input:checkbox[id^="user_lecture"]', function () {
```
Splines: The problem with attaching event handlers to `document` is that this still doesn't guard about the handler being registered twice, e.g. when the `profile.js` is again evaluated. So your proposed approach is also not idempotent as it should be, especially in conjunction with Hotwire.

But luckily Hotwire offers the solution for this directly: Stimulus controllers. They automatically connect and disconnect controllers, so no need to manually do this anymore. I've migrated the files to Stimulus controllers in 2c19009a5fb.

Splines (dismissed):
Please carefully review the rewrite to Stimulus controllers (see comment below) and ensure that everything still works.

app/frontend/js/profile.js
fosterfarrell9: Migrating to stimulus is definitely the right direction.

Splines (approved):

PR #1179, opened 2026-07-10
Fix GeoGebra applet failing to load in production

fosterfarrell9:
When we switched to vite, GeoGebra assets stopped working in production apparently /(but nobody complained).The GeoGebra page's script was registered as a direct Vite entrypoint via `vite_javascript_tag "js/geogebra.coffee"`. But `.coffee` is not among vite-plugin-ruby's known entrypoint types (js/ts/jsx/tsx/css/…), so in a production build the file is not transpiled but copied verbatim as a raw asset and referenced with a `.coffee` extension. The static server (nginx/ soon Thruster) has no MIME mapping for `.coffee` and serves it as `text/plain`, which the browser rejects for a module script:

> Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "text/plain".

Development worked because the Vite dev server transpiles the `.coffee` on the fly. I converted the entrypoint to plain JavaScript so it is transpiled and served as `.js`.


Splines (changes requested):
Nice. Can we also get (at least a simple) Playwright test for this?

spec/vite_entrypoints_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+     "postcss"]
+  end
+
+  it "never point at a file extension vite cannot transpile" do
+    # only the script/style helpers register entrypoints; vite_image_tag & co.
+    # reference plain assets, which are served as-is and are fine.
+    helpers = /vite_(?:javascript|typescript|stylesheet)_tag/
+
+    offenders = Rails.root.glob("app/{views,frontend}/**/*.erb").flat_map do |path|
+      entrypoints = path.read.scan(/#{helpers}\s+"([^"]+)"/).flatten
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [97e95b8](https://github.com/MaMpf-HD/mampf/pull/1179/commits/97e95b8000a86f951efa3df03c4ce088f55076be).

spec/vite_entrypoints_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+# Anything else is copied verbatim into the production build and served with a
+# MIME type the browser rejects for a module script — so the page silently
+# breaks in production while working fine behind the dev server, which
+# transpiles on the fly. This is what happened to js/geogebra.coffee.
+RSpec.describe("Vite entrypoints") do
+  let(:transpilable_extensions) do
+    ["js", "mjs", "jsx", "ts", "tsx", "css", "scss", "sass", "less", "styl", "stylus", "pcss",
+     "postcss"]
+  end
+
```
Splines, line 17: I'm wondering where you got this list from. Could you add a link as comment, e.g. [this one](https://github.com/ElMassimo/vite_ruby/blob/3edc67485fd4021072b43ad3a073ffa0fa851ede/vite-plugin-ruby/src/constants.ts#L23)? Also, I don't see mjs in this list, but you included it here, is there a specific reason for it?

spec/vite_entrypoints_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  it "never point at a file extension vite cannot transpile" do
```
Splines, line 18: nice, good idea to add this test!

spec/vite_entrypoints_spec.rb
Splines: Might also be worth to split the line before listing all CSS extensions in a new line (and use the [order](https://github.com/ElMassimo/vite_ruby/blob/3edc67485fd4021072b43ad3a073ffa0fa851ede/vite-plugin-ruby/src/constants.ts#L11-L20) Vite defines for the stylesheet extensions)

Splines (changes requested):
Thanks for the tests!

spec/vite_entrypoints_spec.rb
fosterfarrell9: Good points, i incorporated these in [8ab1b0f](https://github.com/MaMpf-HD/mampf/pull/1179/commits/8ab1b0f0c50b87fb4a731aaf9ff48254f84ae13a).

Splines (approved):

PR #1180, opened 2026-07-11
Close the vignette editor subheadings with the correct tag

fosterfarrell9:
The "info slides" and "slides" subheadings in the questionnaire editor open an `<h2>` but close with `</h1>`, which makes the markup invalid. The page already has a proper `<h1>`, so `<h2>` is the right level - only the closing tags were wrong.


Splines (approved):
Good catch. Might want to also activate Herb for these files at some point to fix issues like this directly via linting.

PR #1181, opened 2026-07-11
Don't block registration for students who are in a group they can't leave

fosterfarrell9:
### The bug

While setting up registration for a proseminar (in staging, luckily), I ran into this. I created an "interest group" cohort, had a student register their interest, and closed that campaign. Then I opened a second, preference-based campaign for the actual talks, with the interest campaign as a prerequisite.

The student who had registered their interest — and therefore met the prerequisite — still couldn't pick a talk. Every talk tile was disabled with "you cannot join this tutorial because you cannot leave your tutorial", even though the prerequisite check right above it showed green.

Screenshot 2026-07-11 134706

### What was going on

Being in the interest cohort tripped a guard meant for a different situation: if you're already in a tutorial you can't leave, you shouldn't be able to self-enrol into a second one. But that guard was applied lecture-wide — to every group type and every campaign — so a locked cohort membership blocked registering for talks, which have nothing to do with it. (The hard-coded "tutorial" in the message is the same over-generalisation; it showed even for cohorts.)

And the guard doesn't belong on campaign registration at all: registering never touches the roster, it just records the registration. The actual group assignment happens at finalization, which already moves students between groups freely. So there's nothing to pre-empt — the student can register regardless of what they're currently in.

### The fix

I removed the "can't leave your group" guard from campaign registration (FCFS and preference-based).
Kept it only where it's real — immediate self-enrolment — and scoped it to tutorials, the one group type where joining one means leaving another. It no longer fires for cohorts or talks, and the "tutorial" wording is now always accurate.
Added specs for the reported case and for the self-enrolment cases that should still block.

One thing worth a conscious nod for review: this settles that "no-leave" means a student can't remove themselves — but a new campaign's allocation can still move them, which is what finalization already does today.

The above scenario now looks like this:

Screenshot 2026-07-11 141206


spec/models/rosters/self_roster_availability_spec.rb
```diff
@@ -5,32 +5,51 @@
...
       rosterable = instance_double(
         "Rosterable",
+        roster_exclusive_within_lecture?: true,
         user_allocated?: true,
-        config_allow_self_add?: false,
-        config_allow_self_remove?: true
+        config_allow_self_remove?: false
       )
-      availability = described_class.new(lecture, user)
+      allow(rosterable).to receive(:config_allow_self_remove?).and_return(true)
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [59715b6](https://github.com/MaMpf-HD/mampf/pull/1181/commits/59715b60564b823fd75a3e5db062a97c342fd88c).

spec/requests/registration/user_registrations_spec.rb
Splines, line 1: The newly added tests here operate on a CSS level, e.g. they test for existance of classes. I think it'd be nicer to have these added tests here as Playwright tests instead that check if a button with name "Withdraw" exists or not and that the user actually sees what is expected.

Splines (changes requested):
Nice, I really like the abstraction here with the `roster_exclusive_within_lecture?` and `conflicting_lecture_membership` overwritten in the tutorial rosterable.

One request down below.

spec/requests/registration/user_registrations_spec.rb
fosterfarrell9: I address this in [aa2d472](https://github.com/MaMpf-HD/mampf/pull/1181/commits/aa2d47299802aee0b21f5a02ece7b8c9eb1eb363).

Splines (approved):

PR #1182, opened 2026-07-11
Modal for self-enrollment modes after finalizing a campaign

fosterfarrell9:
When you finalize a registration campaign, the groups it created default to **self-enrollment disabled**. So the moment allocation finishes, students are locked into whatever group they were assigned — they can't leave or switch — and nothing ever asks the teacher whether that's actually what they want. You can change it afterwards through the little per-tile dropdown, but it's tucked in a corner and easy to miss, so in practice everyone just ends up silently locked in (at least that what happened with my campaign without me noticing it). It is not strictly a bug, but to me it feels like a UX bug, that's why I labelled this as a bug.

This PR **adds a modal right after finalization** that makes the teacher decide: may students join/leave their groups on their own, or not? It reuses the existing self-materialization modes and applies the chosen one to every group in the campaign. Keeping everything closed is still one click if that's what they want.

Screenshot 2026-07-11 172653

#### Two smaller things while I was in there:

The colored bar on top of each group tile already encodes self-enrollment state (green = on, blue = campaign running, grey = off), but nothing told you that — so it now has a tooltip that matches the state, same text as the dropdown icon.
Fixed a bug I hit along the way: the roster drag controller's turbo:stream-render listener was unbound, so this was document and it threw this.clearHighlight is not a function whenever a turbo stream rendered while the participants panel was open (e.g. toggling that very self-enrollment dropdown).


app/controllers/registration/campaigns_controller.rb
```diff
@@ -175,6 +175,29 @@ def reopen
...
+    # allocated group.
+    def self_service
+      unless @campaign.completed?
+        return respond_with_flash(
+          :alert, t("registration.campaign.self_service.only_when_completed"),
+          redirect_path: registration_campaign_path(@campaign)
+        )
+      end
+
+      @campaign.apply_self_materialization_mode!(params[:self_materialization_mode])
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [70da66f](https://github.com/MaMpf-HD/mampf/pull/1182/commits/70da66ffc090188b8e5045b0d96a6a95b13e9e9e).

app/frontend/registration/campaigns/_self_service_modal.html.erb
```diff
@@ -0,0 +1,75 @@
...
+
+          <% current_mode = campaign.shared_self_materialization_mode %>
+
+          <%= select_tag :self_materialization_mode,
+                         options_for_select(
+                           self_service_mode_options(current_mode),
+                           "add_and_remove"
+                         ),
+                         class: "form-select",
+                         data: { testid: "self-service-mode" } %>
```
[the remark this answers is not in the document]
fosterfarrell9: Pushing back on this one, while taking the underlying point.

The modal only ever renders right after finalization, and campaign groups cannot have self-materialization enabled (cannot_enable_both_campaign_and_self_materialization) — so the current mode at that moment is always disabled. Preselecting it would make "Apply" a no-op, leaving students silently locked into their groups: exactly what this modal exists to prevent. The escape hatch is the second button ("Keep self-enrollment disabled"), not the preselection.

You're right that the code reads like a bug, though. Added a comment in [b7c33c1](https://github.com/MaMpf-HD/mampf/pull/1182/commits/b7c33c14bc7253c03c413c51c720a3fdae89577d) explaining the intent plus a regression test pinning both behaviours (current mode labelled as such, recommendation preselected). It fails against your suggested change — so nobody can quietly clean this up later.

app/controllers/registration/allocations_controller.rb
```diff
@@ -94,17 +94,40 @@ def finalize
...
+                                lecture: lecture,
+                                expanded_campaign_id: @campaign.id
+                              }),
+          *refresh_roster_streams(lecture)
+        ]
+
+        if @campaign.registerables.any?
+          # The self-service modal is the post-finalization surface, so we do
+          # NOT also pop a flash over it (fixed, top-of-screen, it would cover
+          # the modal). The finalization summary is folded into the modal.
```
Splines, line 110: Good catch, would not have thought about that myself.

app/frontend/roster/components/group_tile_component.rb
```diff
@@ -195,6 +195,15 @@ def top_bar_class
     end
   end
 
+  # Explains the top-bar colour on hover. Campaign (blue) tiles describe the
```
Splines: nitpick: let's keep American English everywhere as usual, thus "color" instead of "colour"

app/frontend/roster/roster_drag.controller.js
```diff
@@ -17,13 +17,16 @@ export default class extends Controller {
   highlightedTile = null;
 
   connect() {
+    this.boundRefreshDropZones = this.refreshDropZones.bind(this);
```
Splines: I don't like this approach with binding `this`, it doesn't make it clear from the code itself what is going on.

Instead, what about using an action directly in the markup, e.g. something like
```
data-action="turbo:stream-render@document->roster-drag#refreshDropZones"
```

This way, Stimulus will also directly remove the listener for us upon controller disconnect.

But you might want to watch out for our custom "turbo:stream-render" (see `initHotwire.js`), but it should work fine even with that.

Splines (changes requested):
I think this modal is really a nice addition! Here is my first review (I only have to review the tests afterwards, maybe a Playwright test also makes sense here?).

app/frontend/roster/components/group_tile_component.rb
fosterfarrell9: Fixed in [4439bc9](https://github.com/MaMpf-HD/mampf/pull/1182/commits/4439bc9c5438bd88f64a05f6ff31ab240ba07a9b).

app/frontend/roster/roster_drag.controller.js
fosterfarrell9: Yes, this works. I changed this in [7662fc4](https://github.com/MaMpf-HD/mampf/pull/1182/commits/7662fc46628dfc74f0fdf812ec0b80f334dc23c0) and added a playwright test for the new approach.

e2e/roster_drag.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  await factory.create("tutorial", [],
+    { lecture_id: lecture.id, skip_campaigns: true });
+
+  const lectureEditPage = new LectureEditPage(page, lecture.id);
+  await lectureEditPage.goto();
+  await lectureEditPage.groupsTab.click();
+
+  // clicking a tile loads the side panel, which connects the drag controller
+  await page.locator("[data-roster-key]").first().click();
+  await expect(page.locator("[data-controller~=\"roster-drag\"]")).toBeVisible();
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [eb14b1a](https://github.com/MaMpf-HD/mampf/pull/1182/commits/eb14b1a5703a55a27deba243d10b4b689713357c).

e2e/roster_drag.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    let ran = false;
+    controller.clearHighlight = () => {
+      ran = true;
+    };
+
+    document.dispatchEvent(new Event("turbo:stream-render"));
+
+    return ran;
+  });
```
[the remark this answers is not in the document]
fosterfarrell9: Fixed in [eb14b1a](https://github.com/MaMpf-HD/mampf/pull/1182/commits/eb14b1a5703a55a27deba243d10b4b689713357c).

e2e/registration/self_service_after_finalization.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Finalizing a campaign leaves every group with self-enrollment disabled, so
+ * students end up silently locked into their allocation. The modal makes the
+ * teacher decide.
+ *
+ * Worth covering end to end: the modal's form has to submit via Turbo. It once
+ * rendered as a Rails-UJS remote form instead (form_with_generates_remote_forms),
+ * so "Apply" did nothing at all — while the server side stayed perfectly correct,
+ * which is exactly why the request specs were happy.
```
Splines: What does "which is exactly why the request specs were happy" refer to? I don't see any value in this whole last paragraph starting with "Worth covering" and I don't know what this means. In case it's relevant, maybe you can explain it to me? Or fix it directly in this PR?

e2e/registration/self_service_after_finalization.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  test("applies the chosen mode to the campaign's groups",
+    async ({ factory, teacher: { page, user } }) => {
+      const campaign = await finalizeCampaign(page, factory, user.id);
+
+      const modal = page.getByTestId("self-service-modal");
+      await expect(modal).toBeVisible();
+
+      // the recommendation is preselected, not the current state — preselecting
+      // "disabled" would make the primary button a no-op
```
Splines: Please, let's write comments by humans or no comments at all, but not by AI 😥

```
// the recommended option is preselected (instead of the current state being preselected)
```

e2e/roster_drag.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+ * The drop zones have to be rebuilt whenever a turbo stream re-renders the
+ * tiles, which the roster panel wires up with
+ *
+ *   data-action="turbo:stream-render@document->roster-drag#refreshDropZones"
+ *
+ * That event is our own (see initHotwire.js) and does not bubble — it is
+ * dispatched straight onto `document`. Nothing else covers this wiring, and a
+ * "no console errors" check would not either: if the action never fired, there
+ * would be no error, just silently stale drop zones. Hence the spy below.
+ */
```
Splines: I don't see any value in this test: you are essentially testing that Stimulus does its job, i.e. that if this `data-action` is defined, our function defined in the respective Stimulus controller is really called. It's Hotwire's job to test this, not ours.

Instead, what you could do is test implicit behavior that the user sees, e.g. assert that something on screen is correctly shown (which is not correctly shown if the action fails). This might indeed be tricky for this specific case of refreshing drop zones, but the current test is not helping.

spec/components/roster/group_tile_component_spec.rb
```diff
@@ -166,6 +166,45 @@
     end
   end
 
+  # top_bar_tooltip relies on the ViewComponent translate helper, so it is
+  # exercised through the rendered bar's title rather than called directly.
+  describe "top-bar tooltip" do
+    def bar_title(instance)
+      render_inline(instance).css(".tutorial-gtile-top-bar").first["title"]
```
Splines: Don't like too much that we have to depend on the specific CSS class here for the test to work.

spec/components/roster/group_tile_component_spec.rb
```diff
@@ -166,6 +166,45 @@
     end
   end
 
+  # top_bar_tooltip relies on the ViewComponent translate helper, so it is
+  # exercised through the rendered bar's title rather than called directly.
```
Splines: Too poetic: this is code, not prose. What does "it is exercised through" even mean here, this is very vague.

spec/models/registration/campaign_spec.rb
```diff
@@ -968,4 +968,58 @@
...
+        .to eq(["disabled"])
+    end
+
+    it "rejects an unknown mode" do
+      campaign = create(:registration_campaign, :completed,
+                        :first_come_first_served, items_count: 2)
+
+      expect { campaign.apply_self_materialization_mode!("nonsense") }
+        .to raise_error(ArgumentError)
+    end
```
Splines: Would be nice to check that after this rejection, the mode did not get changed.

spec/models/registration/campaign_spec.rb
```diff
@@ -968,4 +968,58 @@
...
+      expect(campaign.shared_self_materialization_mode).to eq("disabled")
+    end
+
+    it "returns nil when the groups differ" do
+      campaign = create(:registration_campaign, :completed,
+                        :first_come_first_served, items_count: 2)
+      campaign.registerables.first.update!(self_materialization_mode: :add_only)
+
+      expect(campaign.shared_self_materialization_mode).to be_nil
+    end
```
Splines, line 1035: This relies implicitly on the default mode not being "add_only", make this explicit instead by setting the self-materization mode manually for _all_ created campaigns.

spec/requests/registration/allocations_spec.rb
```diff
@@ -297,6 +297,75 @@
...
+            .to eq(["disabled"])
+
+          doc = Nokogiri::HTML(response.body)
+          select = doc.css("select[name='self_materialization_mode']")
+
+          # ... the current state is labelled as such ...
+          expect(select.css("option[value='disabled']").text)
+            .to include(
+              I18n.t("registration.campaign.self_service.current_state_suffix")
+            )
```
Splines: Reading something like this, I think it's time to convert these newly added spec/request tests to Playwright tests instead, at least those of `allocations_spec.rb`

Splines (changes requested):
In general, works like it should, nice. Still some more remarks regarding tests.

Could you also replace the image in the PR description with the English version? This is because I frequently want to take those images for the release notes, which should be in English.

e2e/registration/self_service_after_finalization.spec.ts
fosterfarrell9: This refers to a naive implementation done before that wasn't working (even though the spec passed). But this is history, so deleted it in [5355230](https://github.com/MaMpf-HD/mampf/pull/1182/commits/53552306e389ed414aee638577eb7f5d5ab8bacc).

e2e/registration/self_service_after_finalization.spec.ts
fosterfarrell9: Fixed in [5355230](https://github.com/MaMpf-HD/mampf/pull/1182/commits/53552306e389ed414aee638577eb7f5d5ab8bacc).

spec/models/registration/campaign_spec.rb
fosterfarrell9: Fixed in [4e083f2](https://github.com/MaMpf-HD/mampf/pull/1182/commits/4e083f2640ccc7f7616f679813f646e990187e06).

spec/models/registration/campaign_spec.rb
fosterfarrell9: Fixed in [4e083f2](https://github.com/MaMpf-HD/mampf/pull/1182/commits/4e083f2640ccc7f7616f679813f646e990187e06).

spec/components/roster/group_tile_component_spec.rb
fosterfarrell9: Removed in [901dfeb](https://github.com/MaMpf-HD/mampf/pull/1182/commits/901dfebc10a35d6db458b50b8589c831c3382206).

spec/components/roster/group_tile_component_spec.rb
fosterfarrell9: Introduced a data attribute in [901dfeb](https://github.com/MaMpf-HD/mampf/pull/1182/commits/901dfebc10a35d6db458b50b8589c831c3382206).

e2e/roster_drag.spec.ts
fosterfarrell9: I deleted the test in [5355230](https://github.com/MaMpf-HD/mampf/pull/1182/commits/53552306e389ed414aee638577eb7f5d5ab8bacc). I also don't want to spend time on building a (non-trivial) actual test here.

spec/requests/registration/allocations_spec.rb
fosterfarrell9: I moved the preselection test to Playwright (incl. the "(current state)" label) in [395ee27](https://github.com/MaMpf-HD/mampf/pull/1182/commits/395ee27b18d6f37d1fbef56fafc90c505d98ac88). However, I kept the other request specs on purpose: each asserts the turbo-stream response shape via a deliberate absence. The "folds the summary" one proves no flash alert is prepended over the modal; the "no groups" one proves no modal is emitted at all. A request spec can prove that in a fraction of a second; in Playwright, asserting those negatives would be way slower and possibly flakier for nothing.

Splines (approved):
Nice addition!

PR #1191, opened 2026-07-17
Rework lecture registration badges in search

fosterfarrell9:
The "registration open" badge on the lecture search cards only looked at open campaigns. A lecture whose only way in was a self-enrollment group (self-materialization outside a campaign) showed no badge at all, so students couldn't tell they could still join.

The badge is now personalized around "can I still get in / am I in":
- green "registered" if the user already has a campaign registration or is a member of one of the lecture's groups
- blue "registration open" if they aren't in yet but still could be (open campaign or a group that accepts self-enrollment)
- nothing otherwise.

This also avoids the "always open" trap after a campaign: allocated students are members, so they get the green badge, not a permanent "open".

The per-lecture status is computed in a bounded number of queries (Rosters::SelfEnrollmentStatusQuery) so the search cards stay N+1-free.


app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,70 @@
...
+    # not locked, not full). Deliberately user-independent: the card consults
+    # this only for users who are not already rostered, so allow_self_add?'s
+    # "not already a member" clause is covered by that ordering.
+    def enrollable_lecture_ids
+      return Set.new if @lecture_ids.empty?
+
+      candidates.each_with_object(Set.new) do |rosterable, ids|
+        ids << lecture_id_for(rosterable) unless rosterable.locked? || rosterable.full?
+      end
+    end
```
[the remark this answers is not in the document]
fosterfarrell9: This is addressed in [1935ec6](https://github.com/MaMpf-HD/mampf/pull/1191/commits/1935ec66f1049751b0b36498ce0c355718c0f796).

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,70 @@
...
+      def candidates
+        [
+          Tutorial.where(lecture_id: @lecture_ids,
+                         self_materialization_mode: SELF_ADD_MODES),
+          Talk.where(lecture_id: @lecture_ids,
+                     self_materialization_mode: SELF_ADD_MODES),
+          Cohort.where(context_type: "Lecture", context_id: @lecture_ids,
+                       self_materialization_mode: SELF_ADD_MODES)
+        ].flat_map(&:to_a)
+      end
```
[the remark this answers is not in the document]
fosterfarrell9: This is addressed in [1935ec6](https://github.com/MaMpf-HD/mampf/pull/1191/commits/1935ec66f1049751b0b36498ce0c355718c0f796).

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
...
+  # cards and paginates on scroll, so per-card or per-group lookups would N+1.
+  class SelfEnrollmentStatusQuery
+    SELF_ADD_MODES = [:add_only, :add_and_remove].freeze
+
+    # Per rosterable type: how to scope it to a lecture page and which
+    # membership association counts towards capacity.
+    ENROLLABLE_SOURCES = [
+      { klass: Tutorial, membership: :tutorial_memberships,
+        page_scope: ->(ids) { { lecture_id: ids } } },
+      { klass: Talk, membership: :speaker_talk_joins,
```
Splines: Rosterables have `roster_association_name` and Talk even overwrites it to `speaker_talk_joins`, so we should use that here to avoid this going stale should we rename it on the Talk. With this you can get rid of `membership` here entirely and just use something like `source[:klass].roster_association_name`.

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
...
+          # always sets, so this issues no query for these candidates.
+          next if rosterable.locked?
+          next if rosterable_full?(rosterable, counts[rosterable.id])
+
+          ids << lecture_id_for(rosterable)
+        end
+      end
+
+      # One grouped COUNT per type instead of rosterable.full?'s per-record
+      # count. Keep the capacity check in sync with Rosters::Rosterable#full?.
```
Splines: This comment to keep the capacity check in sync is not at the right place, I think it should be above the `rosterable_full` directly.

But even then: could you get rid of this duplication since we already have a `Rosterable#full?` method. For example, what about a `Rosterable.full_for_count?(rosterable, count)` method (in a helper?) that you call from both `#full?` as well as 'rosterable_full?` here.

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
+module Rosters
+  # For the lecture search cards: given a page of lectures, tells which the
+  # user is already rostered into and which still have a group anyone could
+  # self-enroll into. Answered in a fixed, small number of queries regardless
+  # of how many lectures or groups the page holds — the search renders many
+  # cards and paginates on scroll, so per-card or per-group lookups would N+1.
+  class SelfEnrollmentStatusQuery
+    SELF_ADD_MODES = [:add_only, :add_and_remove].freeze
```
Splines: This logic is already implemented in `Rosterable#config_allow_self_add?`.

app/models/rosters/self_enrollment_status_query.rb
Splines: since I just saw it, later you could use something like `candidates.first.roster_association_name` for the counts to replace the `source[:membership]` there

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      def talk_member_lecture_ids
+        Talk.where(lecture_id: @lecture_ids)
+            .joins(:speaker_talk_joins)
+            .where(speaker_talk_joins: { speaker_id: @user.id })
+            .distinct.pluck(:lecture_id)
+      end
+
+      def cohort_member_lecture_ids
+        Cohort.where(context_type: "Lecture", context_id: @lecture_ids)
```
Splines: The shape of context_type and context_id is already defined in `ENROLLABLE_SOURCES`, so might be worth to extract a helper like `cohort_lecture_scope(ids)`.  But maybe not worth it.

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
...
+                      .group(source[:klass].arel_table[:id])
+                      .count
+      end
+
+      def rosterable_full?(rosterable, member_count)
+        rosterable.capacity.present? && (member_count || 0) >= rosterable.capacity
+      end
+
+      def lecture_id_for(rosterable)
+        rosterable.is_a?(Cohort) ? rosterable.context_id : rosterable.lecture_id
```
Splines: So here we have again a special treatment for the Cohort. What about putting this in another lambda like `lecture_id` into the `ENROLLABLE_SOURCES`. Doing so, we could keep this special behavior defined in that array.

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      Set.new(tutorial_member_lecture_ids)
+         .merge(talk_member_lecture_ids)
+         .merge(cohort_member_lecture_ids)
+    end
+
+    # Lectures with a group that still accepts self-enrollment (mode allows it,
+    # not locked, not full). Deliberately user-independent: the card consults
+    # this only for users who are not already rostered, so allow_self_add?'s
+    # "not already a member" clause is covered by that ordering.
```
Splines: I didn't find this comment self-explanatory. Please also write about the caller and that the caller has to ensure the user is already a member of this group (see _lecture.html.erb where `enrollable_lecture_ids` is only invoked for the `elsif` block). Maybe that's what you meant by "ordering"? Not sure what "ordering" refers to otherwise.

app/models/rosters/self_enrollment_status_query.rb
```diff
@@ -0,0 +1,103 @@
...
+                     .where(source[:page_scope].call(@lecture_ids))
+                     .where(self_materialization_mode: SELF_ADD_MODES)
+                     .to_a
+        return if candidates.empty?
+
+        counts = member_counts(source, candidates.map(&:id))
+
+        candidates.each do |rosterable|
+          # locked? short-circuits on skip_campaigns, which self-materialization
+          # always sets, so this issues no query for these candidates.
```
Splines: This is never enforced on the DB-level, instead only in `Rosterable#locked?`. And the hope that every `add_only` or `add_and_remove` candidate also has `skip_campaigns: true` is only guaranteed by the  `enforce_consistency_between_modes`. Maybe it's worth it to add some form of constraint for at the DB-level. But rather something for a new PR I guess.

Splines (changes requested):
Works nicely! Here are some comments on the code and its abstractions. Maybe some of my suggestions turn out to actually over-complicate the logic, let's see.

app/models/rosters/self_enrollment_status_query.rb
fosterfarrell9: I changed the wording in [98f7a6f](https://github.com/MaMpf-HD/mampf/pull/1191/commits/98f7a6f33c69f1646527495794e80403e5e0cf5a).

app/models/rosters/self_enrollment_status_query.rb
fosterfarrell9: I took it one step further than the lambda in [9328e5b](https://github.com/MaMpf-HD/mampf/pull/1191/commits/9328e5baf75adadcb0158c5701a51aa66c195e0d): `Cohort` now has a `#lecture_id` next to its existing `#lecture`, so all three rosterables answer the same question and `lecture_id_for` is gone rather than moved into `ENROLLABLE_SOURCES`:

```ruby
# Like #lecture, but without loading the context — callers that only need the
# id are often iterating over many cohorts.
def lecture_id
  context_id if context_type == "Lecture"
end
```

(`context_type` rather than `context.is_a?(Lecture)` on purpose: `#lecture` loads the context, which would be a query per cohort in exactly the loop this class exists to avoid.) While at it: the very same switch had a second appearance in `Roster::MaintenanceController#set_rosterable`, where `respond_to?(:lecture_id)` was followed by an `is_a?(Cohort)` fallback. That fallback is now dead and the whole thing collapses to `@rosterable.is_a?(Lecture) ? id : lecture_id`. Technically, this is out of scope for this PR, but I still would argue we keep it in here.

app/models/rosters/self_enrollment_status_query.rb
fosterfarrell9: Went with your `candidates.first` variant from the follow-up comment in [5b07b9d](https://github.com/MaMpf-HD/mampf/pull/1191/commits/5b07b9d7789c010fe46d9e2de09ace26c322625a) — the class form can't work, since `roster_association_name` is defined inside included do and is therefore an instance method.

app/models/rosters/self_enrollment_status_query.rb
fosterfarrell9: Indeed, that the knowledge was duplicated, but `config_allow_self_add?` itself can't be reused here: it's a predicate on a loaded record, and this call site needs a value list for the `WHERE`. So instead of one call site I gave it one source of truth in [f33f18b](https://github.com/MaMpf-HD/mampf/pull/1191/commits/f33f18bba1920fba094f953a274b113fa262ea6d) — the constant now lives in `Rosterable` next to `SELF_MATERIALIZATION_MODES`, and both consumers derive from it:

```ruby
# The modes a student may add themselves in, shared by #config_allow_self_add?
# and the self_addable scope so the record and the SQL cannot drift apart.
SELF_ADD_MODES = [:add_only, :add_and_remove].freeze

scope :self_addable, -> { where(self_materialization_mode: SELF_ADD_MODES) }

def config_allow_self_add?
  SELF_ADD_MODES.include?(self_materialization_mode.to_sym)
end
```

The query now reads `source[:klass].self_addable.where(page_scope)` and carries no mode list of its own.
To keep it symmetric, I made the same change to the remove modes.

app/models/rosters/self_enrollment_status_query.rb
fosterfarrell9: Both parts done in a7b04b3, and the second removes the need for the first. I made it an instance method rather than your `Rosterable.full_for_count?(rosterable, count)` — as a class method it would have to re-fetch capacity off the passed record anyway, and full? can then simply delegate. No helper needed.

app/models/rosters/self_enrollment_status_query.rb
fosterfarrell9: You flagged it as maybe-not-worth-it, and the narrow version possibly isn't — but it pointed at something that was (at least in my opinion): `ENROLLABLE_SOURCES` only existed to hold per-type knowledge the models can hold themselves. So instead of a `cohort_lecture_scope` helper next to the table, the knowledge moved into the models and the table is gone:

```ruby
# Rosters::Rosterable
# How this rosterable hangs off a lecture. Cohort and Lecture override it.
scope :for_lectures, ->(lectures) { where(lecture_id: lectures) }

# Cohort
scope :for_lectures, lambda { |lectures|
  where(context_type: "Lecture", context_id: lectures)
}

# Lecture — a lecture is its own rosterable
scope :for_lectures, ->(lectures) { where(id: lectures) }
```

The query is now down to `ENROLLABLE_TYPES = [Tutorial, Talk, Cohort]` plus `klass.for_lectures(@lecture_ids).self_addable.to_a`, and `cohort_member_lecture_ids` uses the same scope — so the `context_type/context_id` shape appears exactly once.

The Lecture override is not decoration: without it, Lecture would inherit the default and query a `lecture_id` column it does not have. (Nothing calls it today, but might have been a future trap, so I added this).

Done in  [d503b2a](https://github.com/MaMpf-HD/mampf/pull/1191/commits/d503b2a3d494bb3e7bb7e5ad1615c89185771a5e) and [856d7f2](https://github.com/MaMpf-HD/mampf/pull/1191/commits/856d7f2fef6947c766184c5ed22120682cd1062b)

Splines (approved):
Nice, thanks for all the fixes. I approve now. Will run the tests before the new release.

PR #1192, opened 2026-07-17
Fix development quick login not working in GitHub Codespaces

Splines:
The "Login as teacher" in development was not working on codespaces as we didn't include all available loopback hosts to  `LOCAL_DEV_HOSTS`.


spec/controllers/dev/base_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    skip_before_action :authenticate_user!
+
+    def index
+      head :ok
+    end
+  end
+
+  before do
+    routes.draw do
+      get "index" => "dev/base#index"
```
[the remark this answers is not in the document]
Splines: fixed in 7bf8084761dbd3f18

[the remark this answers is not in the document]
Splines: No, this is actually wrong, we have to go to the correct route, `anonymous#index` won't work. Reverted in fec1121c03.

spec/controllers/dev/base_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      allow(Rails.env).to receive(:development?).and_return(true)
+    end
+
+    it "allows localhost loopback requests via 127.0.0.1" do
+      request.host = "127.0.0.1"
+
+      get :index
+
+      expect(response).to have_http_status(:ok)
+    end
```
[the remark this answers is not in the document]
Splines: good idea, done in 5591f4a58, which also fixes an issue with ipv6 addresses that are bracketed, i.e. `[::1]` instead of `::1`.

spec/controllers/dev/base_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  before do
+    routes.draw do
+      get "index" => "dev/base#index"
+    end
+  end
+
+  describe "#verify_development_environment" do
+    before do
+      allow(Rails.env).to receive(:development?).and_return(true)
+    end
```
fosterfarrell9: Because you stub this with true, there is no test that that it rejects outside of development. Since this is security relevant I would suggest to add a small test for this.

app/controllers/dev/base_controller.rb
```diff
@@ -1,13 +1,17 @@
...
 
     private
 
       def verify_development_environment
         # Don't use request.local? here since Docker uses 172. as IP range
         # for localhost, which is not considered local by Rails.
-        allowed = Rails.env.development? && request.host == "localhost"
+        normalized_host = request.host.delete_prefix("[").delete_suffix("]")
+        allowed = Rails.env.development? &&
+                  normalized_host.in?(LOCAL_DEV_HOSTS)
```
fosterfarrell9, line 14: I guess this will collide with the changes in #1188 (just as a note for the merge).

fosterfarrell9 (changes requested):

app/controllers/dev/base_controller.rb
Splines: ah true, thanks will take care of that in #1188 then

spec/controllers/dev/base_controller_spec.rb
Splines: Good point, added in 56c9ed42599

fosterfarrell9 (approved):
LGTM now

PR #1197, opened 2026-07-21
Rework video track ready check

fosterfarrell9:
This is an attempt to fix a flaky playwright test for thyme player by changing behaviour at the (possible) root of the issue.


KhyFee (commented):
This is a solid root-cause flake fix.

Tying readiness to the **track element’s** `load`/`error` + `readyState`, with a one-shot `notified` guard, correctly handles the “video metadata/canplay already fired while cues are still parsing” race — especially under test when the media is buffered.

**Looks good**
- Same pattern applied consistently in `chapter_manager.js` and `metadata_manager.js`
- `readyState === 3` path calling `onLoad(false)` avoids hanging `onVideoDataReady` on error
- Immediate `notifyIfReady()` covers the already-loaded case

**Nit (non-blocking):** if track events can fire more than once in some browsers, the `notified` flag already protects you; worth a one-line comment that listeners are intentionally left attached for the player lifetime (or remove them after first notify if you care about GC on long-lived pages).

Happy to see this land — much better than adding Playwright waits around a hung sidebar.

Splines (approved):
> There was a flaky Playwright test in CI that is not related to this here, but probably to #1203. Feel free to squash-merge.


No, it is actually the bug that is fixed in #1216.

PR #1198, opened 2026-07-22
Improve mampf-infra setup within mampf devcontainer

fosterfarrell9:
After I had completely set up one of my PCs afresh and tried to recreate the devcontainers with the "mampf+mampf-infra" workspaces I ran into some troubles that made the whole process quite inconvenient. With this PR, setup should be improved: if there is a sibling `mampf-infra` folder to the `mampf` folder in the host, it is mounted, otherwise it is not.


.devcontainer/devcontainer.json
```diff
@@ -5,7 +5,22 @@
     "name": "MaMpf",
     "dockerComposeFile": [
         "../docker/development/compose.yml",
-        "compose.devcontainer.yml"
+        "compose.devcontainer.yml",
+        "compose.mampf-infra.yml"
+    ],
+    "initializeCommand": [
```
Splines, line 17: You are duplicating the key here: `Duplicate object key. jsonc (520)`

Image

Splines (changes requested):

.devcontainer/devcontainer.json
```diff
@@ -1,32 +1,46 @@
...
+        "docker",
+        "run",
+        "--rm",
+        "--mount",
+        "type=bind,source=${localWorkspaceFolder},target=/workspace",
+        "--mount",
+        "type=bind,source=${localWorkspaceFolder}/..,target=/workspace-parent,readonly",
+        "alpine:3.22",
```
Splines, line 25: Does this mean, we have to download another image, just to execute this one command? Arguably, alpine is just 5MB in size, still wondering.

Splines (changes requested):

.devcontainer/devcontainer.json
fosterfarrell9: This is fixed in [78d07bb](https://github.com/MaMpf-HD/mampf/pull/1198/commits/78d07bbdb44ee784b025206eebaa06707a308324).

.devcontainer/devcontainer.json
fosterfarrell9: Yes, one extra image — but it's about host-shell independence rather than convenience. `initializeCommand` runs on the contributor's own machine, and we can't assume what that is: the same command needs different flags on Linux vs. macOS, and a pure-Windows host has no sh at all. Running the generator in Alpine means the script only has to be correct against one shell. Alpine is 3.8 MB compressed and pulled once.

Splines (approved):
Nice addition!

PR #1203, opened 2026-07-29
Rework display action for manuscripts for iOS compatibility

### LINKED ISSUE #1202: ⁠[Bug] PDFs fail to load on iOS (only thumbnail displayed)⁠
### Problem
When opening a PDF in the web version of MaMpf on iOS, the document fails to load. Instead of rendering the full PDF, only the PDF thumbnail is displayed.
### Steps to Reproduce
1. Open the MaMpf web version on an iOS device (Safari/Chrome).
2. Navigate to a course/lecture containing a PDF resource.
3. Open the PDF.
### Expected Behavior
The full PDF document should render completely and be scrollable/readable.
### Actual Behavior
Only the thumbnail preview of the PDF is displayed.
### Environment
 OS: iOS, iPadOS 
 Platform: Web version, Chrome, Safari…

fosterfarrell9:
This fixes #1202. We need to test in staging whether this works in a production setting with iOS.


Splines (approved):
What?! We were displaying PDFs in an iFrame? Glad this is removed now ^^ Locally, I can still access the PDFs, let's see if it works in staging.

PR #1206, opened 2026-08-02
Reorganize Agents files

Splines:
Previously, we had a `.github/copilot-instructions.md` file. The more standardized approach is however an `AGENTS.md` file that GitHub Copilot [also supports](https://github.blog/changelog/2025-08-28-copilot-coding-agent-now-supports-agents-md-custom-instructions/). Unfortunately, Claude [does not support it](https://code.claude.com/docs/en/memory#agents-md), but one can create a `CLAUDE.md` file and simply use `@AGENTS.md` in it as done here.


fosterfarrell9 (approved):
LGTM

fosterfarrell9 (approved):
LGTM

PR #1208, opened 2026-08-03
Herb-Check every ERB file, not just the ones before the first exclusion

fosterfarrell9:
herb-lint and herb-format stop at the first file excluded by `.herb.yml` and still exit 0, so one excluded file in the changed-files list silently skipped the check for every other file. Four of the five Müsli slice PRs carried unformatted ERB with a green check because of it.

Passing the files one at a time keeps the exclusion list in `.herb.yml`, where it belongs, and costs a few seconds.


Splines (approved):
Ups, thanks for fixing!

PR #1216, opened 2026-08-09
Rasterize manuscript previews with mutool

fosterfarrell9:
The security update in #1201 blocked libvips' poppler-backed PDF loader, which broke screenshot extraction on every manuscript upload. Instead of re-enabling that loader (#1215), this draws the first page with `mutool` and leaves the block alone — poppler no longer renders anything, so CVE-2025-59933 is out of our path rather than accepted. It also runs as a child process with a timeout, so a hang no longer costs a web worker.

Supersedes #1215. Everyone needs to rebuild their dev container for the new `mupdf-tools` package, I guess.


docker/development/Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN which yarn; yarn --version
 # Install other dependencies
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get install -y --no-install-recommends \
-        ffmpeg qpdf shared-mime-info \
+        ffmpeg qpdf mupdf-tools shared-mime-info \
```
Splines, line 45: Before I review the code, note that MuPDF is AGPL licensed, which is kind of restricting, see [their licensing page](https://artifex.com/licensing) and also these two points which are currently not yet met with the current state of the PR:

> When using our open-source under AGPL, you ideally should mention our open-source and include AGPL license details in the PDF metadata. You ideally should also retain the producer line in all PDFs created or modified using our open-source.

> Any modifications to our open-source must be disclosed under the AGPL, including sharing the modified code with any users interacting with it.

That is, we would have to make sure these AGPL license details are included in the PDF metadata, and that users can download the source-code of mupdf and our MaMpf code (or at least link to it).

But to be fair, I just found out that poppler also has a GPL license, see their repo [on Gitlab](https://gitlab.freedesktop.org/poppler/poppler#history-and-gpl-licensing):

> Please note that xpdf, and thus poppler, is licensed under the GPL, not the LGPL. Consequently, any application using poppler must also be licensed under the GPL. If you want to incorporate Xpdf based PDF rendering in a closed source product, please contact Glyph & Cog (www.glyphandcog.com) for commercial licensing options. Note that this only allows you to use xpdf in a closed source product, not poppler itself.Please note that xpdf, and thus poppler, is licensed under the GPL,
not the LGPL.  Consequently, any application using poppler must also
be licensed under the GPL.  If you want to incorporate Xpdf based PDF
rendering in a closed source product, please contact Glyph & Cog
([www.glyphandcog.com](http://www.glyphandcog.com/)) for commercial licensing options. Note that
this only allows you to use xpdf in a closed source product,
not poppler itself.

---

So, as I can see from #1215, the problem seems to be that Debian Trixie doesn't fix [CVE-2025-59933](https://security-tracker.debian.org/tracker/CVE-2025-59933) for the vips package, right? It was marked as minor issue there. See also on [GitHub](https://github.com/libvips/libvips/security/advisories/GHSA-q8px-4w5q-c2r4):

> Processing of such an image halts very shortly after the buffer read overflow, and no output is generated. It is not possible for a consuming application to access the memory area overflowed.

But of course, it is still a vulnerability that we'd like to get rid of.

---

Furthermore, you might want to look into [ActiveStorage Previewers](https://api.rubyonrails.org/classes/ActiveStorage/Preview.html) instead of spawning own processes. For example, here is the source code for the [`PopplerPDFPreviewer`](https://github.com/rails/rails/blob/af448cd8f1d33d05262c7223aded58ba0f351aff/activestorage/lib/active_storage/previewer/poppler_pdf_previewer.rb#L1-L39), which simply uses `pdftoppm` under the hood, a tool that comes with poppler. Rails also comes with a [`MuPDFPreviewer`](https://github.com/rails/rails/blob/af448cd8f1d33d05262c7223aded58ba0f351aff/activestorage/lib/active_storage/previewer/mupdf_previewer.rb#L1-L40).

app/uploaders/pdf_uploader.rb
```diff
@@ -82,10 +82,15 @@ class PdfUploader < Shrine
...
   # extract a screenshot from pdf and store it beside the pdf
   Attacher.derivatives_processor do |original|
-    screenshot = ImageProcessing::Vips.source(original).loader(page: 0, dpi: 150)
-                                      .convert("png")
-                                      .resize_to_limit!(400, 565)
-    { screenshot: screenshot }
+    page = ManuscriptPageRenderer.render_first_page(original.path)
+    begin
+      screenshot = ImageProcessing::Vips.source(page.path)
+                                        .convert("png")
```
Splines, line 90: I was confused about this `.convert("png")` here, which actually doesn't convert the PDF to PNG (this is done already in the `ManuscriptPageRenderer`), but just marks this as PNG, such that the format is not `nil` (which wouldn't be a problem, as Vips would call `determine_format(source_path)` then).

So maybe add a small comment here explaining this.

app/uploaders/pdf_uploader.rb
```diff
@@ -82,10 +82,15 @@ class PdfUploader < Shrine
...
+                                        .resize_to_limit!(400, 565)
```
Splines: Would love to increase this as these thumbnails always look kind of blurry in MaMpf. I think the slightly increased data we have to send is worth it.

Splines (changes requested):

app/uploaders/pdf_uploader.rb
fosterfarrell9: I added a comment i n [76bddd9](https://github.com/MaMpf-HD/mampf/pull/1216/commits/76bddd952d75cf9493aa068ab7d2d6d52f880a2c).

app/uploaders/pdf_uploader.rb
fosterfarrell9: I increased it in [76bddd9](https://github.com/MaMpf-HD/mampf/pull/1216/commits/76bddd952d75cf9493aa068ab7d2d6d52f880a2c).

docker/development/Dockerfile
fosterfarrell9: The AGPL points are about PDFs that MuPDF creates or modifies ore code changes. We do neither — mutool draw -F png reads a PDF and writes a PNG — and we use Debian's mupdf-tools unmodified. Worth noting too: libvips links poppler into our own process, while mutool is a separate program, so this is the looser of the two couplings. I added MuPDF to the footer's "Powered by" line, linking to mupdf.com/releases, which covers both the mention and the source pointer.

Regarding ActiveStoragePreviewers: Manuscripts are Shrine attachments, and a previewer needs an ActiveStorage::Blob — we would have to move them to ActiveStorage first. And in their implementation, MuPDFPreviewer runs the same binary (mutool draw -F png -o - <file> 1), just without -r, -w/-h, rlimits or a timeout. Those limits in our PR are dedicated hardening - see the spec about a 458-byte PDF declaring a 14400pt page then rendering at full size - that are missing in the ActiveStorage solution.

PR #1157, opened 2026-06-21
Add banners for participants and groups in roster views

fosterfarrell9:
In this PR we add two banners, one in the groups tab and one in the participants tab of the lecture edit view, whose purpose it is to inform the users of the ongoing integration of Müsli in MaMpf and the structural changes in the lecture edit view page. In particular they explain why for the current semester the studnet counts are all 0.

Screenshot 2026-06-21 185651
Screenshot 2026-06-21 185701

PR #1222, opened 2026-08-18
Remove Müsli transition banners

Splines:
Follow-up to #1220.

#1157 introduced "transition banners" to explain some quirks during the integration of Müsli related to old lectures. But upcoming lectures will all use our newly established roster-concepts, so there is no need for this banner anymore. This PR thus removes these banners.

### Off-topic

Also fixes a flaky unit test. The reason was that we didn't explicitly add a `short_title` to the course. And `Course#search_by_title` matches against `title` or `short_title` (using [`pg_trgm`](https://www.postgresql.org/docs/current/pgtrgm.html), where the latter was sometimes random as we didn't set it.


fosterfarrell9 (dismissed):
LGTM.

Issue #1224, opened 2026-08-18
Sass-related warnings


Splines:
We get a lot of Sass-related warnings during startup due to [this Bootstrap issue](https://github.com/twbs/bootstrap/issues/40962). However, note [this comment from a team member](https://github.com/twbs/bootstrap/issues/40962#issuecomment-3730409408):

> Bootstrap 5 was built years ago when both Node Sass and Dart Sass were supported. Node Sass has since been deprecated, which means it's no longer being developed. Updating Bootstrap 5 to use the new Dart Sass syntax and features would break backward compatibility for anyone still using Node Sass—plus it just violates SemVer rules.

> As such, we've made the decision to not touch v5 in this regard while we work on v6's development. v6 uses Dart Sass only, has been transitioned to Sass modules (because imports have also been deprecated), and has a ton more changes to modernize the codebase.

> Silencing the deprecations is your best bet on v5 at this time.

So in Bootstrap 6 (in the future), this should be fixed. Until now, we should just silence those warnings.

PR #1225, opened 2026-08-18
Silence Sass if functions

Splines:
This addresses #1224 (but does not fix it).

See also
- https://sass-lang.com/documentation/breaking-changes/if-function/
- https://github.com/twbs/bootstrap/issues/40962#issuecomment-3730409408

### DIFF
--- a/vite.config.mjs
+++ b/vite.config.mjs
@@ -43,7 +43,7 @@ export default defineConfig({
           "mixed-decls",
           "color-functions",
           "global-builtin",
-          "if-function"
+          "if-function",
         ],
       },
     },
--- a/vite.config.mjs
+++ b/vite.config.mjs
@@ -40,7 +40,6 @@ export default defineConfig({
         silenceDeprecations: [
           "legacy-js-api",
           "import",
-          "mixed-decls",
           "color-functions",
           "global-builtin",
           "if-function",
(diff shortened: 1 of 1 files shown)


vite.config.mjs
```diff
@@ -43,6 +43,7 @@ export default defineConfig({
           "mixed-decls",
           "color-functions",
           "global-builtin",
+          "if-function"
```
fosterfarrell9: nitpick: I suppose there should be `,` at the end (which is bizarre IMO, but seems to be the rule)

vite.config.mjs
```diff
@@ -43,6 +43,7 @@ export default defineConfig({
           "mixed-decls",
```
fosterfarrell9: From my understanding, `mixed-decls` has become obsolete with Dart Sass 1.92. So we shlud be able to remove it. But better check this.

fosterfarrell9 (changes requested):
Only minor nitpicks. You may resolve the as well without addressing if you want.

vite.config.mjs
Splines: True, thanks. This is a rule to improve quick editing when you can just press Enter without having to type a comma beforehand. But as always, this is personal taste. Unfortunately, this file is not linted...

Splines: fixed in f494253ce2692d

vite.config.mjs
Splines: Ah, I didn't know that, thanks. See ["The New Way"](https://sass-lang.com/documentation/breaking-changes/mixed-decls/#the-new-way) in the docs. I've removed it in ec8c8bbe4eb.

PR #1228, opened 2026-08-18
Unify searches on one page

Splines:
This is a first step towards the new dashboard. Here, I repurpose the already existing `app/frontend/search/index.html.erb` to hold all searches:

- Tag Search. Instead of having the search bar in the navbar that is _just_ for the tag search, I put that searchbar on the index page itself and replaced the searchbar in the navbar by a simple search icon. I also improved the tag search such that it directly returns the results as you type (via Hotwire).
- Lecture Search & Media Search. Those are just moved 1:1 to the new place.
- The new "fancy" lecture search we have on the `/main/start` page is also referenced here, such that all searches are now in one place and the front page is cleaner.

While all searches now live in one place, we surely want to improve on this later on, e.g. by putting back an all-encompassing searchbar directly to the navbar. But for now, I think this is a quite pragmatic solution (to get the space required on the main page where the new dashboard will sit).

image


Splines:
@coderabbitai review

app/frontend/search/index.html.erb
```diff
@@ -1,20 +1,119 @@
...
+                <%= render partial: 'search/tag_search_results',
+                           locals: { tags: @tags,
+                                     filtered_tags: @filtered_tags,
+                                     search_string: @search_string } %>
+              <% end %>
+            <% end %>
+            <div class="tag-search-spinner">
+              <div class="spinner-border spinner-border-sm text-secondary" role="status">
+                <span class="visually-hidden">Loading...</span>
+              </div>
```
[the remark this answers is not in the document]
Splines: fixed in 6070cfb56a8

app/views/shared/_navbar.html.erb
```diff
@@ -110,15 +110,15 @@
...
-      <% end %>
+      <ul class="navbar-nav">
+        <li class="nav-item">
+          <a href="<%= search_index_path %>"
+            class="nav-link bi bi-search <%= get_class_for_path(search_index_path()) %>"
+            data-bs-toggle="tooltip"
+            title="<%= t('navbar.search') %>">
+          </a>
+        </li>
+      </ul>
```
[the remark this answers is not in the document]
Splines: fixed in 6070cfb56a8

app/controllers/search_controller.rb
```diff
@@ -2,44 +2,30 @@
...
-  end
+  layout "application_no_sidebar"
 
   def current_ability
     @current_ability ||= SearchAbility.new(current_user)
   end
 
   def index
-    return unless set_search_string
+    @search_string = params[:search]
```
fosterfarrell9: `params[:search]` is treated as a scalar here, but a bracketed query parameter delivers `ActionController::Parameters`, and `.length` on line 15 raises `NoMethodError`. Signed in: `http://localhost:3000/search/index?search[x]=ax` → 500.

An array does not crash but puts the raw value on the page: `?search[]=ax&search[]=ay` renders "Your search for the tag ["ax", "ay"] has returned 0 results." Use a term with no hits to see it — search.info sits in the else of _tag_search_results.html.erb:10, so as soon as anything matches you get badges and no text at all.

On next the `rescue_from ActionController::ParameterMissing` at lines 6-8 caught the first case and redirected with controllers.no_search_term. It went away together with `params.expect(:search)`, and nothing replaced it.

Maybe use something like
```ruby
    @search_string = params.permit(:search)[:search]
```
permit returns nil for both a hash and an array, so malformed input behaves like an empty search instead of crashing.

app/frontend/search/index.html.erb
```diff
@@ -1,20 +1,119 @@
...
+              id="lectureSearchLink">
+            <%= t('basics.lecture_search') %>
+          </h4>
+        </button>
+      </div>
+      <div class="collapse subscriptionsCollapse" id="collapseLectureSearch"
+           data-bs-parent="#subscriptionsAccordion"
+           data-link="#lectureSearchLink">
+        <div class="accordion-body">
+          <%= render partial: 'search/lecture_search_form',
```
fosterfarrell9, line 71: Clicking **+** on a result of this search does nothing whenever the subscription needs a dialog — no dialog, no error, no message.

To reproduce: give a released lecture a passphrase (lecture → Edit → Preferences → *Passphrase*), sign in as someone not subscribed to it, search for it here and click **+**. On `next` the same click — start page, collapsed **Lecture Search** accordion — opens the passphrase dialog.

The cards this search renders go through `profile/subscribe_lecture.coffee`. On success it replaces the card and calls `$('#subscriptionModal').modal('hide')`; when it needs input — a passphrase, or a lecture the user is already subscribed to — it calls `.modal('show')` and fills `#subscription-modal-content`. Both elements come from `profile/_subscription_modal`, which stayed behind on the start page (`main/start.html.erb:160`, still needed there by the term lecture lists) when this search moved here, so the calls land on an empty jQuery set. That is also why the ordinary case looks fine: `modal('hide')` on nothing is just as silent as `modal('show')`.

Rendering `profile/subscription_modal` on this page fixes it — at top level, next to line 118, not inside the accordion body, since that is a `.collapse` and a modal in a hidden parent stays hidden.

app/views/shared/_navbar.html.erb
```diff
@@ -110,15 +110,16 @@
...
-      <%= form_tag(search_index_path,
-                  method: 'get',
-                  class: 'form-inline mt-2 mt-md-0',
-                  id: 'searchField') do %>
-        <%= search_field_tag 'search',
-                            nil,
-                            class: 'form-control me-sm-2',
-                            placeholder: t('navbar.search') %>
-      <% end %>
+      <ul class="navbar-nav">
```
fosterfarrell9: The form this replaced carried `id="searchField"`, which `lectures/lectures.js:158` and `:181` use to move the search field between the two navbar rows on narrow screens. `#searchField` no longer exists in the rendered page, so both lines now operate on an empty set.

I could not find a page where that is visible: `lectures.js` is only loaded by `lectures/show` and `lectures/edit`, and on the edit page the navbar is the administration one, so all four `appendTo($("#secondnav"))` calls there already target nothing. So this looks like a loose end rather than a regression — either give the `<ul>` the id back, or drop the two lines.

app/frontend/search/_tag_search_results.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<% else %>
+  <%= t('search.info',
+        search_string: search_string,
+        count: filtered_tags.count) %>
+  <% if tags.count != filtered_tags.count %>
+    <%= t('search.omitted_html',
+          profile: link_to(t('search.settings'), edit_profile_path),
+          count: tags.count - filtered_tags.count) %>
+  <% end %>
+<% end %>
```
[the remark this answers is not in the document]
Splines: fixed in 7e687f7540e6b

app/frontend/search/index.html.erb
```diff
@@ -1,20 +1,119 @@
...
+                        class: "mb-3",
+                        id: "tagSearchForm",
+                        data: { turbo_frame: "tag-search-results",
+                                auto_submit_form_target: "form" } do %>
+            <%= search_field_tag :search,
+                                @search_string,
+                                class: "form-control",
+                                placeholder: t('navbar.search'),
+                                data: { action: "input->auto-submit-form#debouncedSubmit",
+                                        auto_submit_form_target: "searchInput" } %>
```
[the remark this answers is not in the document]
Splines: fixed in 4343c0f7e923

app/controllers/search_controller.rb
Splines: Good point, fixed in 4343c0f7e9. I even went as far as only taking care about the `search` param if the request is a turbo frame request. See aa6a77b9034.

app/frontend/search/_tag_search_results.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if filtered_tags.present? %>
+  <% filtered_tags.natural_sort_by(&:title).each do |t| %>
+    <%= link_to t.short_title,
+                tag_path(t),
+                class: "badge text-dark lecture-tag",
+                data: { toggle: 'tooltip',
+                        placement: 'top' },
+                title: t.title %>
+  <% end %>
+<% else %>
```
fosterfarrell9, line 10: There is one problem here that can be reproduced as follows: In the user your profile, set *I want to be shown content* to **only from the courses I have subscribed to**, subscribe to "(PS) Elementare Zahlentheorie, SS 2025", then run a tag search for `Prim`.

That matches four tags — Primzahl (Elementare Zahlentheorie), plus Primelement, Primideal and Skalarprodukt (Lineare Algebra 2) — so one is visible and three are filtered out. On `next` you get the badge *and* "3 additional results … are not shown because of your profile settings". On this branch you get the badge only.

Counter-check: search `Ring`, where all six hits are in Lineare Algebra 2 and none is visible. Then the sentence appears on both branches, which is what pins this to the change rather than to the data.

The cause is the ordering: on `next` (`search/index.html.erb:1-8`) `search.info` rendered unconditionally and `search.omitted_html` whenever `tags.count != filtered_tags.count`, both above the badges. Here they moved into the `else`, so the message that explains a surprising result set only shows when there is no result set to explain.

Inside this branch `filtered_tags.present?` is false by construction, so `count: filtered_tags.count` is always `0` and the condition below reduces to `tags.count != 0`.

app/frontend/search/index.html.erb
Splines: Oh thanks for the repro and the fix, done in 2dc300e7f.

app/controllers/search_controller.rb
```diff
@@ -2,44 +2,30 @@
 class SearchController < ApplicationController
   authorize_resource class: false
   before_action :check_for_consent
-
-  rescue_from ActionController::ParameterMissing do |_exception|
-    redirect_back_or_to root_path, alert: I18n.t("controllers.no_search_term")
-  end
+  layout "application_no_sidebar"
```
fosterfarrell9, line 5: `Turbo::Frames::FrameRequest` sets `layout -> { "turbo_rails/frame" if turbo_frame_request? }`, but a `layout` declared in the subclass wins over it, so frame requests from the tag search render the full page.

Measured for `?search=Prim`: the response is 47,100 bytes and contains `<html>` and `id="firstnav"`, while the frame uses 198 bytes of it. So every debounced keystroke renders the navbar with its notification queries, the lecture dropdown, all four search forms and the lecture search below — and Turbo throws all of it away.

`layout -> { "turbo_rails/frame" if turbo_frame_request? }` here keeps the full layout for the normal page load and drops it for the frame.

app/controllers/search_controller.rb
```diff
@@ -2,44 +2,30 @@
...
   end
 
   def index
-    return unless set_search_string
+    @search_string = params[:search]
+    return if @search_string.blank?
 
-    @tags = Tag.search_by_title(@search_string)
+    if @search_string.length > 1
+      @tags = Tag.search_by_title(@search_string)
```
fosterfarrell9: We should keep this in mind (we have ~3000 Tags in production), optimizing queries is possibly out of scope for this PR.

`Tag.search_by_title` returns every match — the `pg_search_scope` in `tag.rb:66` ends at `ORDER BY rank DESC, tags.id ASC`, there is no `LIMIT` and the controller adds none. `User#filter_tags` (`user.rb:242-248`) then walks that whole set in Ruby: it calls `related_lectures` inside the `select` block — once per tag, rebuilding `Lecture.where(course: related_courses)` each time — and `Tag#in_lectures?` does `in?(lecture.tags)` per lecture.

Measured against the seeded development data as `teacher@mampf.edu` (`subscription_type=3`, 5 courses): `?search=ab` costs 65 queries, `?search=a` 35, `?search=Prim` 21 — roughly two queries per hit. Harmless with 48 tags in the database, but it scales with the result set, and the two-character minimum means the *first* request of every search is the broadest one. `er` alone already matches 23 tags here.

Not new code, but this PR is what makes it run per keystroke instead of per explicit submit. A `limit` on the tag query would bound it; moving the visibility filter into SQL would remove the per-tag cost altogether.

The cheapest bound in this PR is a limit on the tag query — `Tag.search_by_title(@search_string).limit(50)`. I cannot demonstrate its value here (the seeded database has 48 tags, so a limit of 50 never bites); it is there for the case this data cannot produce. One thing to keep in mind if you add it: `_tag_search_results` reports `tags.count - filtered_tags.count` as "hidden by your profile settings", and a limit would quietly fold truncated results into that number.

The larger win is in `filter_tags` itself, which is outside this diff: `related_lectures` is called inside the `select` block, so it rebuilds `Lecture.where(course: related_courses)` once per tag. Hoisting both out of the loop takes the same search from 92 queries to 48 with an identical result. Worth doing, but not here.

fosterfarrell9 (changes requested):

app/views/shared/_navbar.html.erb
Splines: I added back the missing id in 256ed87f5928 since it is referenced in `lectures.js` as you mentioned. Let's not worry about this too much. We have to redo the "two navs" logic we have anyways in the future to really make this pretty and accessible.

app/frontend/search/_tag_search_results.html.erb
Splines: I think I already fixed that when I addressed [this comment](https://github.com/MaMpf-HD/mampf/pull/1228#discussion_r3820938288), right?

image

app/controllers/search_controller.rb
Splines: Yeah indeed the whole layout was re-rendered. But I didn't fix it as suggested in 80ef6eb210c, instead I now only render the partial in the controller.

app/controllers/search_controller.rb
Splines: Yeah, good point, but let's keep this out of this PR for now. The endpoint now also sends just the response and not the full layout anymore, so at least a few less bytes. But of course, the query itself should be optimized in the future.

fosterfarrell9 (dismissed):

e2e/lecture-subscriptions.spec.ts
```diff
@@ -4,17 +4,26 @@ import { LecturePage } from "./page-objects/lecture_page";
...
     }
 
     await page.goto("/main/start");
+    await page.getByRole("link", { name: "Search", exact: true }).click();
     await page.getByRole("button", { name: "Lecture Search" }).click();
     await page.getByRole("button", { name: "Search", exact: true }).click();
-    await page.getByTitle("Subscribe").nth(0).click();
-    await expect(page.getByTitle("Unsubscribe").nth(0)).toBeVisible();
+
+    const lectureCard = page.locator(".lectureCard").first();
```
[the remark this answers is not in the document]
Splines: fixed in 7622a364c

PR #1229, opened 2026-08-18
Move announcements above navbar

Splines:
Beforehand, announcements were only visible on the `/main/start` page. To carve some space for the upcoming dashboard, we move these announcements _above_ the navbar and show it everywhere. I think admin-wide announcements deserve that spot as they are very important (and should be used sparingly).

image

image

### DIFF
--- a/app/controllers/main_controller.rb
+++ b/app/controllers/main_controller.rb
@@ -41,7 +41,6 @@ class MainController < ApplicationController
                                                                    :term)
                                        .sort
     end
-    announcements
     next_term_banner
     @talks = current_user.talks.includes(lecture: :term)
                          .select { |t| t.visible_for_user?(current_user) }
--- a/app/controllers/main_controller.rb
+++ b/app/controllers/main_controller.rb
@@ -7,11 +7,6 @@ class MainController < ApplicationController
     @current_ability ||= MainAbility.new(current_user)
   end
 
-  def home
-    cookies[:locale] = current_user.locale if user_signed_in?
-    announcements
-  end
-
   def error
     redirect_to :root, alert: I18n.t("controllers.no_page")
   end
@@ -56,12 +51,6 @@ class MainController < ApplicationController
       return unless user_signed_in?
 
       redirect_to consent_profile_path unless current_user.consents
-    end
-
-    def announcements
-      @announcements = Announcement.where(on_main_page: true, lecture: nil)
-                                   .pluck(:details)
-                                   .join('<hr class="my-3" w-100>')
     end
 
     # Transitional banner pointing to the lectures of the upcoming term
--- a/app/frontend/styles/navbar.scss
+++ b/app/frontend/styles/navbar.scss
@@ -62,10 +62,3 @@
         color: #ffe5ed;
     }
 }
-
-#site-announcement-banner {
-    margin-bottom: 0;
-    padding: 0.5rem 1rem;
-    text-align: center;
-    font-weight: 500;
-}
--- a/app/frontend/styles/navbar.scss
+++ b/app/frontend/styles/navbar.scss
@@ -63,9 +63,7 @@
     }
 }
 
-// Admin-level announcements, shown full-width above the navbar on every page.
-.site-announcement-banner {
-    border: 0;
+#site-announcement-banner {
     margin-bottom: 0;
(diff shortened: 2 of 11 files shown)


app/controllers/main_controller.rb
```diff
@@ -41,7 +41,6 @@ def start
                                                                    :term)
                                        .sort
     end
-    announcements
```
fosterfarrell9, line 44: With `announcements` gone from here, the only caller of this is `#home` action, and `home` doesn't even seem to have a route. So announcements can go, and home with it, actually. Maybe the latter is out of scope, but if you do it here, I wouldn't protest. And while you are cleaning up: ApplicationHelper.rb has a `#main_page_announcements` method which is dead code for a while now.

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,9 @@
 module ApplicationHelper
+  # Returns joined HTML of all admin-level announcements shown above the navbar.
+  def site_announcements_html
```
fosterfarrell9: There is already an `AnnouncementHelper`. Why not use this?

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,9 @@
...
+    Announcement.active_on_main.pluck(:details).join('<hr class="my-2">')
```
fosterfarrell9: Notethat active_on_main has no order, so the order is just whatever postgres decides in this moment. Maybe order by `created_at`?

Also: why not memoize here: `@site_announcements_html ||= Announcement.active_on ...`. Then in `navbar` partial you can just do

```
<% if site_announcements_html.present? %>
```

and save the local variable.

app/frontend/styles/navbar.scss
```diff
@@ -61,4 +61,11 @@
     &:hover {
         color: #ffe5ed;
     }
+}
+
+#site-announcement-banner {
```
fosterfarrell9: Bootstrap's `.alert` which you use this with brings a radius and border (so there are white wedges at all four corners — the bottom two right against the dark navbar.  The declarationsare plain utilities (py-2 px-3 is exactly the 0.5rem 1rem there), so the rule and the ID can go with them — that is also what we do elsewhere, e.g. something like

```html
<div class="alert alert-warning mb-0 rounded-0 border-0 text-center fw-medium py-2 px-3" role="alert">
```
then you could delete these.

fosterfarrell9 (changes requested):

app/views/shared/_navbar.html.erb
```diff
@@ -1,6 +1,13 @@
 <% I18n.with_locale(current_user.try(:locale)) do %>
 <%= vite_stylesheet_tag "styles/navbar.scss" %>
 
+<% announcements_html = site_announcements_html %>
+<% if announcements_html.present? %>
+  <div class="alert alert-warning" id="site-announcement-banner" role="alert">
```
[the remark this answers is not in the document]
Splines: Good catch, fixed in 03e9e49420b452

app/controllers/main_controller.rb
Splines: good point, done in aa26323bde8ae

app/helpers/application_helper.rb
Splines: Yeah, I missed that it existed already. Moved to that helper in a6580d8cd22.

app/helpers/application_helper.rb
Splines: Nice, good idea to memoize here. Done that and ordered `active_on_main` by `created_at` (directly in the scope, hope that's fine) in 13587236a2da.

app/frontend/styles/navbar.scss
Splines: Thanks I didn't notice it was rounded as the color was very bright^^ Fixed as you suggested with in-line Bootstrap styles in 3708d7966dab.

app/helpers/application_helper.rb
Splines: In my opinion, `created_at: desc` is better as this shows the newest announcements first, which I think makes sense. Changed that in ed87bf6a968ce.

fosterfarrell9 (approved):
LGTM now!

PR #1242, opened 2026-08-23
PR 1: Background Scheduler and Subtitles

f-buerckel:
# PR 1: Background Scheduler and Subtitles

## 1. Overview & Motivation

This pull request introduces the first part of the **MampfSearch** integration: the background synchronization scheduler, authentication layer, and subtitle rendering. It serves as a foundation for future PRs that will introduce the actual retrieval features. The corresponding PR in the Mampfsearch repository can be found [here](https://github.com/MaMpf-HD/mampf-ai-search/pull/1).

---

## 2. Security & Authentication Architecture

Communication between MaMpf and MampfSearch operates across two directions:

### 1. Direction A: MaMpf $\to$ MampfSearch
- Every HTTP call from MaMpf (via `SearchClient`) attaches a short-lived (~60s) HMAC token in the `Authorization: Bearer` header signed with the shared `MAMPFSEARCH_API_SECRET`. MampfSearch checks if the token is signed with the correct key, the payload matches the endpoint, and the token has not expired.
- **Payload:** `{ purpose: "mampfsearch_api", scope: request_path, expires_at: timestamp, nonce: hex }`.
- **Endpoint Scoping:** The token is strictly bound to the target path (e.g. `scope: "/lesson/search"`). A search token cannot be replayed against the ingestion endpoint.

### 2. Direction B: MampfSearch $\to$ MaMpf
When MampfSearch receives an ingestion request from MaMpf, the request contains two special URLs:

- The `video_stream_path` temporarily allows access to a video without user session authentication. It is used by the API to download the video even if it is private.
- The `transcript_upload_path` is the link to the Shrine uploader to upload the final `.vtt` transcript file to MaMpf.

Both paths contain an HMAC token (`TranscriptionToken`), signed using the private Rails secret (`secret_key_base`), as a URL parameter. The payload contains the purpose (to upload a transcript or to download a video), the exact medium ID for which those URLs are valid, and the TTL (how long the link is valid).

Additionally, both endpoints are also secured with the `SearchApiToken`, ensuring that only the MampfSearch API can reach those. Even if the URLs are exposed somewhere, no one can access the endpoints without a valid `SearchApiToken`.

> [!NOTE]
> The `TranscriptionToken` is not strictly necessary if we fully trust the API server, because the `SearchApiToken` already prevents any access to the endpoints from outside the API. However, if the API server were compromised, this token would limit access to MaMpf private media and prevent uploading arbitrary transcripts to other videos. Especially with the WireGuard setup, this level of security might not be strictly necessary, but it adds little code and provides an additional layer of security.

---

## 3. Database Synchronization

The MampfSearch database mirrors the lecture structure of the MaMpf database (Course -> Lectures -> Lessons -> Medias). The Mampf database now additionally stores which videos are already transcribed. Therefore, it is very important to ensure that both stay in sync. For this, two layers are used:

### 1. Ingestion Triggers
Basic synchronization is provided by automatically triggering an ingestion once a video is uploaded or changed. Similarly, if a medium or video is deleted, the `MampfsearchDeleteJob` is triggered to prune orphaned records.

### 2. The Scheduler (`MampfsearchSyncJob`)
However, this alone is not enough: if an ingestion fails (e.g. MampfSearch is down when a medium is uploaded), the databases would get out of sync. It also does not process the existing backlog.

For that, a scheduler runs every 30 minutes (`config/schedule.yml`). Every medium tracks one of the following `transcription_status` values:

1. `not_transcribed`: The medium has not yet been transcribed.
2. `queued`: A request was dispatched to transcribe this medium, but it has not yet completed.
3. `completed`: The medium was successfully transcribed and the `.vtt` file is attached.
4. `failed_temporarily`: The previous request to transcribe it was not successful (for any reason); scheduled for automatic retry.
5. `failed_permanently`: Multiple attempts were made to transcribe the medium, which all failed (`MAX_TRANSCRIPTION_ATTEMPTS = 3`). Human intervention is necessary.

The scheduler executes in 3 phases:
- **Phase 1 (`recover_stuck_jobs`):** Detects media stuck in `queued` status past the timeout threshold (`2.hours`, e.g. If the whole Mampfsearch server crashes during transcription), increments attempt counters, and moves them to `failed_temporarily` (or `failed_permanently` if attempts exceed the threshold).
- **Phase 2 (`feed_next_batch`):** Queries `Medium.needs_transcription` (`not_transcribed` or `failed_temporarily` with attempts < threshold) and requests ingestion for media needing transcription (respecting a limit of jobs in the queue at one time).
- **Phase 3 (`reconcile_search_index`):** Compares MaMpf's video IDs with MampfSearch's indexed IDs (`/lesson/list`):
  - Automatically enqueues `MampfsearchDeleteJob` for media deleted in MaMpf but still present in MampfSearch.
  - Automatically re-enqueues `MampfsearchIngestJob` for media marked `completed` in MaMpf but missing from the MampfSearch index (e.g. after a vector database reset).

---

### 1. Setup & Starting Both Services

Precondition: Start MaMpf devcontainer on branch `pr1-mampfsearch-integration`

Then:

1. Clone the [Mampfsearch repository](https://github.com/MaMpf-HD/mampf-ai-search)
2. Switch to the correct branch: `git checkout pr/ingestion-scheduler`
3. Create a .env file based on the development template: `cp .env.example .env`
4. Start in VS Code and open as a container.
5. Seed the Mampfsearch database: `just seed`. This downloads the final transcripts and embeddings for the videos present in the mampf pre-seed.

---

### 2. Automated Test Suites

Run the tagged MampfSearch tests inside the MaMpf devcontainer:
```bash
RAILS_ENV=test bundle exec rspec --tag mampfsearch
# Expected: 84 examples, 0 failures
```

---

#### Via the Web UI (Development Mode)
1. Open MaMpf in your browser (`http://localhost:3000`) and log in as an editor or administrator.
2. Navigate to any lecture with a video medium.
3. In development mode, a **"Transcribe"** button is displayed next to the media player button.
4. Click **"Transcribe"**: the controller accepts the request and enqueues `MampfsearchIngestJob`.

#### Option B: Via Rails Console
1. Open the Rails console:
   ```bash
   rails c
   ```
2. Pick a medium with video and trigger ingestion:
   ```ruby
   medium = Medium.where.not(video_data: nil).first
   medium.update!(transcription_status: :not_transcribed, transcription_attempts: 0)

   MampfsearchIngestJob.perform_now(medium.id)
   ```

#### Monitoring Ingestion & Verifying Output:
- **Watch the logs:**
- **Verify database state in Rails console:**
  ```ruby
  medium.reload
  puts medium.transcription_status   # => "completed"
  puts medium.transcript.present?    # => true
  puts medium.transcription_attempts # => 0
  ```
- **Verify player subtitles:**
  Open transcribed video in your browser and click the **CC** button on the player controls to verify subtitles appear.

---

#### Trigger Full Batch Synchronization
In the Rails console:
```ruby
MampfsearchSyncJob.perform_now
```

---

#### Overall Transcription Status Breakdown
```ruby
Medium.where.not(video_data: nil).group(:transcription_status).count
=> {"not_transcribed" => 4, "completed" => 10}
```

#### List all currently `queued` media (in-flight transcription)
```ruby
Medium.where(transcription_status: :queued).pluck(:id, :description, :transcription_requested_at)
```

#### List all `completed` media (with transcripts attached)
```ruby
Medium.where(transcription_status: :completed).pluck(:id, :description)
```

#### List transcribable media NOT yet transcribed (pending backlog)
```ruby
# Media ready to be picked up by the scheduler
Medium.needs_transcription.pluck(:id, :description, :transcription_status, :transcription_attempts)

#### List failed media (inspecting error messages)
```ruby
Medium.where(transcription_status: [:failed_temporarily, :failed_permanently])
      .pluck(:id, :transcription_status, :transcription_attempts, :transcription_error)
```


app/controllers/media_controller.rb
```diff
@@ -1,12 +1,16 @@
 # MediaController
 class MediaController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: [:add_transcript]
```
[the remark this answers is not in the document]
f-buerckel: This is intended behavior and does not allow a CSRF.

add_transcript is a webhook which can only be called by the MampfSearch background worker. not an interactive browser session. Because the worker does not have Rails session cookies or CSRF authenticity tokens, CSRF protection must be bypassed for this action.

The limited access is enforced by the verify_search_api_token! function at the beginning of the add_transcript action, which checks if the request has a valid token that can only be created by the Mampfsearch API.

app/controllers/media_controller.rb
```diff
@@ -263,6 +276,59 @@ def search
...
+    return head :not_found unless Rails.env.local?
+
+    authorize! :transcribe, @medium
+    unless @medium.transcribable?
+      redirect_back_or_to(root_path, alert: "Medium cannot be transcribed.")
+      return
+    end
+
+    @medium.update!(transcription_attempts: 0, transcription_error: nil)
+    MampfsearchIngestJob.perform_later(@medium.id)
```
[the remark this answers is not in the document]
f-buerckel: This endpoint is only ever used in local development so duplicate request may be desired in case the transcription gets stuck on the API side. To prevent accidental duplicate transcriptions, I added a short notification after pressing the button in https://github.com/MaMpf-HD/mampf/commit/9f74c2bb74d37fca1156259e059f3065cfd74cca.

app/controllers/media_controller.rb
```diff
@@ -263,6 +276,59 @@ def search
...
+    status = if attempts >= SearchClient::MAX_TRANSCRIPTION_ATTEMPTS
+      :failed_permanently
+    else
+      :failed_temporarily
+    end
+    @medium.update!(
+      transcription_status: status,
+      transcription_attempts: attempts,
+      transcription_error: params[:error].presence || "MampfSearch transcription failed."
+    )
```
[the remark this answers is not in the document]
f-buerckel: Extremely unlikely and not dramatic if it would happen, but fixed anyways in https://github.com/MaMpf-HD/mampf/commit/d2e74e5e4763abff2f350bc1a8d31c305feec794.

app/jobs/mampfsearch_sync_job.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      # 2. Re-ingest media marked completed in MaMpf but missing from index (e.g. after reset)
+      completed_ids = Medium.where.not(video_data: nil)
+                            .where(transcription_status: :completed)
+                            .pluck(:id).to_set
+      missing_from_search = completed_ids - indexed_ids
+      missing_from_search.each do |missing_id|
+        MampfsearchIngestJob.perform_later(missing_id)
+      end
```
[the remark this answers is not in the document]
f-buerckel: Good point. Fixed in https://github.com/MaMpf-HD/mampf/commit/92a2d069b732297ecd0b47a66619faac125a7806.

app/models/medium.rb
```diff
@@ -1102,4 +1143,41 @@ def release_state
...
+      end
+    end
+
+    def reset_transcription_state
+      self.transcript = nil
+      self.transcription_status = :not_transcribed unless transcription_status_changed?
+      self.transcription_attempts = 0 unless transcription_attempts_changed?
+      self.transcription_error = nil
+      self.transcription_requested_at = nil unless transcription_requested_at_changed?
+    end
```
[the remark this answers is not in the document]
f-buerckel: Fixed in https://github.com/MaMpf-HD/mampf/commit/ab072845798534f2454cc3e6e6f16c6179e7b872.

lib/search_client.rb
```diff
@@ -0,0 +1,116 @@
...
+      payload[:transcription_failed_url] =
+        transcription_failed_url
+    end
+
+    payload[:lecture_rails_id] = lecture_rails_id if lecture_rails_id.present?
+    payload[:lesson_rails_id] = lesson_rails_id if lesson_rails_id.present?
+
+    perform_request(scope: "/lesson/ingest") do |client|
+      client.post("/lesson/ingest", params: payload)
+    end
```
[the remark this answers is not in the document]
f-buerckel: Implemented in https://github.com/MaMpf-HD/mampf/commit/ff7bd4a513bcac9b51605141f2cf9d6ea60d4c16 and correspondingly for mampfsearch in https://github.com/MaMpf-HD/mampf-ai-search/commit/bc807bbd75239b0c0670819bedb7c8b9401cb1dc.

docker/development/compose.yml
```diff
@@ -148,3 +152,7 @@ volumes:
   sprockets-cache:
   bundle_cache:
   clamav-db:
+
+networks:
+  mampf_dev:
+    external: true
```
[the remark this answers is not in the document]
f-buerckel: Reflected the network creation in the just file in https://github.com/MaMpf-HD/mampf/commit/89bd05d512e1277964bd217728b63561e39e45c2.

app/controllers/media_controller.rb
```diff
@@ -263,6 +276,61 @@ def search
...
+
+    authorize! :transcribe, @medium
+    unless @medium.transcribable?
+      redirect_back_or_to(root_path, alert: "Medium cannot be transcribed.")
+      return
+    end
+
+    @medium.update!(transcription_attempts: 0, transcription_error: nil)
+    MampfsearchIngestJob.perform_later(@medium.id)
+    redirect_back_or_to(root_path, notice: "Transcription started.")
```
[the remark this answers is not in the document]
f-buerckel: Fixed in https://github.com/MaMpf-HD/mampf/commit/a410c02521ca676c38a788fb19b9f6f43821fd6e.

app/jobs/mampfsearch_sync_job.rb
```diff
@@ -0,0 +1,97 @@
...
+        MampfsearchDeleteJob.perform_later(orphan_id)
+      end
+
+      # 2. Re-ingest media marked completed in MaMpf but missing from index (e.g. after reset)
+      completed_ids = Medium.where.not(video_data: nil)
+                            .where(transcription_status: :completed)
+                            .pluck(:id).to_set
+      missing_from_search = completed_ids - indexed_ids
+      Medium.where(id: missing_from_search).find_each do |medium|
+        claim_and_enqueue(medium)
```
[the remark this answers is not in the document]
f-buerckel: Fixed and refactored in https://github.com/MaMpf-HD/mampf/commit/6aa83ed13969f96f64d9b59524a15167996cba9e.

db/migrate/20260911000000_add_transcript_data_to_media.rb
```diff
@@ -0,0 +1,13 @@
+class AddTranscriptDataToMedia < ActiveRecord::Migration[8.0]
```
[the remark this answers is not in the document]
f-buerckel: Fixed during merge of next in https://github.com/MaMpf-HD/mampf/commit/3ffca9f0686860ea70773f319afb5a97216939ea.

fosterfarrell9 (commented):
Reviewed this revision against MampfSearch PR #1 at `bc807bbd75239b0c0670819bedb7c8b9401cb1dc`, including the surrounding authorization, storage, playback, queue and database code.

Validation: the MaMpf integration suite passed (87 examples), the surrounding medium model/controller/request run passed (92 examples; overlaps with the integration run), and the linked service's boundary/offline tests passed (32 tests). All MaMpf RSpec commands used `VITE_RUBY_PORT=3036 RAILS_ENV=test bundle exec rspec ...`. Additional focused checks reproduced the inline findings, including a local HTTP concurrency reproduction.

The ordinary transcript stream retains the existing playback permission predicate; checks denied guests and unrelated authenticated users access to private transcripts. The service endpoints require both tokens, subtitle rendering uses `textContent`, and file serving adds `nosniff`. I found no confirmed browser-script execution path.

Deployment questions, separate from the confirmed defects: verify the actual MAMPFSEARCH_BASE_URL transport and WireGuard coverage, production secret configuration, ingress/body limits, and capability-URL handling in proxy/RQ/error logs. Rails filters the token parameter, but worker exceptions can retain full URLs.

The confirmed FFmpeg TLS-verification defect is posted on the affected worker code in the [linked MampfSearch review](https://github.com/MaMpf-HD/mampf-ai-search/pull/1#pullrequestreview-5198415229).

app/models/medium.rb
```diff
@@ -1102,4 +1147,39 @@ def release_state
 
       "unpublished"
     end
+
+    def handle_video_attachment_change
+      return unless video_attacher.changed?
+
+      @video_detached_to_purge = true if video.blank?
+      reset_transcription_state
```
fosterfarrell9, line 1155: **[P1] Invalidate the remote transcript when replacing a video**

Replacing a video clears MaMpf's transcript and queues ingestion, but only detaching it schedules remote deletion. The [linked pipeline](https://github.com/MaMpf-HD/mampf-ai-search/blob/bc807bbd75239b0c0670819bedb7c8b9401cb1dc/app/services/ingest.py#L35-L43) calls `load_transcript(media_rails_id)` and reuses the cached transcript without downloading the replacement. Since the cache is keyed only by medium ID, the old subtitles and indexed speech return after a normal replacement. Replacing a confidential recording with a publishable edit can consequently expose speech removed from the replacement.

Focused checks reproduced both the missing invalidation in MaMpf and the skipped transcription in MampfSearch. Please include an immutable video version in the ingestion contract/cache key and invalidate obsolete versions with ordering guarantees; independently enqueueing deletion and ingestion would introduce another race.

app/controllers/media_controller.rb
```diff
@@ -621,6 +708,31 @@ def feedback
...
+      SearchApiToken.verify!(token, scope: request.path)
+      true
+    rescue SearchApiToken::InvalidTokenError, SearchClient::ServiceUnavailableError
+      head :unauthorized
+      false
+    end
+
+    def verify_transcription_token!(purpose:)
+      payload = TranscriptionToken.verify!(params[:token], purpose: purpose)
+      return true if payload.fetch("medium_id").to_i == @medium.id
```
fosterfarrell9, line 728: **[P1] Bind resource capabilities to the current video version**

This verifies the medium ID but not the video attachment or generation. A token issued for video A remains valid after the same medium is changed to video B. The upload action also accepts the old capability after the video is detached, so delayed workers or a compromised service can restore removed transcripts or overwrite newer subtitles.

Request checks reproduced an old download token returning HTTP 200 after replacement, and an old transcript token returning HTTP 200 after detachment. Please bind the capability to an immutable attachment version, revoke it on replacement/removal, and verify that version atomically when accepting callbacks.

app/controllers/media_controller.rb
```diff
@@ -259,6 +272,61 @@ def search
...
+      attempts = @medium.transcription_attempts + 1
+      status = if attempts >= SearchClient::MAX_TRANSCRIPTION_ATTEMPTS
+        :failed_permanently
+      else
+        :failed_temporarily
+      end
+      @medium.update!(
+        transcription_status: status,
+        transcription_attempts: attempts,
+        transcription_error: params[:error].presence || "MampfSearch transcription failed."
```
fosterfarrell9, line 324: **[P2] Count each dispatched job's terminal failure only once**

Every accepted callback increments the attempt count regardless of job identity or current state. This is exercised by the actual [linked failure callback](https://github.com/MaMpf-HD/mampf-ai-search/blob/bc807bbd75239b0c0670819bedb7c8b9401cb1dc/app/services/ingest_worker.py#L27-L34): RQ invokes it before deciding whether to retry, and it reports failure even when retries remain. With `Retry(max=2, interval=[30, 120])`, one dispatched job can consume all three MaMpf attempts during a short outage and become permanently failed instead of getting the intended later retry.

Focused checks also showed that replaying one callback three times produces `transcription_attempts == 3`, and that a delayed failure downgrades an already completed medium. Please report terminal RQ failures only and identify dispatched attempts so their terminal result can be accepted atomically once; acknowledge duplicate/stale callbacks without changing state.

app/services/transcription_token.rb
```diff
@@ -0,0 +1,69 @@
+class TranscriptionToken
+  ALGORITHM = "SHA256".freeze
+  VIDEO_TTL = 5.minutes
+  TRANSCRIPT_TTL = 1.hour
+  # The callback token must survive all RQ attempts and their job timeouts.
+  FAILED_TTL = 4.hours
```
fosterfarrell9, line 6: **[P2] Account for service queueing and retries in capability lifetimes**

These tokens are generated before the [service queues the job](https://github.com/MaMpf-HD/mampf-ai-search/blob/bc807bbd75239b0c0670819bedb7c8b9401cb1dc/app/endpoints/ingest.py#L38-L50). The configured worker processes jobs sequentially, while MaMpf submits batches of ten and each service job may run for 3,500 seconds. A job waiting more than five minutes therefore starts with an expired video URL. RQ retries reuse that URL; generating a fresh bearer token does not renew the resource capability. The one-hour upload token has the same queue/runtime mismatch.

A request check advanced time by six minutes and received HTTP 403 with an otherwise valid, newly generated bearer token. Please obtain or renew video-version-bound capabilities when the worker starts, using an explicit bounded job lease, rather than relying on these pre-queue deadlines.

app/controllers/media_controller.rb
```diff
@@ -259,6 +272,61 @@ def search
...
+  def add_transcript
+    return unless verify_search_api_token!
+    return unless verify_transcription_token!(purpose: :transcript)
+
+    if params[:transcript].present?
+      old_transcript = @medium.transcript
+      @medium.transcript = params[:transcript]
+      @medium.transcription_status = :completed
+      @medium.transcription_attempts = 0
+      @medium.transcription_error = nil
```
fosterfarrell9, line 298: **[P2] Reset ingestion attempts only after indexing succeeds**

The [linked pipeline uploads the VTT before chunking, embedding and database insertion](https://github.com/MaMpf-HD/mampf-ai-search/blob/bc807bbd75239b0c0670819bedb7c8b9401cb1dc/app/services/ingest.py#L45-L49). Marking the medium completed and resetting attempts here therefore acknowledges an ingestion that may still fail. During an embedding outage each retry successfully uploads the cached VTT, resets attempts to zero, then fails again. A focused request check ran three upload/failure cycles and ended at `failed_temporarily` with just one attempt, so the retry limit is never reached.

Premature completion also lets reconciliation enqueue another ingestion while the original worker is still indexing. Please separate transcript availability from ingestion completion and reset the ingestion attempt counter only after an explicit acknowledgement of a successful index transaction.

lib/search_client.rb
```diff
@@ -0,0 +1,117 @@
...
+      response = @pool.with do |client|
+        request_client = if scope
+          token = SearchApiToken.generate(scope: scope)
+          client.headers(authorization: "Bearer #{token}")
+        else
+          client
+        end
+        yield(request_client)
+      end
+      handle_response(response)
```
fosterfarrell9, line 95: **[P2] Consume the response while holding the pooled connection**

The HTTP response body is lazy, but `handle_response` reads it after `@pool.with` has returned the persistent connection. A second thread can reuse that connection while the first is still reading its body. In a controlled local-server reproduction with two requests and a delayed first body, both requests failed with closed-stream errors despite the server being healthy. Moving response consumption inside the pool block made both requests succeed in the same reproduction.

Please consume/parse the response before releasing the connection, and drain or close unsuccessful responses before returning it to the pool.

app/models/medium.rb
```diff
@@ -1102,4 +1147,39 @@ def release_state
...
+
+      @video_detached_to_purge = nil
+      MampfsearchDeleteJob.perform_later(id)
+    end
+
+    def trigger_transcription_if_video_attached_or_changed
+      return unless saved_change_to_video_data? && transcribable?
+      return unless not_transcribed?
+
+      MampfsearchIngestJob.perform_later(id)
```
fosterfarrell9, line 1183: **[P2] Synchronize hierarchy changes as well as video changes**

Reassigning a completed medium to another lecture leaves `video_data` unchanged, so this callback queues nothing. The reconciliation job compares only medium IDs; because the ID is still indexed, it never updates MampfSearch's old course/lecture/lesson association. A focused model check confirmed that moving a completed medium queues no synchronization.

Please explicitly synchronize hierarchy changes and reconcile a hierarchy version or fingerprint. Changes to a parent lecture's course or a lesson's lecture also need propagation to the affected media.

app/jobs/mampfsearch_sync_job.rb
```diff
@@ -0,0 +1,100 @@
...
+                                 "#{SearchClient::MAX_TRANSCRIPTION_ATTEMPTS} attempts"
+          )
+          Rails.logger.error("Mampfsearch transcription permanently timed out for medium " \
+                             "#{medium.id} after " \
+                             "#{SearchClient::MAX_TRANSCRIPTION_ATTEMPTS} attempts")
+        else
+          medium.update!(
+            transcription_attempts: medium.transcription_attempts + 1,
+            transcription_status: :failed_temporarily,
+            transcription_error: "Job timed out in worker, will retry"
```
fosterfarrell9, line 27: **[P3] Determine timeout status from the incremented attempt count**

When a queued medium has two previous failures, this increments attempts to three but assigns `failed_temporarily`. `needs_transcription` excludes attempts >= 3, and `stuck_transcriptions` only selects queued records, so the medium never retries or transitions to `failed_permanently`, despite the error saying it will retry. A focused check reproduced this stranded state.

Please calculate the new attempt count first and assign `failed_permanently` when that count reaches the limit.

app/models/medium.rb
f-buerckel: Fixed in: https://github.com/MaMpf-HD/mampf/commit/14ab3e07d8818ce11061a874786f420c53bf22f2 and https://github.com/MaMpf-HD/mampf-ai-search/commit/c029faf9efb0b0abed1c86c73a2878e399a386fe.

app/controllers/media_controller.rb
f-buerckel: Fixed in https://github.com/MaMpf-HD/mampf/commit/3b1062cc25bdec4826e520fa2a2bf3bf4e239b11.

app/controllers/media_controller.rb
f-buerckel: Yes I agree. Probably a better approach to call it a single failure after all rq retries failed. Changed it in https://github.com/MaMpf-HD/mampf/commit/14ab3e07d8818ce11061a874786f420c53bf22f2. The case for a delayed failure is an extreme edge case and I see no way of this happening in production. If it would happen the consequences would not be severe enough to justify the additional complexity.

app/services/transcription_token.rb
f-buerckel: Yes the TTL was indeed too short for the full batch in most cases. I increased the time to 48 hours in https://github.com/MaMpf-HD/mampf/commit/7dec928c443618c7a6da7cc6607f5dba0055c7cd. This is definitely enough for one batch except something goes seriously wrong, in which case its okay if it fails.

app/controllers/media_controller.rb
f-buerckel: Mainly fixed in https://github.com/MaMpf-HD/mampf-ai-search/commit/0712879ef52ef615f09289d6c250f6d8a7d60aa0 by dropping duplicate jobs, so if a medium gets ingested by  the sync job, while it is already in the redis queue, it gets dropped.

This sadly has one side effect: If a transcription job is not finished but has uploaded the transcript. A sync job could ingest that media again and therefore it will have the transcription status 'queued'. The api immediately drops this job because the same one is already in the queue. However, if the original job already uploaded the subtitles (transcription_status -> completed) the sync job enqueues it again (transcription status -> queued), then this job will get dropped and it will stay in the transcription status queued until it will be marked as temporarily_failed despite being successful. Therefore now, the sync job will not set an already completed job to queued if it gets enqueued again (done in https://github.com/MaMpf-HD/mampf/commit/d3078cd0666f2795df5c9622198fabb56c8c19ed). This could result in an edge case where mampf could enqueue more jobs then its queue limit. A more complete fix would have an additional ingest success callback but I don't want to keep callbacks and complexity to handle specific edge cases.

Also stopped resetting the transcription attempts after transcript upload in https://github.com/MaMpf-HD/mampf/commit/2cc922eb435c7a355fa5d44877e87a26c67b316c

lib/search_client.rb
f-buerckel: Fixed in: https://github.com/MaMpf-HD/mampf/commit/ac61eaf8388ba4069977383534991a18dcc94513

app/models/medium.rb
f-buerckel: Good point, didn't realize the hierarchy can change but this is possible both in explicit in the UI and implicit if a parent media is deleted. This is addressed in https://github.com/MaMpf-HD/mampf/commit/78c9d0bc0ec94f9e20af907e4384c3c866faaded and https://github.com/MaMpf-HD/mampf-ai-search/commit/abb171ccae7da5a29fb1633edfcc3daee2f80749. MaMpf now queues metadata sync when a medium changes its association, or when a parent lecture, lesson, or talk moves. The Mampfsearch Sync Job compares the saved hierarchy in Mampfsearch and updates them if necessary.

app/jobs/mampfsearch_sync_job.rb
f-buerckel: Fixed in https://github.com/MaMpf-HD/mampf/commit/c6544d212bffa60e9ac88bc592f2e81ea12b4281.

PR #1265, opened 2026-08-28
Show a card for a medium that belongs to nothing

fosterfarrell9:
Reported from staging: `watchlists#show` answers 500 with `undefined method 'card_header_path' for nil`.

`Medium#teachable` is optional and a quiz medium has none, but the card partial asks it for the header and its path without checking. A watchlist takes any medium, so one such entry takes the whole page down; the seed has one (`RandomQuiz`, no teachable) and a student's watchlist points at it.

The card now renders without a header link for those, which is what the partial already does when the path comes back empty. A request spec on `watchlists#show` covers it and fails without the change.

### DIFF
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -508,10 +508,12 @@ class Medium < ApplicationRecord
 
   # methods that create card header and subheader for a medium card
 
-  delegate :card_header, to: :teachable
+  # A medium does not have to belong to a teachable, and one that does not still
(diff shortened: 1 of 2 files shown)


fosterfarrell9:
Closed in favour of #1267. A random quiz has no card worth rendering: it belongs to no lecture, has no page, and the cleanup throws it away a day later. The fix belongs where the entry is made — a watchlist does not accept one — and in the seed, which carries two such entries from an earlier edition.

PR #1291, opened 2026-09-19
Take the stray "vignettes/question:" out of the team-size message

fosterfarrell9:
A line indented one level too deep in `de.yml` (since #737) was read as part of the folded message above it, so a student refused for a full team saw "… hätte als erlaubt. vignettes/question:". The spec asserts a one-line message in every locale.


fosterfarrell9:
Not needed on next: no submissions run there yet, and the fix is on muesli-integration (248ccf816), where the new submission system goes through the same validation.

PR #1296, opened 2026-09-21
Fold and restamp the migrations for the merge into next

fosterfarrell9:
Twenty-one of the branch's migrations sort before next's newest; they are dated after it now, in their old order. Three that undid or extended an earlier one on the same table are folded in (the student_performance_records column, the lecture columns, the assignments index + kind). The deletion-date backfill no longer goes through the models; old and new give the same dates on a copy of the development database, edge cases included. Schema unchanged but for the version and one constraint: `exams.self_materialization_mode` is NOT NULL like on every other roster table.

A development database that has run the old versions will refuse the new ones (`relation "assessment_assessments" already exists`). Easiest: check out `next`, `just seed`, check this branch out, `bin/rails db:migrate` - the path production will take.


fosterfarrell9:
On the exams column (overview, no thread): right - tutorials, cohorts, talks and lectures carry it NOT NULL, exams did not. Fixed in c94079f20 inside `create_exams`, since nothing has run it; the schema now differs from the base in that one constraint besides the version.

